PR #28, opened 2016-09-23
Updated extracts table

Merfoo:
Modification to extracts table
* Removed required `quantity_concentration` of type decimal column
* Added optional `concentration_value` of type decimal column
* Added optional `concentation_unit` of type string column

### DIFF
--- a/app/controllers/extracts_controller.rb
+++ b/app/controllers/extracts_controller.rb
@@ -84,6 +84,6 @@ class ExtractsController < ApplicationController
 
     # Never trust parameters from the scary internet, only allow the white list through.
     def extract_params
-      params.require(:extract).permit(:quantity_value, :quantity_unit, :concentration_value, :concentration_unit, :verbatim_anatomical_origin, :year_made, :month_made, :day_made, :created_by_id, :updated_by_id, :project_id)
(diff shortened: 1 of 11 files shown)


app/controllers/extracts_controller.rb
```diff
@@ -84,6 +84,6 @@ def set_extract
 
     # Never trust parameters from the scary internet, only allow the white list through.
     def extract_params
-      params.require(:extract).permit(:quantity_value, :quantity_unit, :quantity_concentration, :verbatim_anatomical_origin, :year_made, :month_made, :day_made, :created_by_id, :updated_by_id, :project_id)
```
mjy: Remove created_by_id, updated_by_id, project_id from here, those are handled by housekeeping.

app/models/extract.rb
```diff
@@ -4,7 +4,6 @@ class Extract < ActiveRecord::Base
 
   validates_presence_of :quantity_value 
   validates_presence_of :quantity_unit
-  validates_presence_of :quantity_concentration 
   validates_presence_of :verbatim_anatomical_origin 
   validates_presence_of :year_made
```
mjy: Check the validation rules on collecting event start_ fields.  Dupe them over here for the date related validations.  Maybe write a spec for them.

PR #29, opened 2016-10-01
Dates validations

Merfoo:
Added 6 custom validators to use on model attributes

* DateDayValidator
* DateMonthValidator
* DateYearValidator
* TimeSecondValidator
* TimeMinuteValidator
* TImeHourValidator

The validators check if a model attribute is empty, of an integer value, and within range.

To use a custom validator on a model attribute, in the model file write
`validates :model_attribute, date_day: true`

To pass in parameters to the custom validator for customization
`validates :model_attribute, date_day: { min_day: 2, max_day: 28 }`

Optional parameters
* `allow_blank`: if true allows model attribute to be empty, defaults to true
* `message`: error message to be attached to model attribute if validation fails
* `min_day/month/year/etc`: minimum value of  model attribute
* `max_day/month/year/etc`: maximum value of  model attribute

Optional parameters unique to DateDayValidator
* 'year_sym`: the symbol of the model attribute containing the year value for the model
* 'month_sym`: the symbol of the model attribute containing the month value for the model

`validates :model_attribute, date_day: { year_sym: :year, month_sym: :month }`

If both `year_sym` and `month_sym` are given to a DateDayValidator then the `max_day` value will be the amount of days that exists of the month and year specified

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ branches:
       - dna
       - eyelet
       - batch_file_load
+      - dates-validations
 addons:
   postgresql: "9.4"
   firefox: "46.0"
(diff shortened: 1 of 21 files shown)


mjy:
Thanks much Fauzi, nicely done.  For another layer of abstraction we should pull this out and make it a Gem. ;)

PR #74, opened 2016-12-21
Dwc occurrences

mjy:
New Model
-----------

Adds a Darwin Core Occurrence index, DwcOccurrence.

* New database table is based of the IPT standard, generated from dwc-meta, a code base I have yet to push.
* Has polymorphic relationship to CollectionObject and AssertedDistribution, only the former is implemented, the latter is stubbed.

New Concern b/w DwcOccurrence and Models
-----------------------------------------------

Shared functionality in serializing DwcOccurrence records is in Shared::IsDwcOccurrence.  Includes full set of specs.

New layer b/w Dwc/TW
-----------------------

Refactors approach to mapping TW to DWC (see _extensions in CollectionObject, for e.g.)

Modified Task
--------------

Modifies the dwca report task to report the data in DwcOccurrence.  Currently set to generate the index if requested (per page), as an experiment.  Include simple download link.

New Library - 'Download'
-------------------------

Adds a generic library to convert scopes into CSV downloads.  Refactors Otu to show example use.  Need to remove download code in individual models once merged.

New pattern - delegate
-----------------------

Introduce the use of `delegate` to reference methods in subclasses.  See TaxonName for example.  Used extensively in CollectionObject.  Deprecated associate methods where made redundant (e.g. in CollectionObject).

Introduces use of delayed_job
------------------------------

After creation of CollectionObjects their (future) indexing is handled by delayed_job.  There are consequences for configuring environments that will be detailed on the development branch.


lib/utilities/strings.rb
```diff
@@ -58,9 +59,18 @@ def self.termify(string)
...
   def self.is_i?(string)
     /\A[-+]?\d+\z/ === string
   end
 
+  # @return [String, param]
+  #   the goal is to sanitizie an individual string such that it is usable in *TAB* delimited, UTF-8, column.  See Download
+  #   TODO: Likely need to handle quotes
+  def self.sanitize_for_csv(string)
+    return string if string.blank?
+    string.to_s.gsub(/\n/, '\n').gsub(/\t/, '\t')
```
LocoDelAssembly: Consumers won't necessarily understand this C-like character escape sequences. Typically you would replace this by empty string and a space respectively. However, it looks like CSV tolerates new lines and tabs as demonstrated by the following script:
```
hernan@hernan-pc:~$ irb
2.2.1 :001 > require 'csv'
 => true 
2.2.1 :002 > csv_string = CSV.generate do |csv|
2.2.1 :003 >       csv << ["Multi\nLine", "name:\tHernán", "Computers are \"silly\" devices"]
2.2.1 :004?>   end
 => "\"Multi\nLine\",name:\tHernán,\"Computers are \"\"silly\"\" devices\"\n" 
2.2.1 :005 > puts csv_string
"Multi
Line",name:	Hernán,"Computers are ""silly"" devices"
 => nil 
```
But yet, to be extra safe I would suggest replacing \t with " " and the rest of the control characters with "". Quotes are escaped by duplicating them.

lib/download.rb
```diff
@@ -0,0 +1,75 @@
...
+  # @params trim_columns [Array] delete columns from result if no data provided
+   # @return [String]
+  #   translate a scope into a CSV table, with optional tweaks to the data
+  def self.generate_csv(scope, exclude_columns: [], trim_rows: false, trim_columns: false)
+    column_names = scope.columns_hash.keys
+    headers = CSV::Row.new(column_names, column_names, true)
+    tbl = CSV::Table.new([headers])
+    
+    scope.find_each do |o|
+      tbl << o.attributes.values_at(*column_names).collect{|v| Utilities::Strings.sanitize_for_csv(v) }
```
LocoDelAssembly: See the comment for Utilities::String.sanitize_for_csv, but here you could explore not sanitize at all and let Ruby's csv library do that for you (but again see the other comment).

app/models/dwc_occurrence.rb
```diff
@@ -0,0 +1,58 @@
...
+#
+# Important: This is a cache/index, data here are periodically (regenerated) from multiple tables in TW.  
+#
+# TODO: The basisOfRecord CVTs are not super informative.
+#    We know collection object is definitely 1:1 with PreservedSpecimen, however
+#    AssertedDistribution could be HumanObservation (if source is person), or ... what? if 
+#    its a published record.  Seems we need a 'PublishedAssertation', just like we model the data.
+#
+# DWC attributes are camelCase to facilitate matching
+# dwcClass is a replacement for the Rails reserved 'Class'
```
mjy: Thanks for reminding me of this- there are a couple exceptions we have to catch like this.

lib/utilities/strings.rb
mjy: Good points.  I was thinking round tripping perhaps, but you're right.  If we do use markup we should pick something like Markdown.  Perhaps make it optional.

lib/download.rb
mjy: Aha!  I battled with this in a number of different ways.  I suppose we need a spec to ensure the intent is clear.  I'm not positive that the CSV::Table translates to CSV with bad values in it, something nagging reminding me that I tried this.

PR #144, opened 2017-06-29
Psu import task

TuckerJD:
Adding a generic DWCA tab-separated-value batch-load importer.
Adding tests for the batch loader, with test files.
Adding  an interface for the importation of the files from Penn State (https://docs.google.com/spreadsheets/d/1kzqHU_guLZP3J-fRPs1cC98Ua7lbOP8SXNeR2zi1LA0/edit?usp=sharing).


app/controllers/tasks/import/dwca/psu_import_controller.rb
```diff
@@ -36,9 +37,50 @@ def do_psu_import
...
+    params.permit(:dwca_namespace, :file, :import_level).merge(user_id: sessions_current_user_id, project_id: sessions_current_project_id).symbolize_keys
   end
 
+  # what to do (because you are PSUC_FEM) before you try to load the entire file
+  def pre_load
+    pre_load = {}
+    root     = Protonym.find_or_create_by(name:       'Root',
+                                          rank_class: 'NomenclaturalRank',
+                                          parent_id:  nil,
+                                          project_id: $project_id)
```
TuckerJD: In process...

app/helpers/tasks/import/dwca/psu_import_helper.rb
```diff
@@ -1,3 +1,35 @@
 module Tasks::Import::Dwca::PsuImportHelper
 
+  def text_row(row)
```
TuckerJD: Eventually, perhaps: The name makes it consistent with the controller and HTML file names as per convention.

app/controllers/tasks/import/dwca/psu_import_controller.rb
```diff
@@ -8,19 +8,20 @@ def index
   end
 
   # POST
-  def preview_psu_import
+  def do_psu_import
```
TuckerJD: There is an isolation of the DWCA part, and the PSU preamble (the controller is, after all, the PSU import controller) The actual batch-load is called from the PSU controller, and is "basic DWCA" (with one exception). There are some PSU-specific objects built in the controller method pre_load.
The exception to which I referred (above) is that the PSU data includes the 'PSUC_FEM' preamble for catalogNumber, which I have used as a namespace, and have isolated from actual catalog number, as per the 'catalogNumber' section of the design document (https://docs.google.com/document/d/18iTlim9lLQXXTdkNUJGXP1WIPcRbYNfkz8dkLUq1YFw/edit)
This one bit of PDU-specific code is in the make_ident section.
In addition, only the columns specific to the PSU data are processed in the current version, although adding additional DWCA-defined (http://rs.tdwg.org/dwc/terms/#dcindex) terms should be simple enough.

app/helpers/tasks/import/dwca/psu_import_helper.rb
```diff
@@ -1,3 +1,35 @@
...
+        # end
+      end
+    end
+    output += '</li>'
+    output.html_safe
+  end
+
+  # @return [String]
+  #   short string of use in autocomplete selects
+  def collection_object_namespace_select_tag
```
TuckerJD: Done.

app/views/tasks/import/dwca/psu_import/do_psu_import.html.erb
```diff
@@ -0,0 +1,21 @@
+<h2> Penn State Import Results</h2>
+
+<p> Import includes <%= @result.count -%> records. </p>
+
+<!--<p> To complete this import, select the file again and click create:-->
```
TuckerJD: Done.

config/routes.rb
```diff
@@ -431,7 +431,7 @@
       scope :dwca do
         scope :psu_import, controller: 'tasks/import/dwca/psu_import' do
           get 'index', as: 'psu_import_task'
-          post 'preview_psu_import', as: 'preview_psu_import'
+          # post 'preview_psu_import', as: 'preview_psu_import'
```
TuckerJD: Done.

spec/lib/batch_load/import/dwca_spec.rb
```diff
@@ -0,0 +1,78 @@
+require 'rails_helper'
+require 'batch_load/import/collection_objects'
+
+describe BatchLoad::Import::DWCA, type: :model do
+  # iconv -f iso-8859-15 -t UTF-8  spec/files/batch/dwca/PSUC3-Test.txt >spec/files/batch/dwca/PSUC3-Test.utf8.txt
```
TuckerJD: Encapsulated in an explanatory context, but I think it is where it belongs, given that it is used to generate the test file for the spec.

spec/lib/batch_load/import/dwca_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    let(:upload_file) {fixture_file_upload(file_name)}
+    let(:root) {Protonym.find_or_create_by(name:       'Root',
+                                           rank_class: 'NomenclaturalRank',
+                                           parent_id:  nil,
+                                           project:    project)
+    }
+    let(:kingdom) {Protonym.find_or_create_by(name:       'Animalia',
+                                              parent:     root,
+                                              rank_class: NomenclaturalRank::Iczn::HigherClassificationGroup::Kingdom,
```
TuckerJD: Done.

TuckerJD: Done.

PR #153, opened 2017-07-20
Documentation Rake Task

Merfoo:
Adds rake task to create Documents and attaches them to object identified by an identifier via Documentations. Task requires a directory of files to create as Documents and a specifying which object to attach the document to and which file to create as a document. The file format is with columns of 
"namespace_short_name", "identifier", "filenames". Filenames can contain multiple filenames when separated with a ", " (comma and space). Adds 'application/octet-stream' as a valid content_type for Documents and adds 'ab1' as a valid file extension for ''application/octet-stream' files.

### DIFF
--- a/lib/tasks/batch_load/documentations.rake
+++ b/lib/tasks/batch_load/documentations.rake
@@ -25,7 +25,7 @@ namespace :tw do
 
         csv = CSV.parse(File.read(meta_data_file_path), { headers: true, col_sep: "\t" })
 
-        puts "Processing files".yellow
+        puts Rainbow("Processing files").yellow
         documents_created = 0
         documentations_created = 0
         rows_processed = 0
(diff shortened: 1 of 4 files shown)


mjy:
@Merfoo You could also merge the identifier/short name into just one column "identifier", so the user provides the unified value, that you match against identifier.cached.

lib/tasks/batch_load/documentations.rake
```diff
@@ -49,9 +47,11 @@ namespace :tw do
                 documentation.save!
```
mjy:
```ruby
begin
ActiveReord::Base.transaction do
#...
end
rescue ActiveRecord::RecordInvalid
# ... log/report
end
```

Merfoo:
If a Document already exists for a new Documentation I should use the existing Document
If a Document and Identifier Object pair already exists, don't create a new Documentation
@mjy

PR #157, opened 2017-07-30
Removed returning values from validation functions in global identifiers

Merfoo:
Removed `retval` and other returning values in the validation functions for most of the global identifiers since it doesn't matter what gets returned for validation functions.

### DIFF
--- a/app/models/identifier/global/isbn.rb
+++ b/app/models/identifier/global/isbn.rb
@@ -25,7 +25,6 @@ class Identifier::Global::Isbn < Identifier::Global
   validate :using_isbn_class
 
   def using_isbn_class
-    retval = true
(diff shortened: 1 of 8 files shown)


mjy:
Thanks for noticing this and tuning it up @Merfoo.

My only comment is that there is an inherent pattern that could also be improved with these tests. It comes down to this- ideally a validation should only include a single error check, not a set of nested, or sequential checks.  This would mean refactoring the code to include more `validates :foo` lines, with more corresponding methods.  Since you are looking at this code now it might be a good time to do some of that refactoring (along with corresponding specs if there is such).  Let me know your thoughts. Of course this request could be done in a subsequent pull request as well.

PR #272, opened 2017-10-19
Filter user date

TuckerJD:
New facet for CollectionObject filter page.

New class methods of User:
* `User.get_user_id(user)` to get one specific user, accepts `String`, `User`, `Integer`, requires exact match if string is used (name or email)
* `User.get_user_ids(user, user, user, ...)` to get array of Ids, accepts `String`, `User`, `Integer`, uses partial string match (name or email)

New helper for users:
* `project_users_select_tag(user_element, default_name = nil)`: Select of all users in the current project, returns 'All users' or id of selected user, If default id is not supplied, 'All users' is selected.
* `user_select_tag(user_element, user_id_list, default_name = nil)`: As project_users_select_tag, but not limited to current project users.


mjy:
@TuckerJD can you resolve the pack-lock issue?

app/controllers/tasks/collection_objects/filter_controller.rb
```diff
@@ -47,22 +47,60 @@ def set_otu
...
 
+  # GET
   def set_id_range
-    render json: {html: collection_objects.count.to_s}
+    render json: {html: collection_objects.count}
+  end
+
+  # GET
+  def set_user_date_range
+    params[:user] = params[:user].to_i
```
mjy: This seems odd, why here?

app/controllers/collection_objects_controller.rb
```diff
@@ -98,7 +98,9 @@ def destroy
 
   # GET /collection_objects/list
   def list
-    @collection_objects = CollectionObject.with_project_id(sessions_current_project_id).order(:id).page(params[:page]) #.per(10) #.per(3)
+    @collection_objects = CollectionObject.with_project_id(sessions_current_project_id)
+                            .order(:id)
+                            .page(params[:page]) #.per(10) #.per(3)
```
mjy: Remind me, is params[:page] "safe"?

app/controllers/tasks/gis/report_controller.rb
```diff
@@ -144,7 +144,8 @@ def repaint
                       params[:selection_object],
                       params[:page])
     # remove the list from the session so that they are not included in the links generated by pagination
-    render 'location_report_list'
+    redirect_to '/tasks/gis/location_report_list'
+    # render 'location_report_list'
```
mjy: Remove comments.

app/helpers/tasks/collection_objects/area_and_date/area_and_date_helper.rb
```diff
@@ -17,4 +17,22 @@ def id_range_select_tag(id_range_element, option_list)
     select_tag(id_range_element, options_for_select(option_list), prompt: 'Select an identifier')
   end
 
+  def get_first_created_at
+    colobj = CollectionObject.with_project_id(sessions_current_project_id).first_created
+    if colobj.blank?
+      '1700/01/01'
```
mjy: Reference a constant?

app/helpers/tasks/collection_objects/area_and_date/area_and_date_helper.rb
```diff
@@ -17,4 +17,22 @@ def id_range_select_tag(id_range_element, option_list)
...
+    else
+      colobj.created_at.to_date.to_s.gsub('-', '/')
+    end
+  end
+
+  def get_last_created_at
+    colobj = CollectionObject.with_project_id(sessions_current_project_id).last_created
+    if colobj.blank?
+      Date.today.to_s.gsub('-', '/')
```
mjy: For future consideration: https://stackoverflow.com/questions/41965408/rails-4-erb-consistent-date-format

app/models/user.rb
```diff
@@ -149,11 +149,65 @@ def administers_projects?
     administered_projects.any?
   end
 
+  # @param [String, User, Integer] user
+  # @return [Integer] selected user id
+  def self.get_user_id(user)
+    user_id = $user_id
```
mjy: NO.  Get rid of $user_id.

lib/housekeeping/timestamps.rb
```diff
@@ -9,16 +9,22 @@ module Housekeeping::Timestamps
...
     # related_table_name = self.table_name
 
-    scope :created_before_date, -> (date) { where('created_at < ?', "#{date}") }
+    scope :created_before_date, ->(date) { where('created_at < ?', "#{date}") }
+    scope :created_in_date_range, ->(start, c_end) { where('created_at >= ? and created_at <= ?', start, c_end) }
+    scope :updated_in_date_range, ->(start, u_end) { where('updated_at >= ? and updated_at <= ?', start, u_end) }
+    scope :first_created, -> { all.order(created_at: :asc).first }
+    scope :last_created, -> { all.order(created_at: :desc).first }
+    scope :first_updated, -> { all.order(updated_at: :asc).first }
+    scope :last_updated, -> { all.order(updated_at: :desc).first }
```
mjy: Are these scopes returning Scopes or instances?!  A scope with limit(1) I expect?

lib/utilities/dates.rb
```diff
@@ -257,7 +257,7 @@ def self.normalize_and_order_dates(start_date, end_date)
 
     start_date, end_date = order_dates(start_date, end_date)
 
-    return start_date, end_date
+    return start_date.gsub('-', '/'), end_date.gsub('-', '/')
```
mjy: #strftime rathern than gsub?

spec/features/tasks/collection_objects/filter_spec.rb
```diff
@@ -214,6 +214,95 @@
           end
         end
       end
+
+      context 'with records specific to users and dates' do
+        let!(:pat) { User.find(1) }
+        let!(:pat_admin) { User.where(name: 'Pat Project Administrator').first }
+        let!(:joe) { FactoryGirl.create(:valid_user) }
+        let!(:prepare) {
+          if $user.blank?
```
mjy: Please, no global references.

spec/features/tasks/collection_objects/filter_spec.rb
```diff
@@ -214,6 +214,95 @@
...
+            $user_id = pat.id
```
mjy: No.  Not allowed.

mjy:
@TuckerJD Added some things to tweak.  Most critically, thou shalt not use `$` constants, get rid of them please.

app/controllers/tasks/collection_objects/filter_controller.rb
TuckerJD: If not here, it complicates things a bit down line, but I fixed that.

app/controllers/collection_objects_controller.rb
TuckerJD: As far as I know, yes. That is how pagination is always done in the controllers, as far as I can tell.

app/controllers/tasks/gis/report_controller.rb
TuckerJD: Fixed.

app/helpers/tasks/collection_objects/area_and_date/area_and_date_helper.rb
TuckerJD: Done.

app/models/user.rb
TuckerJD: Done.

lib/housekeeping/timestamps.rb
TuckerJD: My bad. Originally, they were originally scopes, but I changed the way I was thinking about them. Do you think they should be scopes, or just methods?

mjy: If we want them to return ActiveRecord::Relation then they can remain scopes, but you'd have to change them to something like {order(created_at: :asc).limit(1)}.  I'm not sure how useful that scope would be.

In present form they should be methods, since they return an instance.

Note that I don't think you need 'all'.

app/helpers/tasks/collection_objects/area_and_date/area_and_date_helper.rb
TuckerJD: Do you want the default date format for the system set to `yyyy/mm/dd`?

lib/utilities/dates.rb
TuckerJD: Done.

spec/features/tasks/collection_objects/filter_spec.rb
TuckerJD: Fixed.

spec/features/tasks/collection_objects/filter_spec.rb
TuckerJD: Try this:
```ruby
let!(:joe) { 
    member = factory_girl_create_for_user(:valid_user, @user) 
    @project.users << member
    member
}
```
But I made do without the global reference.

TuckerJD:
I think I hit all the issues...

PR #295, opened 2017-12-15
[WIP] Browse sequences task

Merfoo:
This pull request is mainly to just get feedback to make sure the pattern is correct for developing tasks using VueJS. This task is to retrieve all sequences that are associated with a Gene Descriptor and display them in a table. It uses the `autocomplete` and `tableList` component for querying Gene Descriptors and displaying the sequences. A `pagedTableHeader` component was also made to add paging capabilities similar to how the list view works for data models e.g. Taxon Names.

Couple of questions I have
1) Currently when searching for sequences it creates an ajax call behind the scenes for the sequences, this leaves the url unchanged making it impossible to share the current page. Should I pursue modifying the url on each request so that sharing a link is possible?
2) I considered making a `pagedTable` component that encapsulated the `tableList` component in it with the `pagedTableHeader` component but ditched the idea as that would mean I would have to pass all the props from the `pagedTable` component to its child `tableList` component which would require a bunch of code just for passing those props. Should I have continued with the idea anyways?

### DIFF
--- a/app/controllers/tasks/sequences/browse_controller.rb
+++ b/app/controllers/tasks/sequences/browse_controller.rb
@@ -7,15 +7,11 @@ class Tasks::Sequences::BrowseController < ApplicationController
 
   # POST
   def sequences
-    sequences = []
     gene_id = params[:gene_id]
     gene = Descriptor.find_by(id: gene_id)
-
-    if gene
-      sequences = gene.sequences
-      puts gene
-    end
-
+    sequences = []
+    sequences = gene.sequences if gene
+    
     render :json => sequences
   end
 end
--- a/app/controllers/tasks/sequences/browse_controller.rb
+++ b/app/controllers/tasks/sequences/browse_controller.rb
@@ -1,0 +1,21 @@
+class Tasks::Sequences::BrowseController < ApplicationController
+  include TaskControllerConfiguration
+
+  # GET
+  def index
+  end
+
+  # POST
+  def sequences
+    sequences = []
+    gene_id = params[:gene_id]
+    gene = Descriptor.find_by(id: gene_id)
+
+    if gene
+      sequences = gene.sequences
+      puts gene
+    end
+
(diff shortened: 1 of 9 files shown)


jlpereira:
1- Not sure of what part of the code we're talking here. An example? You mean something like Edit Taxon Name Task? when you can open a taxon just passing the param from the url? or set it in the app after create a new one? Depends of the function you want for the task, could be useful pass directly. You can change the browse url doing this:

history.pushState(null, null, `/tasks/sequence/browse/sequences?gene_id=/${id}`);

2- Sounds good to me, in fact we're already talked with Matt about to add a pagination in other task, but we left it for the moment. Go ahead if you want to proceed in that way. In the case of making a component to encapsulate both, you need to pass the props to the child, and I imagine you can emit a event from the header component to make the father sets the page on the tableList. Other thing is you gonna need to emit a event from the tableList and emit again from pagedTable. This is in the case if you want to use the table to edit or delete, or do other stuff with the elements inside it. It will be like make a "chain" to send the events. Just remember we need to have the table with or without headers, but in this cases it won´t be a problem because we still can use the tableList separatly. If you want to simplified the things, you can touch tableList and add all what you need there. Feel free to make any change what you need in the code

Merfoo:
For clarification on question 1), if you go to `/taxon_names/list?page=2` the page that you're on is in the url as part of the requests data, but what happens in this task is that it loads all the sequences into an array and then slices that array to get only the elements that are wanted depending on what page you're on.

mjy:
@Merfoo Modifying the URL is nice, it reflects being able to enter the app from any point.  It may require updating the response.  

Is there anything else you need feedback on for this request?  If @jlpereira "understands" your code I'm happy to pull in the request.

Merfoo:
I'm going to close this pull request as there are still features I need to add. I will open a new pull request once those features are completed, thanks for the feedback.

PR #392, opened 2018-04-23
Uniquify people (modifing is_data, and Person)

TuckerJD:
- `::identical(attr)` and `#identical` return records matching all columns *not* removed because of `RESERVED_ATTRIBUTES`, and the constant from the specific class `IGNORE_IDENTICAL`. 
-- If the instance record contains (non-ignored) columns (key) which contain (value) `nil`, they will have to match a `nil` in a persisted records for it to be returned. 
--`#identical` *will not* return the instance itself. 
--`::identical(attr)` *will* return the record the attributes of which were used as parameter, _if_ it has been persisted. 

- `::similar(attr)` and `#similar` return records matching all columns *not* removed because of `RESERVED_ATTRIBUTES`, and the constant from the specific class `IGNORE_SIMILAR`. 
-- If the instance record contains (non-ignored) columns (key) which contain (value) `nil`, these columns will also be ignored, as if they has been included in `IGNORE_SIMILAR`. 
--`#similar` *will not* return the instance itself. 
--`::similar(attr)` *will* return the record the attributes of which were used as parameter, _if_ it has been persisted. This permits a new, non-persisted,  record to be used as a prototype for the attributes, requiring only those attributes which are of interest in a particular search.


app/models/concerns/shared/is_data.rb
```diff
@@ -47,31 +108,75 @@ def errors_excepting(*keys)
...
+  end
+
+  # @param [Symbol] keys
+  # @return [Array]
   def full_error_messages_excepting(*keys)
     errors_excepting(*keys).full_messages
   end
 
-  module ClassMethods
+  class Stripper
```
mjy: @TuckerJD just one request- make this a module, not a class (you have no attributes etc. that you need to track across instances).  The invocations will be Stripper.<method> (remove new) and the methods will need the def self.<method>.

Issue #532, opened 2018-09-27
Taxa browser - asserted distributions matching not showing?


mjy:
When I have an asserted distribution, for say "texas", and I draw a big polygon around texas, then it does not return the OTU associated with that OTU. 

@TuckerJD can you look into this.  I could have conceivably broken the filter with my changes to OTU filter.  Note that that code is basically forked into tow methods, I haven't reconciled your result from the new attributes.

TuckerJD:
This issue goes deeper than first imagined: the javascript has to be modified, because it **only** asks for CollectionObjects, although `selection_object` as a parameter is not actually used in the filter. Additionally, It should be possible to find both `CollectionObject`s and `AssertedDistribution`s through named GeographicArea, even though that area may not have a shape.

mjy:
Let's focus on the models/unit tests that let us return the data, then we'll get to the implementation.

You're right with 2 ways of finding things:

* Through the controlled vocabulary/parent-child (geographic_area#id -> children)
* Through the spatial calculations.

Focus on the latter, by default it should resolve most cases because most people will use GAs *with* GIs.

TuckerJD:
While testing this with the user interface, I have found what looks to be a performance issue WRT finding a small number of objects in one project, when another project has a very large population of objects. I am hoping that `project_id` can be accounted for earlier in the query chain. In addition, there seems to be a limit of 65535 items (in this case, `id`)  passable through to the query engine, and I have a (reasonable) example of this failing with no indication at the user interface.

mjy:
@TuckerJD Can we close this?

TuckerJD:
Given that we have not exhausted the ways in which we might search, and that the specific complaint has been resolved, I think so, yes.

PR #573, opened 2018-10-18
Add asserted distribution and collection object to geographic_area/shape search

TuckerJD:
Some progress, but there are still some other ways (currently unaccounted for) to find things.
See #532


app/controllers/otus_controller.rb
```diff
@@ -12,8 +12,8 @@ def index
         render '/shared/data/all/index'
       end
       format.json {
-        @otus = Queries::Otu::Filter.new(filter_params).all
-          .where(project_id: sessions_current_project_id).page(params[:page]).per(500)
+        @otus = Queries::Otu::Filter.new(filter_params.merge({project_id: sessions_current_project_id})).all
```
mjy: Not sure I like this pattern.  We need to come up with guidlines for whether filters should include project_id reference or not.

lib/queries/otu/filter.rb
```diff
@@ -135,8 +136,13 @@ def geographic_area_scope
       end
 
       gi_sql = GeographicItem.contained_by_where_sql(target_geographic_item_ids)
-      co_ids = CollectionObject.joins(:geographic_items).where(gi_sql).distinct.ids
-      ad_ids = AssertedDistribution.joins(:geographic_items).where(gi_sql).distinct.ids
+      if project_id.present?
+        co_ids = CollectionObject.with_project_id(project_id).joins(:geographic_items).where(gi_sql).distinct.ids
```
mjy: These are the real problem with speed and scoping, we aren't using joins, but rather arrays of ids.

Issue #813, opened 2019-02-25
Move to FactoryBot 5.0


mjy:
There are significant changes that break our present setup, these can likely be handled with a config file.

https://github.com/thoughtbot/factory_bot/blob/master/NEWS.md

LocoDelAssembly:
@mjy looks it was simpler to fix. Or did you want to re-arrange all the code to some new style?

In case it is all good please wait for Travis to process it before merging.

mjy:
Too late ;). Crossing fingers.

LocoDelAssembly:
Its OK, most likely nothing will burn. I tested locally before pushing. This is the expected set of errors:
```
Finished in 25 minutes 59 seconds (files took 16.03 seconds to load)
6383 examples, 5 failures

Failed examples:

rspec ./spec/features/authentication_spec.rb:25 # Authentication /signin when credentials do not match existing user should not sign user in
rspec ./spec/features/taxon_determinations_spec.rb:51 # TaxonDeterminations signed in as a user, with some records created testing new taxon determination can create a new taxon determination
rspec ./spec/features/type_materials_spec.rb:67 # TypeMaterials when signed in and a project is selected creating a new type_materials testing the new type_materials form filling out the form
rspec ./spec/features/users_spec.rb:276 # Users when user is not signed in forgotten password reset valid email request password reset
rspec ./spec/features/users_spec.rb:307 # Users when user is not signed in flagged for password reset requesting password reset from sign in process
```

mjy:
Those are all known and based on UI changes.  I should have them fixed today.

PR #865, opened 2019-03-07
Migration to FactoryBot 5.0 (#813)

LocoDelAssembly:
The breaking change in 5.0.0 was "Changed: use_parent_strategy now defaults to true, so by default the build strategy will build, rather than create associations".
Fixed by setting use_parent_strategy back to false.


mjy:
I concurr.  I thought I had tried this, but nope.  Thanks.

Issue #699, opened 2018-12-03
Batch load - GPX format 2 Collecting Event


mjy:
As a collector I want to translate my GPX data to collecting events with georeferences.

# Reference

* https://wiki.openstreetmap.org/wiki/GPX
* https://github.com/dougfales/gpx
* https://en.wikipedia.org/wiki/GPS_Exchange_Format
* https://www.topografix.com/gpx.asp

TuckerJD:
Some things to do:
- [ ] Isolate all the CSV processing and replace it with GPX processing.
- [ ] Decide how to process `gpx.waypoints`, `gpx.tracks`, gpx.`track_points`, etc.
- [ ] Decide if there is a way to save segments, distances, times (other than `create` times).
- [ ] What to do with `gpx.name`.

mjy:
* gpx.name == CollectingEvent#verbatim_label
* Times can likely go in/between CollectingEvent `time_` fields

TuckerJD:
Ok, change in direction: Looking into converting the GPX file to a TSV file so that the CSV file processing in class Import can be used without over-riding.

TuckerJD:
OK, new paradigm for processing GPX files: 

1. Only Waypoints, Routes, and Tracks are used (any first-level object which has a `name` attribute).
2. Each of the above objects will generate a separate `CollectingEvent`.
3. If the object has a time, it will be used to fill in start and stop times for the `CollectingEvent`.
4. If the object contains subobjects each of which contains a time, the first and last subobjects will determine the start and stop times for the `CollectingEvent`.

TuckerJD:
Closed accidentally.

TuckerJD:
Converting `file.gpx` into CSV data:

The class `Import` is designed to process a single file, which has one row of headers, and one or more self-consistent rows of tab-separated data entries.

Convertine a GPX file to CSV, on the other hand, will result in one *or more* CSV segments, one for each of the possible first-level objects which might be encountered (See comment above).

PR #888, opened 2019-03-25
Batchload gpx to collecting events

TuckerJD:
Basic GPX batch load, as described in Issue #699.


lib/batch_load/import/collecting_events/gpx_interpreter.rb
```diff
@@ -0,0 +1,96 @@
...
+      @ce_namespace = args.delete(:ce_namespace)
+      super(args)
+    end
+
+
+    # methode override for GPX processing which is quite different from CSV
+    # @return [Hash, nil]
+    def csv
+      gpx_file = GPX::GPXFile.new(gpx_file: @file.tempfile.path)
+      # @csv = Hash.from_xml(gpx.to_s)
```
mjy: @TuckerJD Can we clear this up as well- and other comment blocks in this PR?  THanks.

Issue #876, opened 2019-03-14
Lib/queries - reconcile related queries


mjy:
The methods in `lib/queries/collection_object_filter_query.rb` need to merged, where still needed, into `lib/queries/collection_object/filter.rb`.    Corresponding specs should be moved/added as needed.

PR #889, opened 2019-03-25
Reconcile related queries

TuckerJD:
Issue #876: `CollectionObject` queries `Filter` and `FilterQuery` have been merged into `Filter`, resulting in `#all` or `#result` methods. No AREL was required to reconcile the two different versions, but further streamlining may be possible.


lib/queries/collection_object/filter.rb
```diff
@@ -2,18 +2,175 @@ module Queries
   module CollectionObject
 
     # !! does not inherit from base query
-    class Filter 
-      # Boolean (TODO: reconcile with other use) 
-      attr_accessor :recent
+    # class SimpleFilter
```
mjy: @TuckerJD You can remove comment blocks here and elsewehre I think?

PR #903, opened 2019-04-03
Search for locality of collecting events

jrflood:
Find locality by filter, search (by map area or named area) or tag.
  Uses leaflet drawable/editable map. 
  Displays map of georeferences from search selector.
  Filter and tag selectors only compile into list currently.
Alpha/beta functionality.


mjy:
Closing in Lieu of new.

PR #904, opened 2019-04-03
Make Gandalf older

dshorthouse:
...so as to compete with longevity of Sarah Knauss. See https://en.wikipedia.org/wiki/List_of_the_verified_oldest_people.


mjy:
As if by magic! Waiting for the sorcery of Travis to be "recharged" before we pull this in.

PR #955, opened 2019-04-23
Parallel travis

LocoDelAssembly:
Changes:
* Parallelized test using five workers
* Clear assets and VCR caches on CRON event.
* Conditional assets precompilation based on set of specs the worker runs.
* Configured coveralls to merge all workers into single report
* Using native headless Firefox rather than emulated screen buffer.


.travis.yml
```diff
@@ -53,23 +54,18 @@ install:
...
-  - ls -alh public/packs-test 
 script:
-  - bundle exec rake assets:precompile
-# - bundle exec parallel_test spec/ -n 6 --only-group $TEST_GROUP --group-by filesize --type rspec
-  - bundle exec rspec
+  - .travis/spec_runner.sh
 notifications:
   email:
     recipients:
       - diapriid@gmail.com
```
mjy: Should we add others here? @LocoDelAssembly?

.travis/spec_runner.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
```
mjy: @LocoDelAssembly Is this code based on existing code somewhere "out there"?  Please reference in comments if so.  If not- very nice.

.travis.yml
LocoDelAssembly: @mjy I guess I could add myself there.
```suggestion
      - ellocodelassembler@gmail.com
```

.travis/spec_runner.sh
LocoDelAssembly: @mjy I believe that other searching for the correct bash syntax it was all mine.

PR #992, opened 2019-06-05
Taxon names api

LocoDelAssembly:
Includes taxon_names API endpoint to experiment. Please merge only if development won't be merged into master very soon.


app/controllers/taxon_names_controller.rb
```diff
@@ -18,6 +18,19 @@ def index
     end
   end
 
+  # GET /api/v1/taxon_names
```
mjy: @LocoDelAssembly Do you anticipate this collapsing into the default `show` and `index` when we get it pollished?

lib/queries/taxon_name/filter.rb
```diff
@@ -138,166 +137,81 @@ def nomenclature_code
...
-
-        a = o.alias('tfa_')
-        b = o.project(a[Arel.star]).from(a)
-
-        c = h.alias('tfa_r')
 
-        b = b.join(c, Arel::Nodes::OuterJoin)
-          .on(
-            a[:id].eq(c[:descendant_id])
+        ::TaxonName.where(
```
mjy: Since this is a merge clause I think this is OK

app/controllers/taxon_names_controller.rb
LocoDelAssembly: @mjy not so sure at this point. The API must have some stability whereas the regular index/show which are used internally for tasks, etc. might not as it could be adapted to better suit our internal needs and potentially break the API spec.

spec/lib/queries/taxon_name/filter_spec.rb
```diff
@@ -155,7 +184,33 @@
 
   specify '#validity 2' do
     query.validity = false 
-    expect(query.all.map(&:id)).to contain_exactly() 
+    expect(query.all.map(&:id)).to contain_exactly()
+  end
+
+  specify 'all filters combined' do
```
mjy: Thanks for this.

spec/requests/api/v1/rest/otus_spec.rb
```diff
@@ -6,29 +6,15 @@
 
   context 'otu/index' do
```
mjy: Are you thinking that this `rest` pattern will isolate "true" rest endpoints from "service" endpoints?

app/controllers/taxon_names_controller.rb
mjy: @LocoDelAssembly I think it's important that they converge and be the same.  Utlimately we balance the API to serve application development and general JSON requests.  If the API does not serve applications then it's unlikely to be useful for general calls.  I.e. I anticipate that service calls will be more useful than RESTful calls to many folks doing analyses.  Let's merge to experiment, but let's try get these sunk back together as soon as possible.  @jlpereira and I can work with the optimizations in respect to the interface changes required.  It's better todo this sooner than later, otherwise we're managing two (3?) different ways of exposing the data.

app/views/shared/data/all/_metadata.json.jbuilder
```diff
@@ -2,7 +2,7 @@ json.object_tag object_tag(object)
 json.object_label label_for(object)
 json.global_id object.persisted? ? object.to_global_id.to_s : nil
 json.base_class object.class.base_class.name
-json.url url_for(only_path: false, format: :json)
+#json.url url_for(only_path: false, format: :json)
```
LocoDelAssembly: @mjy I noticed that for /taxon_names.json it always shows the same URL (e.g. https://sandcastle.taxonworks.org/taxon_names.json). Here it is commented but in development it is not. Is it used somewhere? Could json.object_url be changed to show the full URL instead? (which would eventually vary if you call from API or internally).

spec/requests/api/v1/rest/otus_spec.rb
LocoDelAssembly: @mjy sorry I don't understand what you mean. Please note that OTU is not officially part of the API yet, I only refactored the spec to more or less follow the same pattern as taxon_names.

app/views/shared/data/all/_metadata.json.jbuilder
mjy: I'm pretty sure it is used, likely in an annotator.  It should dynamically generate the whole URL I think.  It might omit the base in development, is that what you are asking?

spec/requests/api/v1/rest/otus_spec.rb
mjy: @LocoDelAssembly Sorry for the confusion.  I didn't remember adding the `rest/` path.

app/views/shared/data/all/_metadata.json.jbuilder
LocoDelAssembly: @mjy this is what I mean is happening (look at url and object_url):
```JSON
{
    "id": 44517,
    "name": "rhodiensis",
    "parent_id": 44487,
    "cached": "Saga rhodiensis",
    "cached_html": "\u003ci\u003eSaga rhodiensis\u003c/i\u003e",
    "feminine_name": null,
    "masculine_name": null,
    "neuter_name": null,
    "cached_author_year": "Salfi, 1929",
    "etymology": null,
    "year_of_publication": null,
    "verbatim_author": null,
    "rank": "species",
    "rank_string": "NomenclaturalRank::Iczn::SpeciesGroup::Species",
    "type": "Protonym",
    "created_by_id": 5,
    "updated_by_id": 24,
    "project_id": 2,
    "cached_valid_taxon_name_id": 44517,
    "cached_original_combination": "Saga rhodiensis",
    "cached_original_combination_html": "\u003ci\u003eSaga rhodiensis\u003c/i\u003e",
    "cached_secondary_homonym": "Saga rhodiensis",
    "cached_primary_homonym": "Saga rhodiensis",
    "created_at": "2013-02-12T18:23:00.000Z",
    "updated_at": "2019-03-12T23:47:20.361Z",
    "nomenclatural_code": "iczn",
    "verbatim_name": null,
    "object_tag": "\u003ci\u003eSaga rhodiensis\u003c/i\u003e",
    "object_label": "Saga rhodiensis Salfi, 1929",
    "global_id": "gid://taxon-works/Protonym/44517",
    "base_class": "TaxonName",
    "url": "https://sandcastle.taxonworks.org/taxon_names.json",
    "object_url": "/taxon_names/44517",
    (OUTPUT OMITTED)
}
```

What I'm suggesting is that "object_url" be https://sandcastle.taxonworks.org/taxon_names/44517 and remove "url". However, I don't know if this won't break anything. In any case, taxon-name-api branch has "url" commented out, so it only outputs "object_url". Do I just revert the change?

LocoDelAssembly: (The JSON code comes from the non-API https://sandcastle.taxonworks.org/taxon_names.json?per=5 using Orthoptera project)

mjy: Revert for now.

PR #1157, opened 2019-11-01
Coldp

LocoDelAssembly:
delayed_job worker is set up as service and also ColDP downloads are handled asynchronously.


exe/delayed_job
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+set -e
+
+cd /app
+exec chpst -u app env QUEUE=coldp_export /usr/bin/bundle exec rails jobs:work
```
mjy: Does rails jobs:work take params?

LocoDelAssembly: I think it doesn't other that QUEUE(S) as env var.

https://www.rubydoc.info/gems/delayed_job/4.1.8
> Rails 4: replace script/delayed_job with bin/delayed_job
> 
> Workers can be running on any computer, as long as they have access to the database and their clock is in sync. Keep in mind that each worker will check the database at least every 5 seconds.
> 
> You can also invoke rake jobs:work which will start working off jobs. You can cancel the rake task with CTRL-C.
> 
> If you want to just run all available jobs and exit you can use rake jobs:workoff
> 
> Work off queues by setting the QUEUE or QUEUES environment variable.
> 
> QUEUE=tracking rake jobs:work
> QUEUES=mailers,tasks rake jobs:work
>

PR #1311, opened 2020-02-11
Biodiversity4

LocoDelAssembly:
Use newer Go-based Biodiversity Parser. Result structure changed since previous mayor version. Significantly faster without the aid of Redis plus likely other improvements and bug fixes.


lib/batch_load/import/dwca.rb
```diff
@@ -727,29 +727,29 @@ def make_tn(row)
...
-            species        = snp[:scientificName][:details][0][:species]
+            species        = snp[:details][0][:specificEpithet]
             t_n.parent     = @genus
             t_n.rank_class = Ranks.lookup(:iczn, :species)
-            t_n.name       = species[:string]
+            t_n.name       = species[:value]
             # t_n.cached_author_year = snp[:scientificName][:details][0][:species][:authorship]
-            t_n.year_of_publication = species[:basionymAuthorTeam][:year].to_i
-            author_name             = species[:basionymAuthorTeam][:authorTeam]
+            t_n.year_of_publication = species[:authorship][:basionymAuthorship][:year][:value].to_i
```
mjy: I think we need to allow for nil?

```
nirb(main):001:0> nil.to_i
=> 0
```

LocoDelAssembly: @mjy With new Biodiversity gem I think it cannot happen that :year exists and :value is nil. However I changed the code to use dig and parse the year differently so nil or "" stays nil. Thanks :+1:

PR #1324, opened 2020-02-24
Imagemagick 7 with HEIC support enabled

LocoDelAssembly:
Just issuing so Travis builds (but not publishes) the docker image.


LocoDelAssembly:
Somewhat serious problem with HEIC support is that browsers don't actually support it, not even on MacOS. Tried FF, Chrome and Safari (screenshot is from Safari), all of them when showing original are not able to display it or trigger binary file download. What do you recommend doing here @mjy, @jlpereira?

image

mjy:
Not showing the original is fine and anticipated, in fact we should likely never try to do this but rather show the full size web/derived version.

LocoDelAssembly:
Problem is that derived versions look too tiny. Original file was 3992x2992, but the web medium version is only 300x225 and web thumb 100×75.

I think several views just display the original file when something larger than web thumb is needed. So what is actually needed is tweak the `styles` attachment attribute to generate extra sizes and/or change current sizes and regenerate everything? On some cases (like Comprehensive task), scale_to_box is used which seems to work fine:

image

Would make sense to go ahead and merge this branch anyway and solve HEIC-related problems later? Any luck remembering the issue that originated this requirement?

mjy:
Don't worry about the derived sizes, we need to completely re-do all that.  That's  a different issue that I think we talked about waiting to use the new Rails library for... which we could do now (but I think after DwC lands).

mjy:
Yes, definitely merge btw.

PR #1490, opened 2020-05-12
Global names finder playground task

mjy:


mjy:
Agreed.  There are almost certainly broken corners in this task, but should have low impact overall.  Merging.

PR #1697, opened 2020-09-02
Biodiversity ipc mode

LocoDelAssembly:
Official biodiversity gem is causing TW to freeze and therefore k8s pods to be restarted. This custom gem uses a separate process to run gnparser, should no longer freeze TW.

When merged remember to add entries to CHANGELOG. Let at least Travis CI pass before merging.


LocoDelAssembly:
Bad merge. Aborting....

Issue #1866, opened 2020-11-01
Model - NomenclaturalRankOrder


mjy:
It feels like we need to cache NOMEN order in a table to address a lot of sort/ordering outside of SQL.   Something like this:

* [ ] Add a migration to build the table 
```
table :nomenclatural_rank_order
:rank_class, :string
:position, :integer # just start each code at the next Int, don't start each from 1
```
* [ ] Build a way to re-generate the table based on inspection of `NomenclaturalRank.descendants`
* [ ] Add a data migration that triggers the population of the table so that test databases include an index on build, not on spec run
* [ ] Add a deploy hook so that `nomenclatural_rank_orders` <- note plural is wiped and rebuilt _on every deploy_
* [ ] Add `has_one :nomenclatural_rank_class, foreign_key: :rank_class ...` to TaxonName
* [ ] Add `scope :order_by_nomenclatural_rank` to TaxonName:
```
  joins(:nomenclatural_rank_class).order(nomenclatural_rank_class: :position)
```

Start recording `.sort` methods by just including the scope.  Specs should catch that things are still OK.

mjy:
Need to also ensure that test database does *not* wipe the table between runs (it shouldn't I think, but I'd have to think about this).

LocoDelAssembly:
https://www.postgresql.org/docs/10/datatype-enum.html#id-1.5.7.15.6 <- Almost, but `DETAIL:  Labels must be 63 characters or less.`...

Thinking a bit more. Do you see problems doing this like https://stackoverflow.com/a/23620658 ?

mjy:
Interesting re ENUM.

I did think about about the `CASE` solution.  Things that I don't like, but they're not necessarily game breakers:
1) it will be called a lot, and there are 40(?) +  case checks, this might expand logs a lot
2) Unsure about performance?  Is CASE on par with a clean join?
3) We'd still need to generate code every deploy, to reflect possible changes in the models, so there is no win there.

LocoDelAssembly:
>2. Unsure about performance? Is CASE on par with a clean join?

Unsure as well, but sounds a bit more expensive to deal with an extra table than doing comparisons in-place.

>3. We'd still need to generate code every deploy, to reflect possible changes in the models, so there is no win there.

[...]

PR #1867, opened 2020-11-02
Added order_by_nomenclatural_rank scope (refs #1866)

LocoDelAssembly:
DO NOT MERGE


app/models/taxon_name.rb
```diff
@@ -1576,4 +1580,11 @@ def sv_hybrid_name_relationships
     true # see validation in Hybrid.rb
   end
 
+  private
+
+  def self.case_rank_class_to_position
+    # DO NOT REMOVE .asc (.last does not change default ASC to DESC if removed)
+    @@case_rank_class_to_position ||= RANKS.each_with_index.inject(arel_table[:rank_class]) { |a, (r, i)| a = a.when(r).then(i) }.asc
```
mjy, line 1587: Can you use a frozen constant rather than a @@?

LocoDelAssembly:
Still failing even when populating at migration. Now realized it doesn't work on local setup first try, but subsequent calls to the same spec work. I don't understand this... I even added populate_db on migration and still nothing.
```
rails db:drop && rails db:create && rails db:migrate RAILS_ENV=test && rspec ./spec/models/taxon_name_spec.rb:638 || \
  rspec ./spec/models/taxon_name_spec.rb:638 && rspec ./spec/models/taxon_name_spec.rb:638
```
image

PR #2177, opened 2021-04-10
Fix invalid parent problem

LocoDelAssembly:
- Get the closest ancestor without superspecies and supersuperspecies ranks and assign as parent.
- When duplicate OTUs, pick the one with Otu.name=nil (if only one is nil).

https://github.com/CatalogueOfLife/testing/issues/87


mjy:
@LocoDelAssembly Thanks much for the tweaks. We are probably at the point where we should add some specs that describe specific scenarious we expect to see.  This will prevent, in theory, wac-a-mole as we converge our model with CoLDP.  This will also help when we optimize, for example I'd like to ultimately return the target parent id in a single SQL statement if possible.

LocoDelAssembly:
@gdower @mjy pushed extra changes. When testing new sandcastle data remember this is not merged. With last commit the default is to not resolve OTUs conflicts to maintain original behavior, make sure you mark the checkbox when producing exports from sandcastle data.

If you think it is good to go please merge.

PR #2223, opened 2021-05-03
Redirect back on destroy with override to :index when needed

LocoDelAssembly:


LocoDelAssembly:
@mjy do you think this is ready to merge?

PR #2237, opened 2021-05-10
Upgrade rgeo-proj4 to next major version

LocoDelAssembly:


LocoDelAssembly:
Turned out that the spec was probably faulty and working right because older gem tolerated the point being considered included in the polygon when in reality it was intersected by one edge of the polygon (didn't realize initially that the polygon is a triangle and not a square).

@mjy can you think of any reason to consider this not safe to marge?

config/application.rb
```diff
@@ -64,7 +64,7 @@ class Application < Rails::Application
       config.default = RGeo::Geographic.projected_factory(
         srid:                    4326,
         projection_srid:         4326,
-        projection_proj4:        '+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs',
```
mjy, line 67: @LocoDelAssembly Only this makes me hesistate on this pull.  What's the difference b/w this and the proposed?

LocoDelAssembly: That one is not recognized by the new gem version, but 'EPSG:4326' is supposed to be the same (or very close): https://spatialreference.org/ref/epsg/wgs-84/ (click on Proj4 link)

LocoDelAssembly: Extra info: Original code causes multiple 'proj_crs_get_geodetic_crs: Object is not a CRS' at boot and also, the factory is not actually instantiated so then lots and lots of undefined method for nil:NilClass.

mjy:
@LocoDelAssembly are there any "Z" considerations, i.e. IIRC we handled altitude in the prior model, is that the case here too? "Scope: Horizontal component of 3D system. Used by the GPS satellite navigation system and for NATO military geodetic surveying."  I don't remember if that was a factor or not in the original model.

LocoDelAssembly:
@mjy looks we have 4326 everywhere (including DB schema), and not 4979 which would be 3D. Proj4 string seems to be the same with either: https://spatialreference.org/ref/epsg/4979/proj4/ https://spatialreference.org/ref/epsg/4326/proj4/

mjy:
Merge.

PR #2274, opened 2021-05-25
2242 taxon name validity

mjy:
# Added
* TaxonName `cached_is_valid` boolean, takes into account Relationships _and_ Classifications

# Fixed
* CoL Data Package scoping

# Changed
* `that_is_valid` scope now references `cached_is_valid`
* `that_is_invalid` scope now references `cached_is_valid`
* `calculated_valid` replaces `that_is_valid`
* `calculated_invalid` replaces `that_is_invalid`
* remove unused `TaxonName#cached_higher_classification`


app/models/combination.rb
```diff
@@ -449,6 +449,7 @@ def sv_combination_duplicates
   def sv_cached_names
     is_cached = true
     is_cached = false if cached_author_year != get_author_and_year
+    is_cached = false if cached_is_valid || cached_is_valid.nil?
```
LocoDelAssembly, line 452: Not sure logic here. Looks it will always execute?

app/models/protonym.rb
```diff
@@ -198,23 +198,31 @@ def list_of_coordinated_names
...
-            where('name in (?)', search_name)
+#        r = TaxonNameRelationship.where_subject_is_taxon_name(self).with_type_array(TAXON_NAME_RELATIONSHIP_NAMES_MISSPELLING)
+#        if !search_name.nil? && r.empty?
+        if !search_name.nil? && is_available?
+          list = Protonym
+            .that_is_valid
+            .ancestors_and_descendants_of(self)
+            .not_self(self)
+            .with_rank_class_including(search_rank)
+            .where('name in (?)', search_name)
```
LocoDelAssembly, line 225: `.where(name: search_name)` should do it and would work for both simple values and arrays.

lib/export/coldp/files/synonym.rb
```diff
@@ -31,29 +31,34 @@ def self.generate(otus, reference_csv = nil)
...
           #   2) of 1) eliminate Combinations with identical names to current placement
-          TaxonName
-            .where(cached_valid_taxon_name_id: o[2]) # == .historical_taxon_names
-            .where("( ((taxon_names.id != taxon_names.cached_valid_taxon_name_id) OR ((taxon_names.cached_original_combination != taxon_names.cached))) AND NOT (taxon_names.type = 'Combination' AND taxon_names.cached = ?))", o[1]) # see name.rb
-            .pluck(:id, :cached, :cached_original_combination, :type)
-            .each do |t|
+          #
+          a = TaxonName.that_is_invalid
+            .where(cached_valid_taxon_name_id: o[2])
+            .where.not("(taxon_names.type = 'Combination' AND taxon_names.cached = ?)", o[1])
```
LocoDelAssembly, line 47: mmmm, probably too expensive but here it goes anyway:
```ruby
2.7.3 :011 > puts TaxonName.where(cached_valid_taxon_name_id: 42).merge(TaxonName.where.not(type: 'Combination').or(TaxonName.where.not(cached: 'Forty two'))).to_sql
SELECT "taxon_names".* FROM "taxon_names" WHERE "taxon_names"."cached_valid_taxon_name_id" = 42 AND ("taxon_names"."type" != 'Combination' OR "taxon_names"."cached" != 'Forty two')
 => nil
```

Issue #1486, opened 2020-05-07
User - add person_id


mjy:
As a user I want to assign myself an ORCID by proxy of the data I am curating so that I can proxy myself as data.  

Solution: Add person_id to User, NOT add Identifiers to User.

Issue #1467, opened 2020-04-29
Task - DWC Health


mjy:
As a curator I want a summary of the status of my DWC index.

# Delivery
* [ ] Check for presence of Download

# Data meta
* [ ] Check for Kingdom in classification
* [ ] Simple "staleness" graph
* [ ] Objects/indecies

tmcelrath:
Data meta:
- collection objects w/o identifiers (not including containerized specimens) Report
- collection objects w/o Collecting Events Report
- collection objects w/o ANYTHING

A lot of the reports from Collection Report Task would be good here for health checks.

mjy:
@jlpereira TODO list here:

* [ ] Change columns in Recent downloads to: `created_at`, `expires`, `total_records`, `times_downloaded`, `is_public`, show.  
* [ ] Make `is_public` a green toggle that touches `PATCH /downloads/123` to update this boolean so we can turn them on/off
* [ ]  Add text below "Download", and include preview count (get from header, for page size 1) for each button.
<img width="245" alt="Screen Shot 2021-09-12 at 1 14 45 PM" src="https://user-images.githubusercontent.com/22600/132998396-6b832617-5d0a-4b01-8963-fa10461bed7a.png">

mjy:
* [ ] Add link from Collection Object filter to DwC Index rebuilder - enable only when < 5k records.

Issue #1303, opened 2020-02-06
DWC Downloads - Save as Download (model)


mjy:
When I request a DWC download (preview page), then I use the Async Download to register it as a file first.

Issue #1269, opened 2020-01-03
DwC index - new fields (was: Not enough displayed/exported fields in "Collection objects filter")


tmcelrath:
When I do a search in "Collection objects filter" I want to see more fields associated with that Collection Object. 

All fields needed are mentioned below in the comments. [EDITED TCM]

@tmcelrath - are these actually dwc fields?  Links?

Missing fields:
<strike>"Buffered" fields e.g. "verbatimDetermination" "verbatimLocalityLabel" "verbatimOther"
* [ ] verbatimDetermination
<strike>* [ ] verbatimLocalityLabel already exists - verbatim_locality from CE, not buffered</strike> 
* [ ] verbatimOther</strike>

<strike> Determination fields: "determinationOTU (or whatever it's called) and associated "determinedBy" and "determinationYear" </strike>

@tmcelrath - below are drawn from TaxonDetermination, only first, or all or what todo?

<strike>* [ ] determinedBy
* [ ] determinationYear
</strike>

<strike>Also, they should be displayed by type of field, not alphabetically. E.g. all "Locality" fields should be logically displayed e.g. "Country" then "stateProvince" then "county" then "decimalLongitude" etc ... and then "Family" then "Genus" then "specific epithet".
</strike> See #1314.

All these fields also need to be exported as well as displayed.

tmcelrath:
Also need to search by custom "Internal Attribute" fields such as "SpecialCollection"

mjy:
If by search you mean in the filter vs. scanning a table then add that internal attribute as a new issue.

tmcelrath:
Right - yeah, got the issues conflated.

tmcelrath:
You mean #1258? :)

mjy:
@tmcelrath If you can help out by finding the DWC terms that you want each of the check boxes to fill it would be appreciated.  Can you double check that this is the list you have in mind?

tmcelrath:
@mjy working on it - is "recordedBy" being properly parsed? Just updated a record to have two collectors and neither is exported in the "recordedBy" field.

tmcelrath:
Need DWCA Export fields from Class Identification (http://rs.tdwg.org/dwc/terms/Identification):
typeStatus: http://rs.tdwg.org/dwc/terms/typeStatus (from Type material)
identifiedBy: http://rs.tdwg.org/dwc/terms/identifiedBy (from Determinations)
[...]

PR #2300, opened 2021-06-07
DwC Export, new fields, refactor etc.

mjy:
### Added
- Updated framework for producing and downloading DwC Archives (DwCA) [#1775] [#1303]
- Increased from 21 to 53 the number of fields referenced in the (DwCA) dump [#1269] [#1230]
- Task `DwC Dashboard` facilitating DwCA download, metadata reporting, and "health" checks [#1467]
- Endpoint `/tasks/dwc/dashboard/index_versions` returns the dates at which DwcOccurrence indexing was modified. !! TODO: update date of merge.
- Endpoint `/dwc_occurrences/metadata`, for stats on the state of DwcOccurrence index 
- Endpoint `/dwc_occurrencs/predicates` to return a list of Predicates used on CollectionObjects and CollectingEvents
- Endpoint `/dwc_occurrences/status` to check whether DwcOccurrence records are up-to-date
- Endpoint `/dwc_occurrences/collector_id_metadata` to check whether People referenced in DwcOccurences have GUIDs
- Task on Administration panel, "Reindex", with (temporary) options to re-index small blocks of DwcOccurrence records 
- Button on CollectionObject filter to download filter result as DwC Archive [#1303]
- User can select a corresponding Person as their data representation (facilitates Identifiers for Users) [#1486]
- Wikidata (Q) and ORCiD support for people references in DwCA dumps
- New fields `identifiedByID` and `recordedByID` on DwcOccurrence
- Centroid index on GeographicItem
- Field `total_records` on Download
- Index on polymorphic fields of DwcIndex (e.g. faster queries to CollectionObject)
- Index on `data_origin` for GeographicAreasGeographicItem
- `DwcOccurrence` controller (metadata use only)
- Identifiers for AssertedDistributions
- Various relationships enabling the joining of DwcOccurrence directly to other classes of data (e.g. Georeferences)
- Isolated Georeference related utilities into their own module CollectingEvent::Georeference
- A Taxonomy module that caches classification values, used in CollectionObject, and Otu
- Methods to return when a record attribute was updated (e.g. verbatim_locality changed), and who did it for Papertrail including classes of data
- Methods to handle multiple classes of globally unique identifiers on DwcOccurrence records
- Auto-generation of UUIDs for instances that don't have global identifiers during DwCOccurrence record building [#2186]
- Pattern for isolating modules that aid DwC serialization per class of data
- Optimized `to_wkt` to quickly return well-known-text for geo-shapes (in part, [#2526])
- Georeferences can have Confidences assigned to them
- New subclass of UUID `Identifier::Global::Uuid::TaxonworksDwcOccurrence`
- Clarified, via`georeferenceSources` and `georefernceProtocol` why there are many decimal points in DwC latitude/longitude referencing fields [#915] [#1175]

### Changed
- Unified various DwC value returning methods in their own explicitly named extensions
- Isolated CollectionObject filter and API param handling to their own module for reference in multiple controllers
- DwcOccurrence `individualCount` is now Integer
- Database ConnectionTimeoutErrors now result in a 503 response rather than a raise/email-warning
- Added various `:inverse_of` across collection objects related models
- `DwcOccurrence#individualCount` is integer now
- Simplified SQL for ordering GeographicArea shapes
- Tweak validation of ORCIDid format

[#915]: https://github.com/SpeciesFileGroup/taxonworks/issues/915
[#1175]: https://github.com/SpeciesFileGroup/taxonworks/issues/1175
[#1230]: https://github.com/SpeciesFileGroup/taxonworks/issues/1230
[#1269]: https://github.com/SpeciesFileGroup/taxonworks/issues/1269
[#1303]: https://github.com/SpeciesFileGroup/taxonworks/issues/1303
[#1467]: https://github.com/SpeciesFileGroup/taxonworks/issues/1467
[#1486]: https://github.com/SpeciesFileGroup/taxonworks/issues/1486
[#1775]: https://github.com/SpeciesFileGroup/taxonworks/issues/1775
[#2186]: https://github.com/SpeciesFileGroup/taxonworks/issues/2186
[#2526]: https://github.com/SpeciesFileGroup/taxonworks/issues/2526


app/helpers/dwc_occurrences_helper.rb
```diff
@@ -0,0 +1,41 @@
...
+    # without occurrences!
+    #
+
+    a = {
+      data: {
+        total_collection_objects: data.count,
+        enumerated_objects: data.sum(:total),
+        freshness: {
+          never: data.left_outer_joins(:dwc_occurrence).where(dwc_occurrences: {id: nil}).count,
+          one_day: data.where("updated_at > ?", 1.day.ago).count,
```
LocoDelAssembly, line 21: On this one but also those having similar pattern: `.where(updated_at: 1.day.ago..)`

```ruby
> puts CollectionObject.where(project_id: 42).where(updated_at: 1.day.ago..).to_sql
SELECT "collection_objects".* FROM "collection_objects" WHERE "collection_objects"."project_id" = 42 AND "collection_objects"."updated_at" >= '2021-06-06 15:49:41.657006'
```

*(`>=` instead of `>` but given the time resolution this hardly makes a difference)*

app/models/biocuration_classification.rb
```diff
@@ -24,7 +24,7 @@ class BiocurationClassification < ApplicationRecord
   acts_as_list scope: [:biological_collection_object_id, :project_id]
 
   belongs_to :biocuration_class, inverse_of: :biocuration_classifications
-  belongs_to :biological_collection_object, class_name: 'CollectionObject::BiologicalCollectionObject', inverse_of: :biocuration_classifications
+  belongs_to :biological_collection_object, class_name: '::CollectionObject::BiologicalCollectionObject', inverse_of: :biocuration_classifications
```
LocoDelAssembly, line 27: Was `::` added by mistake or is actually required?

LocoDelAssembly: Tried on console. Looks harmless.

mjy: This is indeed necessary for namespacing some of the modules that have identical names.  This says "from the root, find the module", rather than "from the relative position of this <method>, find the module".

LocoDelAssembly: I was fearing that for `type` columns it would attempt to match with leading `::` but I looks it is smart enough to not do that.

mjy: Right.  It took me a while to figure it out when the problem came, but it's definitely tested/working pattern.

LordFlashmeow:
Can we synchronize this with the development branch?

mjy:
Will do today.

LocoDelAssembly:
Historical problem but could produced data filename extension be `.tsv` or `.txt`? Currently it is `.csv` and Ruby's csv module is involved in the process, but columns separator is set to `\t` which no longer makes it coma-separated.

LocoDelAssembly:
The active jobs this PR introduced need to change the amount of work done per job. Processing entire scopes will block the worker until done. This is more noticeable on sandboxes which run without replicas, but in production it could also become problematic after 5 jobs are running concurrently.

The importer does this when running asynchronously when importing (maybe not when staging but in general it is not a too lengthy operation).

We may also consider alter our production deployment to have workers dedicated to specific queues, but even with this it is not a good idea to have very long running jobs.

mjy:
> Historical problem but could produced data filename extension be `.tsv` or `.txt`? Currently it is `.csv` and Ruby's csv module is involved in the process, but columns separator is set to `\t` which no longer makes it coma-separated.

"c" for "character"  I think, though many (most?) treat it as comma - https://www.ibm.com/docs/en/mufz/4.1?topic=files-character-separated-value-csv-reports

mjy:
> The active jobs this PR introduced need to change the amount of work done per job. Processing entire scopes will block the worker until done. This is more noticeable on sandboxes which run without replicas, but in production it could also become problematic after 5 jobs are running concurrently.

Are you thinking of both job types, or one in particular? I'm not sure of the best way forward here.

* In theory the download can be much much faster if we write faster empty column detection and add a couple more tweaks (it's just downloading a single table).  I started to work on this but didn't resolve in it in SQL instead of Ruby.  We could also shell out and manipulate the file that way if needed.  There are other things we can do as well, but download generation in general shouldn't be an issue once indexing is done in the background.
* I'm not sure how we'd go about composing the download in small jobs.  I suppose we could break composition into 2k records at a time, and know the job is done when totals add up.
* We could limit the re-index capabilities greatly.  We'll re-index the whole dataset in the background after deploying version updates, so these aren't going to be super important, they are more just to transition us for small sets of records.  We could limit them to small runs.  Not sure if this what you're getting at.
* We can speed re-indexing for large shapes by 3/4 at least I think.  Though this won't result in short times for large-re-indexing, but it's something we know we can do.

LocoDelAssembly:
image

I pressed `[10K]` there more than an hour ago **and also yesterday** (sandwich sandbox). I also uploaded a dataset and still not staging. Does the indexer have a way to tell when refresh is needed or it always redoes everything?

Download generation is probably fine, I think it always completed relatively fast.

mjy:
> Does the indexer have a way to tell when refresh is needed or it always redoes everything?

It does the all again, those buttons were largely for testing purposes.  We'll isolate them I think, and limit the re-indexing to datasets smaller than some filter size.  

It takes 0.5-8s to re-index a record.  Not much has been done to optimize.  The big changes, as noted, will be deciding to use centroids on large areas or not.

PR #2347, opened 2021-07-15
Strip namespace delimiter when parsing catalogNumber from DwC import

LordFlashmeow:
When extracting the identifier from the catalogNumber in the DwC import, the prefix isn't removed. As a result, a catalogNumber like `MCZ.14376w` is parsed into an identifier of `.14376w` and displayed as `MCZ..14376w`.

I recognize there may be an edge case when the delimiter is set to `NONE`, but I don't think it'll be a problem.

### DIFF
--- a/app/models/dataset_record/darwin_core/occurrence.rb
+++ b/app/models/dataset_record/darwin_core/occurrence.rb
@@ -73,7 +73,7 @@ class DatasetRecord::DarwinCore::Occurrence < DatasetRecord::DarwinCore
 
         if attributes[:catalog_number]
           namespace = attributes.dig(:catalog_number, :namespace)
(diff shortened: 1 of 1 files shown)


LocoDelAssembly:
`delete_prefix!` returns `nil` if nothing is changed:
```ruby
"hello".delete_prefix!("hel") #=> "lo"
"hello".delete_prefix!("llo") #=> nil
```
This means the code will not remove the delimiter if the namespace string was not stripped off. 

Alternatively this could be done this way:
```ruby
attributes.dig(:catalog_number, :identifier)&.tap { |i| i.delete_prefix!(namespace.verbatim_short_name || namespace.short_name) }&.delete_prefix!(namespace.delimiter)
```

Please let me know what would be best, if your code is already doing what you intend to do I'll merge as-is.

Thanks

PR #2377, opened 2021-07-29
Create type material simple batch importer

LordFlashmeow:
Most of the code is borrowed from `Import::Namespaces::SimpleInterpreter`, but it gets the job done.

This is a temporary solution (famous last words!) until it can be implemented with the new staging importer.

The importer defaults to using ICZN nomenclature, and I don't know if it's worth implementing a radio button to change it.

There's very little hard validation in the `TypeMaterial` class, so importing records multiple times will create duplicates.

I'm leaving this as a draft until `TypeMaterial` has better validation, since creating duplicates is a big problem.


mjy:
@LordFlashmeow thanks much.  I completely concur that we should have much strong validation on TypeMaterial, I've argued for this for a while.  @proceps can we add better constraints so that data incoming contain fewer problems?

proceps:
@mjy could you define the issues? uniqueness of type?

proceps:
in most cases when I do uploader I either use .find_or_create construction, which finds existing record, if it was imported before; or create! which hard fails if validation does not pass.

mjy:
Hard validations:
* `sv_single_primary_type` should be hard is the big one, this really needs to be hard coded with curators notes on conflicts potential issues etc. as annotations
* require para -> downstream relationships be in place, etc.

LordFlashmeow:
Closing — no desire for this.

PR #2378, opened 2021-07-29
Fix TypeMaterial legal_type_type validation for ICN

LordFlashmeow:
`legal_type_type` was validating against the ICZN types, even when the code was set to ICN.


LocoDelAssembly:
Not sure why I thought I asked you @proceps to review if there is something else to look after (or was in an issue?). Merging this now as it clearly improves the situation, please take a look if you see something else that should be done.

proceps:
@LocoDelAssembly I believe this is it, could be just a typo, or copy/paste issue.

PR #2381, opened 2021-08-03
Fix missing namespace labels in dwc importer

LordFlashmeow:


LocoDelAssembly:
Thanks @LordFlashmeow. Original code right before pushing was done with several `if`s and `return`s, then realized I could keep the ternary operator and accidentally removed the `return`.

PR #2400, opened 2021-08-09
Don't show catalogNumber with stripped namespace in table after import

LordFlashmeow:
Updating `attributes[:catalog_number][:identifier] in-place would cause the value in the table to be modified since it's the same string. Cloning the string prevents modifications from changing the table value.


LocoDelAssembly:
@LordFlashmeow I moved `clone` into `get_field_value` since returned value is not supposed to allow altering internal state and should always return the same value (unless the data field is explicitly set to a different value).

PR #2405, opened 2021-08-10
Extract namespace prefix deletion to method

LordFlashmeow:
We're removing the prefix from an identifier in two separate places, so I think we should convert it to a method

### DIFF
--- a/app/models/dataset_record/darwin_core/occurrence.rb
+++ b/app/models/dataset_record/darwin_core/occurrence.rb
@@ -102,7 +102,7 @@ class DatasetRecord::DarwinCore::Occurrence < DatasetRecord::DarwinCore
 
         if attributes[:catalog_number]
           namespace = attributes.dig(:catalog_number, :namespace)
-          delete_namespace_prefix(attributes.dig(:catalog_number, :identifier))
+          delete_namespace_prefix(attributes.dig(:catalog_number, :identifier), namespace)
 
           identifier = Identifier::Local::CatalogNumber
                       .create_with(identifier_object: specimen)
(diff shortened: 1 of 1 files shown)


app/models/dataset_record/darwin_core/occurrence.rb
```diff
@@ -298,6 +298,16 @@ def clear_empty_sub_hashes(hsh)
     hsh
   end
 
+  # Remove the namespace short name and delimiter from start of string.
+  #
+  # If the namespace has a verbatim_short_name, that is removed instead of the short_name.
+  # The delimiter is only removed if the short_name was found in the identifier.
+  # @param [String] identifier_str
+  # @param [Namespace] namespace
+  def delete_namespace_prefix(identifier_str, namespace)
```
LocoDelAssembly, line 307: Please rename to `delete_namespace_prefix!` to indicate this method alters supplied `identified_str` rather than returning a modified copy.

app/models/dataset_record/darwin_core/occurrence.rb
```diff
@@ -153,7 +153,7 @@ def import(dwc_data_attributes = {})
 
             identifier_attributes[:namespace] = namespace
 
-            event_id&.delete_prefix!(namespace.verbatim_short_name || namespace.short_name)&.delete_prefix!(namespace.delimiter)
+            delete_namespace_prefix(event_id, namespace)
```
LocoDelAssembly, line 156: delete_namespace_prefix!

app/models/dataset_record/darwin_core/occurrence.rb
```diff
@@ -102,7 +102,7 @@ def import(dwc_data_attributes = {})
 
         if attributes[:catalog_number]
           namespace = attributes.dig(:catalog_number, :namespace)
-          attributes.dig(:catalog_number, :identifier)&.delete_prefix!(namespace.verbatim_short_name || namespace.short_name)&.delete_prefix!(namespace.delimiter || '') if namespace
+          delete_namespace_prefix(attributes.dig(:catalog_number, :identifier), namespace)
```
LocoDelAssembly, line 105: delete_namespace_prefix!

PR #2413, opened 2021-08-12
Handle abbreviated ISO8601 dates (resolves #2406) 

### LINKED ISSUE #2406: DWCA Importer - eventDate import help and recommendations conflicting
When I am formatting a date range, the help (show below from DWCA) recommends that event date ranges are formatted like so:
XXXX-XX-XX/XX e.g. 2015-10-05/07
image
image
As shown in TW help: 
image

But, when I format an eventDate range like that, the date is errored and recommends a different format, XXXX-XX-XX/XXXX-XX-XX, e.g. 2015-10-15/2015-10-17:
image

We need to pick one or the other and stick to it.

LordFlashmeow:


LordFlashmeow:
Hmm, this can't handle inputs like 2015-10/11. Might need to do some more rewriting.

LocoDelAssembly:
> Hmm, this can't handle inputs like 2015-10/11. Might need to do some more rewriting.

What would you translate that into?

LordFlashmeow:
Sometime during October and November 2015, I think.

LocoDelAssembly:
But I mean concrete values, how either the OpenStruct object or CollectingEvent attributes would look like.

====================

(Also reposting a misplaced comment here)
Also make sure this one works, originally `parse_iso_date(:eventDate)` returned empty array (or `nil`?).
```ruby
2.7.4 :057 > parse_iso_date_str(nil)
Traceback (most recent call last):
        3: from (irb):56
        2: from (irb):57:in `rescue in irb_binding'
        1: from (irb):3:in `parse_iso_date_str'
NoMethodError (undefined method `with_index' for nil:NilClass)
```

LordFlashmeow:
It would be like`[#<OpenStruct year=2015, month=10, day=nil, hour=nil, minute=nil, second=nil>, #<OpenStruct year=2015, month=11, day=nil, hour=nil, minute=nil, second=nil]`

LordFlashmeow:
CollectingEvents can handle empty start and end days (or even months)

PR #2419, opened 2021-08-14
Fix rubocop config

LordFlashmeow:


LordFlashmeow:
How the heck does this decrease coverage by ~5%?

mjy (approved):
Simply re-anables runs.  OK as a start.

PR #2434, opened 2021-08-17
Fix extra semicolon in CE label when locality is blank

LordFlashmeow:
`!(collecting_event&.verbatim_locality&.blank?)` returns false if verbatim_locality is `nil`, so it creates an empty `<span></span>` element. This change fixes it so that it returns `nil` if `verbatim_locality` is `nil`

### DIFF
--- a/app/helpers/collecting_events_helper.rb
+++ b/app/helpers/collecting_events_helper.rb
@@ -100,7 +100,7 @@ module CollectingEventsHelper
   end
 
   def collecting_event_verbatim_locality_tag(collecting_event)
-    return nil unless !(collecting_event&.verbatim_locality&.blank?)
+    return nil unless collecting_event&.verbatim_locality&.present?
     content_tag(:span, collecting_event.verbatim_locality)
   end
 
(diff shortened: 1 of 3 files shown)


app/helpers/collecting_events_helper.rb
```diff
@@ -100,7 +100,7 @@ def collecting_event_dates_tag(collecting_event)
   end
 
   def collecting_event_verbatim_locality_tag(collecting_event)
-    return nil unless !(collecting_event&.verbatim_locality&.blank?)
+    return nil unless collecting_event&.verbatim_locality&.present?
```
LocoDelAssembly, line 103: Last safe navigation operator is technically not required (only on Rails):
```ruby
$ rails c
Loading development environment (Rails 6.0.4)
irb: warn: can't alias context from irb_context.
2.7.4 :001 > nil.present?
 => false 
2.7.4 :002 > nil&.present?
 => nil 
```

@mjy can you confirm it is safe for this helper to return nil (given it wasn't able to previously)?

LordFlashmeow: The only place it's called is here https://github.com/SpeciesFileGroup/taxonworks/blob/19853faef939f400e93b17919e5d3aa8a2b177e3/app/helpers/collecting_events_helper.rb#L14-L26

Since `compact` is called on it I can't imagine it would cause any issues.

PR #2448, opened 2021-08-20
Parse elevation precision from verbatim label of CE

LordFlashmeow:


app/javascript/vue/tasks/collecting_events/new_collecting_event/components/parsed/Elevation.vue
```diff
@@ -56,7 +56,7 @@ export default {
       units: [
         {
           label: 'Meters',
-          value: undefined
+          value: 'm'
```
LocoDelAssembly, line 59: @jlpereira can you confirm this doesn't trigger unintended behavior?

LordFlashmeow: Yeah, I noticed that unless you’re creating anew record, neither radio option is selected. 

I wonder if there’s a way to clean up the UI to show that it’s always meters (unless you’re inputting feet to be converted)

jlpereira: We can remove line 59 directly. value property it's not really needed for meters

LordFlashmeow: Without it the radio button isn’t selected though. Is that ideal?

jlpereira: Mmmmm I think it should, because the value would be undefined by default.

LordFlashmeow: What if we made the elevation default to meters instead of leaving it empty? I think that would make it more clear what the units are everywhere.

jlpereira: Sounds good to me

LocoDelAssembly: @LordFlashmeow please let me know when this PR is ready for merging.

LordFlashmeow: @LocoDelAssembly Should be set now.

PR #2455, opened 2021-08-23
String util bugfix and tests

LordFlashmeow:
Fixed bug in alphabetic_strings, add specs for more coverage

### DIFF
--- a/lib/utilities/strings.rb
+++ b/lib/utilities/strings.rb
@@ -99,10 +99,10 @@ module Utilities::Strings
 
   # @param string [String]
   # @return [Array]
-  #   whitespace split, then any string containing a digit eliminated
+  #   whitespace and special character split, then any string containing a digit eliminated
   def self.alphabetic_strings(string)
     return [] if string.nil? || string.length == 0
-    string.split(/\W/).select{|b| !(b =~ /\d/) }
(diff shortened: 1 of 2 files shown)


lib/utilities/strings.rb
```diff
@@ -96,12 +97,17 @@ def self.authorship_sentence(last_names = [])
...
+  # so splitting on accents creates pseudo accent insensitivity in searches.
+  #
   # @param string [String]
   # @return [Array]
-  #   whitespace split, then any string containing a digit eliminated
+  #   whitespace and special character split, then any string containing a digit eliminated
   def self.alphabetic_strings(string)
     return [] if string.nil? || string.length == 0
-    string.split(/\W/).select{|b| !(b =~ /\d/) }
+    string.split(/\W/).select{|b| !(b =~ /\d/) }.reject { |b| b.empty? }
```
LocoDelAssembly, line 110: Would `string&.split(/\W/)&.reject{ |b| b.blank? || b =~ /\d/ } || []` work as replacement for the entire code of this method?

LordFlashmeow: IMO it's a lot harder to read as a one-liner, this is clear how it works.

PR #2457, opened 2021-08-23
Only require matching parens in authorship for ICZN

LordFlashmeow:


LordFlashmeow:
Hmm, this workaround breaks some tests in `protonym_spec.rb` because the nomenclatural_code isn't defined. https://github.com/SpeciesFileGroup/taxonworks/blob/b9e8c2e7612e4b5641880e1b14a67410e105fddb/spec/models/protonym_spec.rb#L34-L56

PR #2459, opened 2021-08-23
Minor dwc_fields bugfixes

LordFlashmeow:


app/models/collection_object/dwc_extensions.rb
```diff
@@ -209,7 +209,7 @@ def dwc_sex
 
   def dwc_verbatim_coordinates
     return nil unless collecting_event
-    [collecting_event.verbatim_latitude, collecting_event.verbatim_latitude].compact.join(' ').presence
+    [collecting_event.verbatim_latitude, collecting_event.verbatim_longitude].compact.join(' ').presence
```
mjy, line 212: Oops, first `verbatim_latitude` is also wrong.

LordFlashmeow: What do you mean?

mjy: I mean I don't know how to read diffs.

PR #2466, opened 2021-08-24
Error handling for DwC Checklist, prevent duplicate taxon names

LordFlashmeow:


LocoDelAssembly:
Looks you opted to raise if name already exists. What happens if a family, genus, etc. already exists but some or all of the children species do not exist in the database yet?

LordFlashmeow:
I wonder if we need a `skipped` status for situations like this? Otherwise we can mark it as imported and skip it silently.

mjy:
Ultimately I would like to have multiple modes as in the simple batch load paradigm.  They let the user say what they will tolerate.

1 - Any single failure anywhere, nothing may be imported
2 - Any single error in a row - nothing in that row is imported
3 - Import anything you can, warning but not breaking on skipped records

lib/utilities/strings.rb
```diff
@@ -151,7 +151,50 @@ def self.only_integers?(string)
     !(string =~ /[^\d\s]/i) && !integers(string).empty?
   end
 
+  # Parse a scientificAuthorship field to extract author and year information.
+  #
+  # If the format matches ICZN, adds parentheses around author name (if detected)
+  # @param [String] authorship
+  # @return [Array] [author_name, year]
+  def self.parse_authorship(authorship)
```
mjy, line 159: A reminder to consider dwc_agent gem, does it help here?  The more we can unify logic to external the better.

PR #2470, opened 2021-08-26
DwC Checklist importer + specs

LordFlashmeow:
Example cases for occurrence and checklist files


mjy:
@LordFlashmeow can you resolve this: https://github.com/SpeciesFileGroup/taxonworks/pull/2470/checks?check_run_id=3606130946#step:21:86

@LocoDelAssembly after that I think we can pull?

LocoDelAssembly:
originalGenusNameID need to be relaxed and not make mandatory (breaks dataset listed in https://ipt.gbif.org/manual/en/ipt/2.4/checklist-data#exemplar-datasets, our original code was able to handle accepted names). Also breaks ITIS' DwC-As as well since them do not have originalGenusNameID either.

But I think it is best I merge this anyway and work on top of it to fix. Getting originalGenusNameID is much more involved for the general case I'm afraid, lots of lookups on other records to see which one seems to be the original (having originalGenusNameID as best source to make a decision when available).

PR #2475, opened 2021-08-28
Fix missing author in checklist, failure to find Protonym in occurrence

LordFlashmeow:


app/models/dataset_record/darwin_core/occurrence.rb
```diff
@@ -39,14 +43,14 @@ def self.execute(origins, parent, name)
...
-        Protonym.find_by(name.merge({ parent: parent })).tap do |protonym|
-          if protonym.nil?
-            raise DatasetRecord::DarwinCore::InvalidData.new({
-              origins[name.object_id] =>
-              ["Protonym #{name[:name]} not found with that name and/or classification. Importing new names is disabled by import settings."]
-            })
-          end
-        end
+        # If there is an intermediate ancestor between parent and name, this will still find the protonym
+        protonym = Protonym.find_by(name.merge({parent: parent})) || parent.descendants.find_by(name)
```
mjy, line 47: I feel like we could isolate this to its own method in this class and add 2-3 tests to demonstrate behaviour?  That would make me feel better about accepting this merge.

LordFlashmeow:
Fixed by #2740

Issue #2494, opened 2021-09-03
Model - Download, add subclasses


mjy:
Reasons:
* filter on type
* custom initialization per type
* facilitates upstream modelling of dataset metadata per specific type

PR #2498, opened 2021-09-04
Add STI to download

mjy:
### Added
* Add `type` column to Download [#2494]
* Add corresponding types of Download (bibtex, coldp, dwc_archive, etc.)

### Changed
* Extract `metamorphosize` from `is_data` and add to its own concern
* Update Download factories to reference `type`

[#2494]: https://github.com/SpeciesFileGroup/taxonworks/issues/2494


LocoDelAssembly:
@mjy cannot see anything wrong. Do we skip next minor release or merge now?

mjy:
Merge after minor.

PR #2534, opened 2021-09-21
2531 slow source smart selector

mjy:


mjy:
This issue fell back to underlying response issues, we will re-open new issues for Source smart selector improvements.

PR #2536, opened 2021-09-22
Add taxonworks.cls, set default Source cached to point to it, update source bibliography rendering

LocoDelAssembly:
Please continue work here and wait some review + tests passing. Also please edit description of pull request.


spec/features/collection_objects_spec.rb
```diff
@@ -92,11 +92,10 @@
                 end
 
                 specify 'source' do
-                  s = FactoryBot.create(:valid_source, by: @user) 
-
+                  s = FactoryBot.create(:valid_source, by: @user)
                   fill_autocomplete('source_id_for_specimen_batchload', with: s.cached, select: s.id)
                   click_button('buf_preview')
-                  expect(page).to have_content("Using '#{s.cached}' as source.")
+                  expect(page).to have_content("Using '#{s.cached.gsub('<i>','').gsub('</i>', '')}' as source.")
```
mjy, line 98: Why is gsub is necessary, did we get rid of `<i>`?

spec/lib/vendor/serrano_spec.rb
```diff
@@ -102,13 +102,13 @@
...
-        'Kevan, D.K.M.E. & Wighton, D.C. (1981) Paleocene orthopteroids from south-central Alberta, Canada. Canadian Journal of Earth Sciences 18, 1824–1837.'
+        'Kevan, D.K.M.E. & Wighton, D.C. (1981) Paleocene orthopteroids from south-central Alberta, Canada. Canadian Journal of Earth Sciences, 18(12), 1824–1837. https://doi.org/10.1139%2Fe81-170'
       )
     end
 
     specify 'html1' do
-      expect(src1.cached_string('html')).to start_with(
-        'Brauer, A. (1909) <i>Die Süsswasserfauna Deutschlands. Eine Exkursionsfauna bearb. ... und hrsg. von Dr. Brauer.</i> G. Fischer, Available from: https://doi.org/10.5962%2Fbhl.title.1086')
+      expect(src1.cached_string('html')).to eq(
+        'Brauer, A. (1909) <i>Die Süsswasserfauna Deutschlands. Eine Exkursionsfauna bearb. ... und hrsg. von Dr. Brauer.</i> G. Fischer, https://doi.org/10.5962%2Fbhl.title.1086')
```
mjy, line 111: Can we include `Available from`?

spec/lib/vendor/serrano_spec.rb
```diff
@@ -118,7 +118,7 @@
 
     specify 'html2' do
       expect(src2.cached_string('html')).to eq(
-        'Kevan, D.K.M.E. &amp; Wighton, D.C. (1981) Paleocene orthopteroids from south-central Alberta, Canada. <i>Canadian Journal of Earth Sciences</i> 18, 1824–1837.'
+        'Kevan, D.K.M.E. &amp; Wighton, D.C. (1981) Paleocene orthopteroids from south-central Alberta, Canada. <i>Canadian Journal of Earth Sciences</i>, 18(12), 1824–1837. https://doi.org/10.1139%2Fe81-170'
```
mjy, line 121: I feel we need a prefix for URL.

spec/features/collection_objects_spec.rb
proceps: not sure. This failed first time the changes were submitted, and I could not test feature-tests locally, had to guess what is is going on.

spec/models/source/bibtex_spec.rb
```diff
@@ -968,20 +968,20 @@
...
           end
 
           specify 'stated_year' do
             src1 = FactoryBot.create(:soft_valid_bibtex_source_article)
             src1.update(stated_year: '1699')
-            expect(src1.cached).to eq('Person, T. (1700) I am a soft valid article. <i>Journal of Test Articles</i>. [1699]')
+            expect(src1.cached).to eq('Person, T. (1700) I am a soft valid article. <i>Journal of Test Articles</i>. [1699].')
             src1.update(volume: '25')
-            expect(src1.cached).to eq('Person, T. (1700) I am a soft valid article. <i>Journal of Test Articles</i> 25. [1699]')
+            expect(src1.cached).to eq('Person, T. (1700) I am a soft valid article. <i>Journal of Test Articles</i>, 25. [1699].')
```
mjy, line 984: Trailing period necessary in this case?

spec/features/collection_objects_spec.rb
proceps: can remove it and push again to see if it fails.

LocoDelAssembly: New code introduced them in cached. On my computer I fixed this spec with `'#{s.cached.gsub(/<\/?i>/,'')}'` after lots of trials and errors because I wasn't noticing the actual problem. Then discarded my code and branched because the specs breakage was more extensive than this.

spec/lib/vendor/serrano_spec.rb
proceps: Yes. Including DOI?

mjy: Probably not for DOI?  Maybe just for URL or that other field (available from or something like it).

proceps: updating URL filed. The other field is "howpublished". I checked all styles, non of them using this field, it is not even a valid variable for CSL. I checked what are valid values for the field (according to bibtex):
The howpublished field is used to store the a notice for unusual publications. Examples:
```
howpublished = "Distributed at the local tourist office"
howpublished = "WebPage"
howpublished = "Museum exhibit"
```

proceps: The field which we are really missing in our bibtex is: "number-of-pages".  This is needed for book chapter to indicated the total number of pages in the book.

lib/queries/source/autocomplete.rb
```diff
@@ -212,12 +212,11 @@ def autocomplete
           # order results by number of times used
           if scope
             a = a.left_outer_joins(:citations)
-              .select('sources.*, COUNT(citations.id) AS use_count, MAX(citations.project_id) AS in_project_id, project_sources.id AS project_source_id')
+              .select('sources.*, COUNT(citations.id) AS use_count, MAX(citations.project_id) AS in_project_id, MAX(project_sources.id) AS project_source_id')
```
mjy, line 215: MAX doesn't seem right here, what is the goal?

lib/vendor/bibtex_ruby.rb
```diff
@@ -37,8 +37,8 @@ def self.namecase_bibtex_entry(bibtex_entry)
         bibtex_entry.names.each do |n|
           n.first = NameCase(n.first) if n.first
           n.last = NameCase(n.last) if n.last
-          n.prefix = NameCase(n.prefix) if n.prefix
-          n.suffix = NameCase(n.suffix) if n.suffix
+          #n.prefix = NameCase(n.prefix) if n.prefix
```
mjy, line 40: Why these changes?

spec/features/collection_objects_spec.rb
mjy: Italics in cached are fine.  SInce this is feature spec this is OK, I see we are just matching the UI.

lib/queries/source/autocomplete.rb
proceps: this does not do anything, practicing with sql and forgot to remove.

lib/vendor/bibtex_ruby.rb
proceps: we had problems with capitalized author prefixes and suffixes. I do not think those should be there. Last and first names should be capitalized. Suffix and preffix should stay as it is.

mjy: Namecase is smart, some prefix/suffix need capitalization.  Can we add a spec to illustrate problem?

proceps: I would rely on what people enter to the DB. I run into similar problem with `Serial` capitalization. There is function for smart capitalization which keeps "a", "the", "with", "on" and other parts lower case. The problem only is, it knows only English grammer. For other languages, the title get completely distorted. I had to remove this capitalization. Prefix/suffix capitalization is also dependant on language, and, in some cases, on author preferences. Yes, you can make generic rules, but I would prefer to keep it simple.

mjy: We can turn this on/off in the interface if required already.  Problematic records are a minority.

We have this because we import from crossref and other bibtex datasets that may have all caps etc., i.e. it fixes big batches of problems.   I don't want to have to manually downcase many many records on batch load.

proceps: In this case the best if we better restrict the functionality to cases when the Author and Editors are not parsed as people and Journals not parsed as Serial. I have problem that you take time to  curate people, but than it get replaced by some other logic.

lib/queries/source/autocomplete.rb
mjy: I will remove.

PR #2621, opened 2021-11-04
Fix DwC Checklist import when current rank doesn't match OC rank

LordFlashmeow:


mjy:
@LordFlashmeow thanks much.  Ready in your mind for review?  If so @LocoDelAssembly I have nothing to add.

LordFlashmeow:
Should be set for review. There’s some other import issues that I’d like to fix, but they’re not related to this one.

LocoDelAssembly:
@LordFlashmeow are you working with `20210527-05_16_02-worldants` (or newer version)? If so, could you push to the repo the script that generates importer-compatible file?

Will take an extra look, but very likely good to merge.

LordFlashmeow:
I’m running a newer version, but the script runs on the antcat server so all I can do is upload the file here

LocoDelAssembly:
What I have available at `antweb-staging` is this:
```
head 20210527-05_16_02-worldants.txt -n1 | sed 's/\t\t*/\n/g'
antcat id
subfamily
tribe
genus
subgenus
species
subspecies
author date
author date html
authors
year
status
available
current valid name
original combination
was original combination
fossil
taxonomic history html
reference id
bioregion
country
current valid rank
hol id
current valid parent
```
Would be possible to have something that the importer can process? (even if it is a smaller set than full-size dataset)

LordFlashmeow:
@LocoDelAssembly I've added `antcat_gbif_v0.04_2021-11-12.tsv`, which is the latest data from antcat, in the DwC format.

LocoDelAssembly:
Perfect, thanks @LordFlashmeow!

PR #2684, opened 2021-11-28
Fix issues with original combination having wrong genus

LordFlashmeow:
Would happen when original combination was a subsequent combination, usually due to a change in name from a conflict.


app/models/dataset_record/darwin_core/taxon.rb
```diff
@@ -225,13 +254,33 @@ def import(dwc_data_attributes = {})
...
+              parents << next_parent
+            end
+
+            # convert DatasetRecords into hash of rank, protonym pairs
+            parent_elements = parents.to_h do |p|
+              [
+                # Key is rank (as set in checklist file)
+                DatasetRecordField.where(dataset_record_id: p)
+                                  .at(get_field_mapping(:taxonRank))
+                                  .pick(:value)
```
LocoDelAssembly, line 272: From here on safe navigation should be used, in case `get_field_mapping(:taxonRank)` and/or `.pick(:value)` return `nil` (former perhaps is prevented from happening somewhere else, but latter can definitively happen).

app/models/dataset_record/darwin_core/taxon.rb
```diff
@@ -286,6 +335,7 @@ def import(dwc_data_attributes = {})
         message: e.message,
         backtrace: e.backtrace
       }
+      raise if Rails.env.development?
```
LocoDelAssembly, line 338: There is one line exactly like this one right below `rescue StandardError => e`. The purpose of re-raising is to crash in dev to inspect problems live rather than being recorded in record's metadata. This one must go away.

LocoDelAssembly:
@LordFlashmeow is this PR ready to merge? I can take care of fixing the lines I commented if that is all that is left to be done.

LordFlashmeow:
@LocoDelAssembly yes, I think that's everything

LocoDelAssembly:
Perfect. Will proceed with merge then. Thanks!

PR #2711, opened 2021-12-07
Include incertae sedis classification during DwC checklist import

LordFlashmeow:


LordFlashmeow:
Closing in favor of #2732, which also incorporates other classifications

PR #2721, opened 2021-12-09
Fix typo Latinized::PartOfSpeach -> Latinized::PartOfSpeech

LordFlashmeow:


LocoDelAssembly:
@proceps just final confirmation that the code that will now be enabled is totally correct? This was stayed this way for over a year.

proceps:
Yes, good to go

PR #2732, opened 2021-12-16
Add new TaxonNameClassifications to names during import

LordFlashmeow:
Includes 
* Gender/Part of Speech
* incertae sedis
* nomen nudum


LordFlashmeow:
I closed the branch while #2740 was in progress, but I think it's good to go now.

PR #2739, opened 2021-12-22
Use splatting in old batch loader `super()` calls 

LordFlashmeow:
I *think* ruby 3.0 removed automatic conversion of keyword arguments, so the splat is needed to convert from a hash to keywords.


LocoDelAssembly:
> I _think_ ruby 3.0 removed automatic conversion of keyword arguments, so the splat is needed to convert from a hash to keywords.

Thanks @LordFlashmeow. That's correct, we fixed code in several places but it seems we didn't catch them all yet.

PR #2740, opened 2021-12-24
Search ancestors when matching protonyms in import

LordFlashmeow:
When importing occurrences, the name might have intermediate parents (like tribe) that aren't described in the DwC-A file.

This fix checks immediate parentage first (as before), but also searches through ancestors if the first search turns up nothing.


LordFlashmeow:
@mjy I think this is ready for review now.

app/models/dataset_record/darwin_core/occurrence.rb
```diff
@@ -44,7 +44,20 @@ def get_protonym(parent, name)
       %I(name masculine_name feminine_name neuter_name).inject(nil) do |protonym, field|
         break protonym unless protonym.nil?
 
-        Protonym.find_by(name.slice(:rank_class).merge!({field => name[:name], :parent => parent}))
+        p = Protonym.find_by(name.slice(:rank_class).merge!({field => name[:name], :parent => parent}))
```
mjy, line 47: You don't need `merge!` here.

app/models/dataset_record/darwin_core/occurrence.rb
```diff
@@ -731,9 +744,19 @@ def parse_identification_class
...
     type_scientific_name = type_status&.[](:scientificName)&.gsub(/\s+/, ' ')
 
-    type_material = {
-      type_type: type_status[:type].downcase
-    } if scientific_name && type_scientific_name&.delete_prefix!(scientific_name)&.match(/^\W*$/)
+
+    if scientific_name && type_scientific_name.present?
+      if type_scientific_name&.delete_prefix!(scientific_name)&.match(/^\W*$/)
+        type_material = {
+          type_type: type_status[:type].downcase
```
mjy, line 751: Feels like you might compare (maybe done elsewhere) that the value here is a legal `type_type` (`TypeMateria::ICZN_TYPES` for example)?

LordFlashmeow: `type_type` is validated during create, which is called here: https://github.com/SpeciesFileGroup/taxonworks/blob/57894801c017e782f8a8ebf6ac4cc851cc40d292/app/models/dataset_record/darwin_core/occurrence.rb#L134-L140

Type material was originally designed to be skipped if it failed. If that's no longer the intent we can make an issue for it.

app/models/dataset_record/darwin_core/occurrence.rb
```diff
@@ -44,7 +44,20 @@ def get_protonym(parent, name)
       %I(name masculine_name feminine_name neuter_name).inject(nil) do |protonym, field|
         break protonym unless protonym.nil?
 
-        Protonym.find_by(name.slice(:rank_class).merge!({field => name[:name], :parent => parent}))
+        p = Protonym.find_by(name.slice(:rank_class).merge({field => name[:name], :parent => parent}))
+
+        # Protonym might not exist, or might have intermediate parent not listed in file
+        # if it exists, run more expensive query to see if it has an ancestor matching parent name and rank
+        if p.nil? && Protonym.where(name.slice(:rank_class).merge({field => name[:name]})).exists?
```
LocoDelAssembly, line 51: @LordFlashmeow preferably `where` should scope to current project here. Not sure if it causes damage because the `if`'s body then uses `parent` and that alone should prevent accessing names out of the project's scope but it would still be best to add `project_id: Current.project_id` to the params hash.

LocoDelAssembly: (or `project_id: parent.project_id`)

LordFlashmeow: Should we also scope it in the `find_by` on line 47?

mjy: Yes.  All finds except those that start with `Source`, `GeographicArea`, `Person`*, `Namespage`, `Repository` should be scoped to project.  This should significantly improve performance as well.

mjy: As a side note, it is sometimes worthwhile creating a single contrary record from another project to your unit test to ensure it's not being included in a result, for example something like `FactoryBot.create(:valid_otu, project: FactoryBot.create(:valid_project))`

mjy: When interacting with the data store we assume, always, that Current.project_id, and Current.user_id are set.  These two "globals" are the *only* ones that we use (more and the state space becomes confusing).

LordFlashmeow: @mjy I don't see it improving performance unless there aren't any conditions in the query referencing the current project.

mjy: @LordFlashmeow not sure I fully understand.  If the field you are hitting is not indexed (even if it is), then scoping to a project will mean the difference of searching among say 10k records and 1 million records?

LordFlashmeow: Running `Protonym.where(parent_id: 123)` has the exact same effect as running `Protonym.where(parent_id: 123, project_id: 1)` because id 123 is in project 1. The primary filter condition will be parent_id, not project_id. The results will be children of id 123, which implies that they are in project 1. 

We know that the results will be in the same project as the filtering object, so adding a check can slightly decrease performance.

For a query like `Protonym.where(name: Aus)`, it makes sense to filter by project id, because the filter condition isn't attached to a project.

LocoDelAssembly: Correct, checking for `project_id` was only needed in the `if` statement since it was only checking by rank and name and that was not enough to conclude the name exists in the current project (although besides false positive and performance hit of executing the `if` body, probably there wasn't any risk of actually using names from other projects when doing the actual import). The initial parent is the root taxon name (which IS checked that it comes from the intended project), then since the algorithm walks the tree the project membership invariant is preserved (and should generally be better to use parent_id index than project_id's).

PR #2828, opened 2022-02-24
Fix find_by scoping in DwC Occurrence importer

LordFlashmeow:


app/models/dataset_record/darwin_core/occurrence.rb
```diff
@@ -765,7 +765,7 @@ def parse_identification_class
         type_material = {
           type_type: type_status[:type].downcase
         }
-      elsif (original_combination_protonym = Protonym.find_by_cached_original_combination(type_scientific_name))
+      elsif (original_combination_protonym = Protonym.find_by(cached_original_combination: type_scientific_name, project_id: self.project_id))
```
LocoDelAssembly, line 768: Not related to this fix, but would be better to attempt matching to the original combination of the protonym detected by `scientificName`, then by some name with same original combination sharing same ancestors and have this current search as last resort?

@LordFlashmeow @mjy

LordFlashmeow: I don't think the occurrence's `Protonym` is available to that method

LordFlashmeow: We could move the identification to later, right before we create the object here: https://github.com/SpeciesFileGroup/taxonworks/blob/9f47503db99c885f73bd30b036c30348ca108e23/app/models/dataset_record/darwin_core/occurrence.rb#L136-L142

PR #2829, opened 2022-02-25
Add DwC Importer option to skip rows where type material creation fails

LordFlashmeow:
I realized that failing silently might not be the desired behavior for everyone, so I added an option to make rows fail if the type material couldn't be created.

### DIFF
--- a/app/javascript/vue/tasks/dwca_import/components/settings/RequireTypeMaterialSuccess.vue
+++ b/app/javascript/vue/tasks/dwca_import/components/settings/RequireTypeMaterialSuccess.vue
@@ -1,0 +1,41 @@
+<template>
+  <div class="label-above">
+    <label>
+      <input
+        type="checkbox"
+        v-model="requireTypeMaterialSuccess"
+      >
+      Skip/error rows where type material creation fails
+    </label>
+  </div>
+</template>
+
+<script>
+
+import { ActionNames } from '../../store/actions/actions'
+import { GetterNames } from '../../store/getters/getters'
+import { UpdateImportSettings } from '../../request/resources'
+
+export default {
+  computed: {
+    requireTypeMaterialSuccess: {
+      get () {
+        return this.$store.getters[GetterNames.GetDataset].metadata?.import_settings?.require_type_material_success
+      },
+      set (value) {
+        UpdateImportSettings({
+          import_dataset_id: this.dataset.id,
+          import_settings: {
+            require_type_material_success: value
+          }
+        }).then(response => {
+          this.$store.dispatch(ActionNames.LoadDataset, this.dataset.id)
+        })
+      }
+    },
+    dataset () {
+      return this.$store.getters[GetterNames.GetDataset]
+    }
+  }
+}
+</script>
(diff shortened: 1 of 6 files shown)


app/models/dataset_record/darwin_core/occurrence.rb
```diff
@@ -134,11 +134,19 @@ def import(dwc_data_attributes = {})
...
+
+          type_material = TypeMaterial.new(
+            {
+              protonym: innermost_protonym,
+              collection_object: specimen,
+            }.merge!(attributes[:type_material])) # protoynm can be overwritten in type_materials hash if OC did not match scientific name / innermost_protonym
+
+          if self.import_dataset.require_type_material_success? # raise error if validations fail and it cannot be imported
+            type_material.save
+          else
```
LocoDelAssembly, line 146: @LordFlashmeow `if` and `else` do the same thing?

app/javascript/vue/tasks/dwca_import/components/settings/RequireTypeMaterialSuccess.vue
```diff
@@ -0,0 +1,41 @@
+<template>
+  <div class="label-above">
+    <label>
+      <input
+        type="checkbox"
+        v-model="requireTypeMaterialSuccess"
+      >
+      Skip/error rows where type material creation fails
```
LocoDelAssembly, line 8: Not sure `Skip/error` will be well understood by users? I guess "Skip" here means maintain original behavior of proceeding with import regardless of type material import success?

What about "Prevent importing records with unprocessable `typeStatus` information"?

app/models/dataset_record/darwin_core/occurrence.rb
LocoDelAssembly: Ah, understood now. You need `!` on `if`'s `save`.

app/javascript/vue/tasks/dwca_import/components/settings/RequireTypeMaterialSuccess.vue
LordFlashmeow: Prevent sounds to me like a preemptive operation, like we're checking beforehand and we'll mark as Not Ready if it's unprocessable. Maybe "Error records with unprocessable `typeStatus` information"? 

It's challenging to phrase it precisely because "fail" is used for records that cause an internal error, rather than a data error.

Alternatively, what if it was an opt-out instead of opt-in? Something like 

"Ignore `typeStatus` information if unprocessable during record creation"

LocoDelAssembly: To not alter current behavior just in case lets go with "Error records with unprocessable typeStatus information"

LordFlashmeow: Works for me.

app/models/dataset_record/darwin_core/occurrence.rb
LocoDelAssembly: @LordFlashmeow this still needs to be fixed?

PR #2835, opened 2022-02-28
Fix DWC export month field

LordFlashmeow:


LocoDelAssembly:
Thanks @LordFlashmeow. @mabecabrera this should fix the month problem you saw in DwC export.

PR #2861, opened 2022-03-25
2839 otu geo json inventory

LocoDelAssembly:


app/models/otu.rb
```diff
@@ -364,6 +273,114 @@ def self.select_optimized(user_id, project_id, target = nil)
...
+      a = b.parent_otu_id(prefer_unlabelled_otus: true)
+    end
+    ids
+  end
+
+  # @return [Array]
+  #   all bilogical associations this Otu is part of
+  def all_biological_associations
+    # !! If self relationships are ever made possible this needs a DISTINCT clause
+    BiologicalAssociation.find_by_sql(
```
LocoDelAssembly, line 335: @mjy would this be equivalent or does `UNION` has something specific we need?
```ruby
BiologicalAssociation.where(biological_association_object: self).or(
  BiologicalAssociation.where(biological_association_subject: self)
)

Comment (mjy) [reply app/models/otu.rb:335]: I believe that's OK in this case.  Earlier I had experimented with some performance vs. UNION, IIRC that was with taxon name relationships, and found a massive gain in that case, but I can't remember if that case was identical to this.  The other thing that tickles my brain is that UNION makes the result DISTINCT, and having to call DISTINCT when scopes were chained might have lead to issues.  It's most likely though that I just used an old pattern before the `or()` syntax was available.

Comment (LocoDelAssembly) [reply app/models/otu.rb:335]: Resulting query is this:
```ruby
SELECT "biological_associations".*
FROM "biological_associations" 
WHERE (
  "biological_associations"."biological_association_object_type" = $1 AND
  "biological_associations"."biological_association_object_id" = $2 OR
  "biological_associations"."biological_association_subject_type" = $3 AND
  "biological_associations"."biological_association_subject_id" = $4
)
[
  ["biological_association_object_type", "Otu"],
  ["biological_association_object_id", 1],
  ["biological_association_subject_type", "Otu"],
  ["biological_association_subject_id", 1]
]
```

In this case `UNION` is not doing much work with respect to query above since it is including `id` so in essence it is already as `DISTINCT` as `UNION` query is.

mjy: I'm fine to change it, let's just make sure there is a unit test behind the code, there may already be.

LocoDelAssembly:
@mjy `inventory/descendants` seems to be removed. Is this replaced with something else or support has been dropped? This happened in https://github.com/SpeciesFileGroup/taxonworks/pull/2861/commits/a0d3726c454790ba1c99e852e46d15110d69483e and subsequent commit fixes an incomplete merge.

mjy:
@LocoDelAssembly It was changed to `inventory/taxonomy`.  See Notes on Page Google doc for summary of changes and new endpoints. Feel free to transcribe them here.

LocoDelAssembly:
Ah, I see, sorry, seen that but didn't connect the dots.

PR #2867, opened 2022-03-30
Fall back to match repository by name in DwC importer

LordFlashmeow:
In the DwC Occurrence importer, some repositories may not have acronyms, only names. We should try matching on that if no acronym matches the `institutionCode`

### DIFF
--- a/app/models/dataset_record/darwin_core/occurrence.rb
+++ b/app/models/dataset_record/darwin_core/occurrence.rb
@@ -385,7 +385,11 @@ class DatasetRecord::DarwinCore::Occurrence < DatasetRecord::DarwinCore
       repository = Repository.find_by(acronym: institution_code)
 
       # Some repositories may not have acronyms, in that case search by name as well
-      repository |= Repository.find_by(name: institution_code)
+      unless repository
+        repository_results = Repository.where(Repository.arel_table['name'].matches(Repository.sanitize_sql_like(institution_code)))
+        raise DarwinCore::InvalidData.new({ "institutionCode": ["Multiple repositories match the name #{institution_code}. Please use the acronym instead."] }) if repository_results.count > 1
+        repository = repository_results.first
+      end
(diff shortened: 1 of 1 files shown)


app/models/dataset_record/darwin_core/occurrence.rb
```diff
@@ -383,7 +383,10 @@ def parse_record_level_class
     institution_code = get_field_value(:institutionCode)
     if institution_code
       repository = Repository.find_by(acronym: institution_code)
-      raise DarwinCore::InvalidData.new({ "institutionCode": ["Unknown #{institution_code} repository. If valid please register it using '#{institution_code}' as acronym."] }) unless repository
+
+      # Some repositories may not have acronyms, in that case search by name as well
+      repository |= Repository.find_by(name: institution_code)
```
LocoDelAssembly, line 388: As fallback I'd suggest to use `Repository.where(Repository.arel_table['name'].matches(Repository.sanitize_sql_like(institution_code)))` so the search is case insensitive. Would be best to still fail if the result brings more than one result. (You may of course use more variable to have the expression more readable)

@mjy Repository seems it doesn't have uniqueness validation for at least `acronym` field? Shouldn't have validation? `Repository.group(:acronym).having('count(*) > 1').count` is not returning an empty set, same with `:name`, only the pair `:acronym, :name` is returning empty set. With current data `Repository.find_by(acronym: institution_code)` might not always be hitting the correct repository.

LordFlashmeow: @LocoDelAssembly Should I make these changes (including the uniqueness validation) now, or should I hold off until you have the uniqueness check on the backend done?

LocoDelAssembly: @mjy, is Repository good as it is currently implemented? @LordFlashmeow I think you may apply the changes to do insensitive compare, but not sure how are we going to deal with duplicate repositories (currently it is decided by a roll of the dice, of course).

Issue #2830, opened 2022-02-25
API - As a developer I want persistent endpoints and triggered builds for some new Download classes


mjy:
## API

These calls all require:
* The user is a project administrator
* User authentication
* Project authentication

* [ ] Endpoint(s) to trigger build(s)
  * [ ] If there is a build in progress then respond with failure and point to future address of the future completed build
* [ ] Endpoint to terminate ongoing builds 
   * [ ]  Considerations for stopping long running builds
* [ ] Endpoints to destroy existing builds

## Model

* [ ] Add Download::Fixed::DwC
* [ ] Add Download::Fixed::CoLDP

LordFlashmeow:
Could builds be assigned a UUID that could be checked on with a status endpoint?

This way remote clients could ping for status and download once the build is complete.

mjy:
Builds will have a Download#id that can be used to check the status endpoint, that might be enough.

kleintom:
See also #1849

Issue #2410, opened 2021-08-11
As a dev-op I want to create a new empty project in development mode only in one click


mjy:
```
When I am demoing in development mode, 
   and I want to clean slate, 
      then I want to setup a new empty project in one click, 
          and that project should have me as a user
          and I should have that project selected
          and that project should be forceably named to be the top of the list on my dashboard
```

PR #2936, opened 2022-04-21
2830 persistent endpoints for downloads

mjy:
This adds the following endpoints, allowing for the creation of DwC Archive dumps via the external api:

* POST `http://127.0.0.1:3000/api/v1/downloads/build?project_token=<>&token=<>&type=Download::DwcArchive::Complete`
* DELETE `http://127.0.0.1:3000/api/v1/downloads/123?project_token=<>&token=<>`
* GET `http://127.0.0.1:3000/api/v1/downloads/123?project_token=<>`
* GET `http://127.0.0.1:3000/api/v1/downloads/123/file?project_token=<>` # return the binary

# TODO:
* [x] Confirm CSRF modifications to allow delete and post to /api/v1 are not otherwise dangerous
* [x] Test time to calculate sha2 for large exports
* [x] Test that the SHA2 calculation (on the zipfile) is repeatble by other non-Ruby calculators
* [x] Allows data_attributes params to pass through in requests
* [x] Ensure basic, non external Download models are not impacts (auto tests should do this)
* [x] Confirm that ActiveJobs are cancelled (raise quietly, but see if we can log if noisy) when the corresponding serialized Download object is missing from the database.  This is the easiest way to cancel the job
* [x] Determine whether we need to cancel jobs "in progress".   If needed the likely way to do this is to 1) update the Download instance at the start of build time to indicate builds are in-progress (new Download attribute/migration) then 2) prevent DELETE requests if "in progress" .   We could extend this idea to use a % complete style "in progress" tracker to indicate roughly how far we are along.  Not nil in this field would indicate the build is in progress.

# Changelog

### Added
* Download API enpoints and model to facilitate DwC Archive building remotely [#2830]
* Option to include project creators as members on new project creation [#2410]

[#2830]: https://github.com/SpeciesFileGroup/taxonworks/issues/2830
[#2410]: https://github.com/SpeciesFileGroup/taxonworks/issues/2410


LocoDelAssembly:
Is my understanding correct that POST/DELETE require user token and project token alone is not enough to authorize access?

Was about to raise concerns about download request flooding but found out that there is protection against that by not allowing more than a download. Perhaps will need some security measures to prevent supplying any download type, but at least I tried with `type=Download::Text&name=test` and it still marked that name cannot be blank.

mjy:
> Is my understanding correct that POST/DELETE require user token and project token alone is not enough to authorize access?

Yes, POST/PATCH should require user.

Only specific download types are allowed, and n=1 right now, so hopefully that is not an issue.

Thanks for testing.

LocoDelAssembly:
@LordFlashmeow I think this is tracking your file download requirements? 

> Allow data_attributes params to pass through in requests
Would that mean filtering by attributes of objects or rather by DataAttribute associations?

LocoDelAssembly:
>     * [ ]  Test time to calculate sha2 for large exports
```
3.1.2 :048 > Benchmark.bm do |x|
3.1.2 :049 >   (1..32).each do |power|
3.1.2 :050 >     data = SecureRandom.random_bytes(2**power)
3.1.2 :051 >     x.report(number_to_human_size(2**power)) { Digest::SHA256.hexdigest(data) }
3.1.2 :052 >   end
3.1.2 :053 > end
       user     system      total        real
2 Bytes  0.000002   0.000008   0.000010 (  0.000009)
4 Bytes  0.000001   0.000003   0.000004 (  0.000004)
8 Bytes  0.000017   0.000000   0.000017 (  0.000003)
16 Bytes  0.000003   0.000000   0.000003 (  0.000003)
32 Bytes  0.000003   0.000000   0.000003 (  0.000003)
64 Bytes  0.000004   0.000000   0.000004 (  0.000003)
128 Bytes  0.000004   0.000000   0.000004 (  0.000003)
256 Bytes  0.000005   0.000000   0.000005 (  0.000004)
512 Bytes  0.000006   0.000000   0.000006 (  0.000005)
1 KB  0.000008   0.000000   0.000008 (  0.000007)
2 KB  0.000011   0.000000   0.000011 (  0.000011)
4 KB  0.000019   0.000000   0.000019 (  0.000018)
8 KB  0.000034   0.000000   0.000034 (  0.000033)
16 KB  0.000063   0.000000   0.000063 (  0.000062)
32 KB  0.000121   0.000000   0.000121 (  0.000121)
64 KB  0.000000   0.000238   0.000238 (  0.000238)
128 KB  0.000000   0.000504   0.000504 (  0.000503)
256 KB  0.000988   0.000000   0.000988 (  0.000987)
512 KB  0.000379   0.001533   0.001912 (  0.001912)
1 MB  0.003588   0.001058   0.004646 (  0.004630)
2 MB  0.008464   0.000000   0.008464 (  0.008475)
4 MB  0.015879   0.000000   0.015879 (  0.015879)
8 MB  0.031285   0.000000   0.031285 (  0.031285)
16 MB  0.060420   0.001607   0.062027 (  0.062027)
32 MB  0.121886   0.000000   0.121886 (  0.121886)
64 MB  0.245944   0.000000   0.245944 (  0.245946)
128 MB  0.488286   0.000393   0.488679 (  0.488710)
256 MB  0.980587   0.000887   0.981474 (  0.981545)
512 MB  1.968927   0.000845   1.969772 (  1.969840)
1 GB  3.945651   0.001307   3.946958 (  3.947090)
2 GB  7.932234   0.000000   7.932234 (  7.932557)
4 GB 16.372113   0.008870  16.380983 ( 16.388204)
```

LordFlashmeow:
> @LordFlashmeow I think this is tracking your file download requirements?
> 
> > Allow data_attributes params to pass through in requests
> > Would that mean filtering by attributes of objects or rather by DataAttribute associations?

Yes. Ideally we could provide a list of predicate IDs to include as a parameter to the request.

mjy:
@LordFlashmeow Predicate params should be there already, I should have say "Allows", i.e. they need testing, not implementation to pass that check:
```Ruby
    params.permit(collecting_event_predicate_id: [], collection_object_predicate_id: [] ).transform_keys(&:to_sym).to_h
 ```

mjy:
@LocoDelAssembly @LordFlashmeow I'd like to merge this, but do need help confirming the last 3 checks are OK.

gdower:
I tested COLDP and DarwinCore dashboard downloads with development merged and both still worked.

gdower:
If you delete a Download while it is running, the job completes and the download file is generated (but it may not be a problem as discussed with @mjy). At least you can't access the file for a deleted download at: http://localhost:3000/api/v1/downloads/{id}/file

I'm not sure how to do a call with the data_attributes or couldn't get it to work.

kleintom:
Apologies if these are questions that have already been discussed:
* to confirm, DELETE can apply to any download id, not just the id of a 'complete' download
* the GET binary file request only requires a project_token, which is publicly available for any project that has a public taxon pages, e.g. So assuming somebody *in* the project has created a complete download, then *anybody*, whether in or out of the project, can download that file (discoverable by the /downloads endpoint). Essentially it means if your taxonomy data is public (TP) then your occurrence data is as well (as soon as somebody creates a complete download). I get that people will typically be using the download precisely so they *can* make that data public (or it's already in GBIF), but just wanted to confirm this is expected.

kleintom:
For reference, here's the syntax for `predicate_extensions` that I was using for testing:
`curl -X POST -H "Content-Type: application/json" 'http://localhost:3000/api/v1/downloads/build?project_token=<>&token=<>&type=Download::DwcArchive::Complete&predicate_extensions\[collecting_event_predicate_id\]\[\]=31&predicate_extensions\[collecting_event_predicate_id\]\[\]=756&predicate_extensions\[collection_object_predicate_id\]\[\]=160&predicate_extensions\[collection_object_predicate_id\]\[\]=162'`

kleintom:
* If `api/build` is submitted, and then the Download it creates is deleted before the background job starts, the job fails - with no email notification - with error message `FAILED (0 prior attempts) with ActiveRecord::RecordNotFound: Couldn't find Download with 'id'=1462 [WHERE (expires >= $1)]`
* If `api/build` is submitted, and then the background job starts creating the dwca, and then its Download is deleted, the job finishes creating the dwca without error and moves it to its final location, but then we notice the Download has been deleted, so the zip is removed and the job errors - with no email notification - with error message `FAILED (0 prior attempts) with TaxonWorks::Error: Complete download build aborted: download '1464' no longer exists.`

kleintom:
This is ready for review, code and/or otherwise, to the extent anybody would like to.

I suggest we add a warning somewhere that *api calls should always/only be made using https://*. That may be automatic from a browser (in some cases), but not from curl, e.g., and I think to do POST and DELETE, as users will be now, users will need to be using something like curl.

kleintom:
Updated plan, arising from discussions with @mjy, but partly my own additions/changes for implementation - further thoughts welcome/encouraged.

GBIF isn't going to call `build` and/or `delete` to update things, so remove those endpoints. There's now a single *public* (requires a project_token only) `get '/downloads/dwc_archive_complete'` endpoint.
* It's guarded by a project-level 'Is public' setting for this one-per-project 'complete' download.
* When 'Is public' is false, calls on the endpoint fail and do nothing. When 'Is public' is true:
  * if a download doesn't exist, the api call fails but a download creation is begun.
  * if a download exists and is `ready` it's returned

Download settings are stored in the Project `preferences` hash:
image
Currently there's only *one* settings page for this 'complete' project download.

The main thing left to figure out I think is how to update these downloads.
* The simplest would be to add a setting 'Max age' or something: if the api endpoint is called and the existing download is older than max age then we return the old download but trigger the creation of a new one. The downside here is that at worst you're always getting a download that's >= max age - maybe that's fine if GBIF downloads once a day and your max age is 0.5 days, maybe not so much if GBIF downloads once a month or twice a year.
* Setup some version of cron to recreate the download every n days.

Other TODO:
* Add DAs, TW-specific attribute selection to preferences (as in dwc dashboard options)
* specs

kleintom:
Okay, ready for feedback/review/testing again. This pr is to provide a GET api endpoint `api/v1/downloads/dwc_archive_complete?project_token=<>` that returns a so-called 'complete' dwc archive of all project dwc occurrences.

* Each project has a complete download prefs page:
image
 
It covers:
  * EML settings (some fields of which are auto-filled when the download is created)
  * an 'Is Public' setting that determines whether or not the new endpoint will return anything
  * Predicate settings for the download (just like from the DwC Dashboard task)
  * A 'Max Age (days)' setting for the download that determines after what day a new request for the download will trigger creation  of a new download with updated data

So to allow downloads, set 'Is Public', edit and save EML (with no more 'STUB' strings). Then call `api/v1/downloads/dwc_archive_complete?project_token=<>` with the project token for your project - that will fail the first time but will trigger creation of a new complete download. When the download is ready, that same api call will return the download as a zip file.

If you call the endpoint when a complete download of age > max age exists, the 'old' existing download will be returned, but creation of a new one will also be triggered, and the new one will replace the old one when complete.

kleintom:
> Expected pattern for GBIF with its default 7-day fetch timeline and a project max-complete-download-age of ~6 days (6 can be a decimal)

More on real-world scenarios here: Assume GBIF calls on Sundays at noon.

>I don't think it's possible to get more than one week-old data in this case, given that GBIF calls once a week and downloads start with a GBIF call.

You can get 8 day old data if a user calls between Saturday noon and Sunday noon: The day "1" (1 day after user called) GBIF call gets "1"-day old data and doesn't trigger a rebuild, the day 8 GBIF call gets 8 day old data and does trigger a rebuild.

At this point the GBIF call at day 15 is guaranteed to get <= 7 day old data. The system is self-correcting in this sense.

That suggests the max-age should be as close to 7 as possible, but of course GBIF doesn't call at exactly the same time every week, so just be sure that max-age never overlaps with the GBIF variation (otherwise you can get *two*-week old data!).

Also note the assumption that *a GBIF call* started things off. That probably won't be the case. Worst case scenario the last regenerating call prior to the first GBIF call was Monday at 1. Then the first GBIF call <6 days later get ~6 day old data and doesn't trigger a rebuild, the second GBIF call ~13 days later gets 13 day old data and does trigger a rebuild. Again at this point the system has auto-corrected and we're in the above scenarios where the max age of GBIF data is ~8 days.

If you want to avoid this then either delete the complete download before the first GBIF call or delete and generate a new call <1 day before the first GBIF call.

kleintom:
Added a required 'default user' (no restrictions on who) to be used as creator/updater of complete downloads when one isn't otherwise set.

Untitled

Uploads to GBIF via IPT are still looking good (both extensions are functioning).

mjy:
> Added a required 'default user' (no restrictions on who) 

Perfect!

Issue #2855, opened 2022-03-18
[Enhancement] Apply tags to CO/CE in DwC Occurrence importer


LordFlashmeow:
### Feature or enhancement

I think it would be very helpful to be able to tag records as an alternative to using a custom predicate, especially for boolean values. 

It enables the flagging of records for follow-up once they're imported, without adding a value to every record (eg. specimens that need loans created).

Similar to the syntax for predicates, I suggest the syntax of `TW:Tag:CollectionObject/Event:TagName`, and a value of `1` or `true` to apply the tag, and a blank value to not apply it. While using more columns, I think that this is clearer and faster to produce and manipulate with tools than having a single column that holds a list of tags to apply.

### Location

DwC Occurrence Importer

LordFlashmeow:
There's currently no way to clear multiple values with the find&replace tool. As a result, the tag feature does not play nice with datasets using `true` and `false` values in a tag column.

Should we change the spec so that `0` or `false` does not apply the tag (but does not remove it if it's already applied)?

This way, users can correct "invalid" tag values in bulk when they do not want it to be applied.

LordFlashmeow:
Implemented in #3019

PR #3019, opened 2022-06-11
Support adding tags to collecting events and objects in DwC importer

LordFlashmeow:
Implements #2855


app/models/dataset_record/darwin_core/occurrence.rb
```diff
@@ -211,6 +211,17 @@ def import(dwc_data_attributes = {})
 
         # TODO: If all attributes are equal assume it is the same event and share it with other specimens? (eventID is an alternate method to detect duplicates)
         if collecting_event
+          # if tags have been specified to be added, update the collecting event
+          if attributes[:collecting_event][:tags_attributes]
+            # get list of preexisting tags, exclude them from update
+            current_tags = collecting_event.tags.pluck(:keyword_id).to_set
```
LocoDelAssembly, line 217: I missed part of the conversation in today's meeting, but is preserving current values intentional or this is related to the validation errors you mentioned?

app/models/dataset_record/darwin_core/occurrence.rb
```diff
@@ -211,6 +211,17 @@ def import(dwc_data_attributes = {})
...
+
+            new_tags = attributes[:collecting_event][:tags_attributes].reject { |t| current_tags.member?(t[:keyword].id)}
+
+            # add tags if there were any new ones
+            collecting_event.update!(tags_attributes: new_tags) if new_tags
```
LocoDelAssembly, line 222: I think this will destroy existing tags too. Is this intended?

mjy: This will not destroy existing tags, unless `_destroy: true` IIRC

PR #3037, opened 2022-06-29
Vitest

jlpereira:
This branch includes changes that make it possible to create unit and integration tests for our frontend environment.

Packages included:
- Vitest: https://vitest.dev/
- @testing-library/vue: https://testing-library.com/
- MSW: https://mswjs.io/


mjy:
@jlpereira How does this parallel what we do in functional tests?  I'm not sure how keen I am in introducing a second testing paradigm.

PR #3045, opened 2022-07-10
DwC Importer tag fix: only save CE if new tags were added

LordFlashmeow:
I missed an if statement for saving collecting events, causing collecting events to be saved even if there were no new tags added. As a result, importing collection objects was significantly slowed down when the collecting event had many related COs.


LocoDelAssembly:
Just in case I'll proceed with merge, but ActiveRecord might actually already been preventing the saving from happening when no changes have occurred since last save/create?

LordFlashmeow:
> Just in case I'll proceed with merge, but ActiveRecord might actually already been preventing the saving from happening when no changes have occurred since last save/create?

I don't think so. I was stepping through the import to figure out why it was so slow, and it was hitting some sort of update-on-save for all the COs. It spent 5 minutes chugging away on a single row with an event that had ~300 linked COs. 

Maybe the DwC indexer? I didn't look too hard at what it was running, but it's easy enough to test since the update-on-save still runs when new tags _are_ provided.

PR #3055, opened 2022-07-24
Fix dependency loop in checklist import 

LordFlashmeow:
When current name rank is genus and original combination is with self at subgenus level, eg: originally described as Aus (Aus) but now just Aus, the staging function would mark names as depending on themselves.

### DIFF
--- a/app/models/dataset_record/darwin_core/taxon.rb
+++ b/app/models/dataset_record/darwin_core/taxon.rb
@@ -447,7 +447,7 @@ class DatasetRecord::DarwinCore::Taxon < DatasetRecord::DarwinCore
 
   private
 
-  # @return [DatasetRecord::DarwinCore::Taxon, Array<DatasetRecord::DarwinCore::Taxon>]
+  # @return Optional[DatasetRecord::DarwinCore::Taxon, Array<DatasetRecord::DarwinCore::Taxon>]
   def get_parent
     DatasetRecord::DarwinCore::Taxon.where(id: import_dataset.core_records_fields
                                                              .at(get_field_mapping(:taxonID))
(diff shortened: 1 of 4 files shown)


app/models/dataset_record/darwin_core/taxon.rb
```diff
@@ -137,6 +137,10 @@ def import(dwc_data_attributes = {})
               original_combination_parents << next_parent
             end
 
+            # in cases where the taxon original combination is subgenus of self eg Sima (Sima), the first parent of the list
+            # should be dropped because it hasn't been imported yet
+            original_combination_parents = original_combination_parents.drop_while {|p| p.status != "Imported" }
```
LocoDelAssembly, line 142: Are preconditions safe enough that using `drop_while` instead of `reject` is good? Or are you actually wanting to have possible tail imported records in the list?

LocoDelAssembly:
@LordFlashmeow besides the comment I just did (if changes are needed at all), is this PR ready for merge?

app/models/dataset_record/darwin_core/taxon.rb
LordFlashmeow: Originally, I chose `drop_while` so that it wouldn't have to keep iterating through the list once it found imported items. 

IIRC, only the first item in the list isn't imported, so a non-imported element somewhere else is a symptom of an unrelated issue. In that case, I don't mind it breaking to alert us.

But yes, the preconditions are very safe on this one, since the parents are added in hierarchical order.

LordFlashmeow:
> @LordFlashmeow besides the comment I just did (if changes are needed at all), is this PR ready for merge?

Yes, I think it's good to go.

Issue #3057, opened 2022-07-27
Add support for verbatim label in collecting event batch uploader


LocoDelAssembly:

LocoDelAssembly:
Related: https://github.com/SpeciesFileGroup/taxonworks/issues/1816

PR #3059, opened 2022-07-29
Added verbatim_label support for CE Castor batch load (refs #3057

LocoDelAssembly:
This is a request from 28 July 2022 technical meeting to add support for verbatim label in CE batch uploader. Although the change is simple enough I need it to be reviewed to make sure it is a good idea to add this. Since the field was added at the end it should not brake compatibility with existing input files.


LocoDelAssembly:
@mjy is this OK to merge? The failure is a random one, tests are actually passing.

mjy:
The change is fine. The question is, though, whether that importer works at all at this point?

LocoDelAssembly:
On that meeting it look it worked, but had this inconvenience of not supporting verbatim label import. Will merge now.

PR #3064, opened 2022-08-03
2733 default namespaces

LocoDelAssembly:
Changes in DwC occurrences importer
====
- Allows mapping by `collectionCode` only as a fallback when primary mapping by `institutionCode`:`collectionCode` does not contain a match.
- Mappings are now sorted

Due to the complexity of implementation, mapping by `instititionCode` only is not supported at this time.


LocoDelAssembly:
@LordFlashmeow can you please take a look at this before we merge it? Thanks

LordFlashmeow:
Looks good to me, though there's still the bug that makes the modal scroll to the bottom when you open it. (I think it's because it focuses on the bottom-most text box on open, making the view scroll down).

mjy:
@jlpereira ^

LocoDelAssembly:
I'm merging this since the UI problem is unrelated to this PR anyway. Submit fixed directly into `development` @jlpereira

PR #3077, opened 2022-08-22
Added 10km tolerance to geographic area validation for verbatim georeference

LocoDelAssembly:
This enables users to specify a geographic area other than the one that a verbatim georeference it is contained in provided it is not farther than 10km.

### DIFF
--- a/app/models/geographic_item.rb
+++ b/app/models/geographic_item.rb
@@ -996,13 +996,13 @@ class GeographicItem < ApplicationRecord
     unless !persisted? || changed?
       a = "(#{GeographicItem.select_geography_sql(id)})"
     else
-      a = "'#{geo_object.to_s}'::geography"
+      a = "ST_GeographyFromText('#{geo_object.to_s}')"
     end
 
     unless !geographic_item.persisted? || geographic_item.changed?
       b = "(#{GeographicItem.select_geography_sql(geographic_item.id)})"
     else
-      b = "'#{geographic_item.geo_object.to_s}'::geography"
+      b = "ST_GeographyFromText('#{geographic_item.geo_object.to_s}')"
     end
 
     ActiveRecord::Base.connection.select_value("SELECT ST_Distance(#{a}, #{b})")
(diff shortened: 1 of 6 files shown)


app/models/geographic_item.rb
```diff
@@ -1067,7 +1086,7 @@ def intersects?(target_geo_object)
   # @param [geo_object]
   # @return [Boolean]
   def distance?(target_geo_object)
-    self.geo_object.distance?(target_geo_object)
```
mjy, line 1070: Was this a uncaught error?

LocoDelAssembly: Yes, undefined method error. But I think this method is not actually being used (but I was hoping it was useful for me, it was not).

LocoDelAssembly:
@mjy do you think this is good to go?

app/models/geographic_item.rb
mjy: Go ahead and remove the method if it is not used.

app/models/georeference/verbatim_data.rb
```diff
@@ -74,6 +74,32 @@ def dwc_georeference_attributes
     h
   end
 
+  # @return [Boolean]
+  #    true if geographic_item.geo_object is completely contained in collecting_event.geographic_area
+  # .default_geographic_item
+  def check_obj_within_distance_from_area(distance)
```
mjy, line 80: Probably nothing optimizable, but had me think a little about whether/how many times we have to hit the database here.  Would likely take more work than worthwhile at this point to tune things.

app/models/geographic_item.rb
LocoDelAssembly: Other than `expect(geographic_item).to respond_to(:distance?)` doesn't seem to be used anywhere (and spec was only testing the method exists without actually executing it). This "go ahead" is for the entire PR or just removing this method?

mjy:
@LocoDelAssembly couple very minor comments.  Form etc. is fine for me.

app/models/geographic_item.rb
mjy: Remove it, and expectionation.

mjy:
Looks good to merge, thanks much.

PR #3166, opened 2022-10-31
Add distribution to CoL data package exporter

gdower:


lib/export/coldp/files/distribution.rb
```diff
@@ -0,0 +1,90 @@
...
+        areaID
+        area
+        gazetteer
+        status
+        referenceID
+        remarks
+      }
+
+      otus.joins(:asserted_distributions).distinct.each do |o|
+        o.asserted_distributions.each do |ad|
```
mjy, line 31: Look at `.includes(:geographic_area)`.

lib/export/coldp/files/distribution.rb
```diff
@@ -0,0 +1,90 @@
...
+
+          ga = GeographicArea.find(ad.geographic_area_id)
```
mjy, line 33: `ad.geographic_area`

PR #3167, opened 2022-10-31
Add basic metadata to CoL data package exporter

gdower:


lib/export/coldp.rb
```diff
@@ -28,13 +29,27 @@ def self.otus(otu_id)
 
     def self.export(otu_id, prefer_unlabelled_otus: true)
       otus = otus(otu_id)
+      Current.project_id = Otu.find(otu_id).project_id
```
mjy, line 32: Don't set Current.project_id.  Use a variable, or reference by the objects project_id.

PR #3208, opened 2022-11-29
3207 coldp remarks

gdower:


lib/export/coldp/files/taxon.rb
```diff
@@ -130,6 +128,15 @@ def self.generate(otus, root_otu_id = nil, reference_csv = nil, prefer_unlabelle
         remarks
       }
 
+      if root_otu_id.nil?
+        Current.project_id = otus[0].project_id
```
mjy, line 132: I feel you should not be setting this here.

lib/export/coldp/files/taxon.rb
```diff
@@ -130,6 +128,15 @@ def self.generate(otus, root_otu_id = nil, reference_csv = nil, prefer_unlabelle
...
+      else
+        Current.project_id = Otu.find(root_otu_id).project_id
```
mjy, line 134: Just use `project_id: otu.porject_id` as needed.  In general don't touch Current.

lib/export/coldp/files/taxon.rb
```diff
@@ -84,11 +84,9 @@ def self.according_to_date(otu)
   def self.link(otu)
     # API or public interface
   end
-
-  # TODO: flag/exclude ! is_public
-  def self.remarks(otu)
-    if otu.notes.load.any?
-      otu.notes.pluck(:text).join('|')
+  def self.remarks(otu, taxon_remarks_vocab_id)
+    if !otu.data_attributes.find_by(controlled_vocabulary_term_id: taxon_remarks_vocab_id).nil?
```
mjy, line 88: Can there just be one?
```ruby
if da = otu.data_attributes.find_by(controlled_vocabulary_term_id: taxon_remarks_vocab)
  da.value
else
  nil
end
```

lib/export/coldp/files/name.rb
```diff
@@ -17,8 +17,12 @@ def self.code_field(taxon_name)
     end
   end
 
-  def self.remarks_field(taxon_name)
-    Utilities::Strings.nil_squish_strip(taxon_name.notes.collect{|n| n.text}.join('; ')) # remarks - !! check for tabs
+  def self.remarks(name, name_remarks_vocab_id)
+    if !name.data_attributes.find_by(controlled_vocabulary_term_id: name_remarks_vocab_id).nil?
```
mjy, line 21: See below for setting variable inline with `if`.

gdower: I see what you mean but I think the line might be too long with it inline. I'll try to use the inline style when possible.

PR #3442, opened 2023-06-06
1852 dwca import geographical areas

### LINKED ISSUE #1852: DWCA Importer - Geographic Areas not imported
Can we use combo of Country, State, County to match to one/two default Geographic Area in TW Gazeteer?

Alternatively, we could have them select one Gazeteer to match against. 

This is mainly an issue when specimens have extremely little location data (e.g. "California") and their collecting event is imported as almost empty, since Geographic Area is not imported. In these cases, at the very least, if no geographic area imported, then whatever geographic areas are indicated in "Country/State/County" should be imported to verbatim Locality to create a slightly better Collecting Event

The extent of Collecting Event information reported in DWCA table:
image

Example "bad" collecting event created:
image

gdower:
Fix #1852

Adds 2 scopes to GeographicArea model:
- has_shape(boolean): 
  - returns only GAs with shapes if true
  - returns GAs with or without shapes if false or nil
- with_data_origin(data_origin)
  - filters to GAs with data_origin
  - does nothing if data_origin is blank

Modifies occurrence to add geographical areas to collecting events:
- If latitude/longitude is provided, GAs are imported using the cached geographic names from the coordinates
  - Not implemented: It might be good to error the import if the latitude/longitude doesn't fall within the DWCA provided  county, state_province, country because those fields are currently ignored if given latitude/longitude
- Otherwise, it attempts to use county, state_province, country_code/country, if provided
  - If only a country_code is given, attempts to lookup country name using country_names_and_code_elements GAs (but the selected GA will still match the import_setting data origin if provided)
- If no GA is found, attempts to remove the finest location level and searches again, because probably at least having state/province or country is better than nothing?
- Reformats GBIF-style basis of records: PRESERVED_SPECIMEN -> PreservedSpecimen so that they'll import

Adds two settings to the DWCA occurrence UI:
- Perhaps it would be better to remove the require_geographic_area_has_shape from the import_settings when the box is unchecked, because I doubt anyone will ever want to only find GAs without shapes and that would allow the has_shape(boolean) scope to be modified to return GAs without shapes which might be useful elsewhere in TW
image


mjy:
@gdower can you expand the PR details a little in your comment with specifics as to what you implemented?

app/javascript/vue/tasks/dwca_import/const/geographicalAreaDataOrigins.js
```diff
@@ -0,0 +1,12 @@
+export default {
+    NONE: '',
+    GEOGRAPHICAL_AREA_DATA_ORIGIN_GADM: 'gadm',
+    GEOGRAPHICAL_AREA_DATA_ORIGIN_SFG: 'SFG',
+    GEOGRAPHICAL_AREA_DATA_ORIGIN_NE_STATES: 'ne_states',
+    GEOGRAPHICAL_AREA_DATA_ORIGIN_COUNTRY_NAMES_AND_CODE_ELEMENTS: 'country_names_and_code_elements',
+    GEOGRAPHICAL_AREA_DATA_ORIGIN_NE_COUNTRIES: 'ne_countries',
+    GEOGRAPHICAL_AREA_DATA_ORIGIN_TDWG_L1: 'tdwg_l1',
```
mjy, line 8: @gdower I think we might have to make these reference groups, like NE (states, countries), TDWG (all 4 levels). Does this make sense?

app/models/dataset_record/darwin_core/occurrence.rb
```diff
@@ -236,11 +236,41 @@ def import(dwc_data_attributes = {})
...
+          has_shape = self.import_dataset.metadata['import_settings']['require_geographic_area_has_shape'] if self.import_dataset.metadata['import_settings'].include? 'require_geographic_area_has_shape'
+          data_origin = self.import_dataset.metadata['import_settings']['geographic_area_data_origin'] if self.import_dataset.metadata['import_settings'].include? 'geographic_area_data_origin'
+          if collecting_event.verbatim_latitude && collecting_event.verbatim_longitude
+            Georeference::VerbatimData.create!({
+              collecting_event: collecting_event,
+              error_radius: get_field_value("coordinateUncertaintyInMeters"),
+              no_cached: true
+            }.merge(attributes[:georeference]))
+            location_levels = [collecting_event[:cached_level2_geographic_name], collecting_event[:cached_level1_geographic_name], collecting_event[:cached_level0_geographic_name]].compact
+          else
```
mjy, line 249: `else` migth be OK, but I think we should try in all cases if the option is set to try?

app/models/dataset_record/darwin_core/occurrence.rb
```diff
@@ -428,6 +458,7 @@ def parse_record_level_class
 
     # basisOfRecord: [Check it is 'PreservedSpecimen', 'FossilSpecimen']
     basis = get_field_value(:basisOfRecord)
+    basis = basis.titleize.gsub(' ', '') if basis.include? '_' # Reformat GBIF occurrence download basis of records (e.g., PRESERVED_SPECIMEN to PreservedSpecimen)
```
mjy, line 461: We may need to isolate (re)formatter methods like this in an agnostic lib/utilities/<lib>.rb.

app/models/geographic_area.rb
```diff
@@ -163,6 +163,17 @@ def self.parent
     end
   }
 
+  scope :with_data_origin, -> (data_origin) {
+    where(data_origin: data_origin) unless data_origin.blank?
```
mjy, line 167: You can use the new Ruby pattern `where(data_origin:)` without reference to the value if there is an identically named variable, like in this case.

LocoDelAssembly:
Somehow https://github.com/SpeciesFileGroup/taxonworks/pull/3442/commits/5ac1dacfa04c44cb6b06aabade1bf9e2175594f6 fixes all problems with specs.

app/javascript/vue/tasks/dwca_import/const/geographicalAreaDataOrigins.js
gdower: I changed the GA scope to handle `tdwg` and `ne` and it sorts them so that the finest level available should be on top.

app/models/dataset_record/darwin_core/occurrence.rb
gdower: I could be misunderstanding you, but it does attempt to setup the GA either way. It just prefers using the latitude/longitude if it is provided over trying to use the county, state/province, country fields.

proceps: You would be surprised, but people make mistakes in lat/long as often as in country names.

mjy (approved):

mjy:
@LocoDelAssembly What do you think?  Can we get this into sandboxes for testing?

LocoDelAssembly:
By merging or by having yet another separate docker image? If merging I think we are fine as long as there is no immediate production update planned.

mjy:
Merging to development.  Next release is likely late next week I'm thinking.

PR #3465, opened 2023-06-24
Support fossil taxonomicStatus in DwC taxon importer

LordFlashmeow:
Since it exists, might as well support it...

I also thought it would be useful to also support fossil status as a separate field (`TW:TaxonNameClassification:Iczn:Fossil`), for when taxonomicStatus is `incertae sedis` (since a lot of fossils have uncertainty), since `taxonomicStatus` can only hold one value (which is `incertae sedis`).


app/models/dataset_record/darwin_core/taxon.rb
```diff
@@ -254,6 +254,7 @@ def import(dwc_data_attributes = {})
               excluded: 'TaxonNameClassification::Iczn::Unavailable::Excluded',
               'nomen nudum': 'TaxonNameClassification::Iczn::Unavailable::NomenNudum',
               ichnotaxon: 'TaxonNameClassification::Iczn::Fossil::Ichnotaxon',
+              fossil: 'TaxonNameClassification::Iczn::Fossil',
```
mjy, line 257: @LordFlashmeow Fossil is basically an abstract class at this point, see `ichnotaxon` above for fossil use.

mjy:
See `ichnotaxon` for fossil use.

PR #3468, opened 2023-06-26
DwC importer name handling fixes (incertae sedis, misspelling, types)

LordFlashmeow:
* Fix how misspellings were handled in the checklist staging process
  * add test case for importing names with misspellings
* Fix type scientific name detection in occurrence import when it doesn't match the current scientific name
* Fix incertae sedis imports when a rank was given (via the `TW:TaxonNameRelationship:incertae_sedis_in_rank` field) and had a different rank than the taxon's parent

### DIFF
--- a/app/models/dataset_record/darwin_core/occurrence.rb
+++ b/app/models/dataset_record/darwin_core/occurrence.rb
@@ -844,7 +844,7 @@ class DatasetRecord::DarwinCore::Occurrence < DatasetRecord::DarwinCore
     if type_status_parsed && scientific_name && type_scientific_name.present?
 
       # if type_scientific_name matches the current name of the occurrence, use that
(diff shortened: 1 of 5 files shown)


LocoDelAssembly:
> Fix type scientific name detection in occurrence import when it doesn't match the current scientific name
What name TypeMaterial ends up associated with?

LordFlashmeow:
1. if the type name exactly matches the scientificName, we use the scientificName’s protonym
2. Otherwise, we search for the type name as a cached original combination and use that protonym if it exists

The bug that was happening was that the first check would modify the saved type name in-place, so we couldn’t use it to search original combinations.

mjy:
@LocoDelAssembly can we merge this?

LocoDelAssembly:
If @proceps agrees yes, LGTM.

PR #3476, opened 2023-07-02
Add eventRemarks to DwC export

LordFlashmeow:
To match import behavior, eventRemarks should be included in DwC Exports. Remarks are generated from CollectingEvent notes.

**dwc:eventRemarks**: Comments or notes about the dwc:Event. (https://dwc.tdwg.org/list/#dwc_eventRemarks)

### DIFF
--- a/app/models/collection_object/dwc_extensions.rb
+++ b/app/models/collection_object/dwc_extensions.rb
@@ -93,6 +93,8 @@ module CollectionObject::DwcExtensions
 
       occurrenceRemarks: :dwc_occurrence_remarks,
 
+      eventRemarks: :dwc_event_remarks
+
       # -- Core taxon? --
       # nomenclaturalCode
       # scientificName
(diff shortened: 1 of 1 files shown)


mjy:
@LordFlashmeow The current behaviour is to add `eventRemarks` to TW Notes at present?

LordFlashmeow:
> @LordFlashmeow The current behaviour is to add `eventRemarks` to TW Notes at present?

Yes, I believe it was the replacement for `fieldNotes`.

https://github.com/SpeciesFileGroup/taxonworks/blob/4b51b0d9aa6ecdfa986248fb5f1f23b1026fc626/app/models/dataset_record/darwin_core/occurrence.rb#L689-L693

app/models/collection_object/dwc_extensions.rb
```diff
@@ -161,6 +163,10 @@ def dwc_occurrence_remarks
     notes.collect{|n| n.text}.join('|')
   end
 
+  def dwc_event_remarks
+    collecting_event&.notes.collect {|n| n.text}.join('|')
```
LocoDelAssembly, line 167: `&.` required after `notes`.

mjy:
Trying to remember past conversations about this.  I have been hesistant to map any semantics to TW Note, but I think I remember reading https://dwc.tdwg.org/terms/#dwc:eventRemarks and saying, well, that is what Notes are.  

I do think it would be cleaner to use a Predicate/URI map, but for now the semantics can stand as long as we note the above.

mjy (approved):

PR #3479, opened 2023-07-03
3464 add modified/modifiedBy to CoLDP exporter

gdower:
This adds modified date and modifiedBy (as either orcid or name) to the coldp exporter for all files
It also changes the User.orcid to use identifier instead of cache like the Person model


app/models/person.rb
```diff
@@ -180,7 +180,7 @@ def full_last_name
   # Return [String, nil]
   #   convenience, maybe a delegate: candidate
   def orcid
-    identifiers.where(type: 'Identifier::Global::Orcid').first&.cached
```
mjy, line 183: @gdower is there a difference?  Cached should == identifier I think?  Think of cached as the getter, it's safer in the long run to point to it.  If there is a problem with it currently we should re-factor how it is populated.

lib/export/coldp.rb
```diff
@@ -27,6 +27,26 @@ def self.otus(otu_id)
...
+          project_members[pm.user_id] = pm.user.name
+        else
+          project_members[pm.user_id] = pm.user.orcid.gsub('orcid.org/', '')  # orcid.org domain likely not expected by the checklistbank importer?
+        end
+      end
+      project_members
+    end
+
+    def self.modified(updated_at)
+      updated_at.iso8601()
```
mjy, line 43: We haven't included empty `()` in TW.

lib/export/coldp.rb
```diff
@@ -27,6 +27,26 @@ def self.otus(otu_id)
         .where('(otus.name IS NULL) OR (otus.name = taxon_names.cached)')
     end
 
+    def self.project_members(project_id)
+      project_members = {}
+      ProjectMember.where(project_id: project_id).each do |pm|
+        if pm.user.orcid.nil?
+          project_members[pm.user_id] = pm.user.name
+        else
+          project_members[pm.user_id] = pm.user.orcid.gsub('orcid.org/', '')  # orcid.org domain likely not expected by the checklistbank importer?
```
mjy, line 36: Is there a standard?  I expect our `cached` to match a standard.  Jim was very good at working this out, but we might have added ORCiD after he left.

mjy (commented):
@gdower looks great, thanks.  A couple very minor bits/Qs.

app/models/person.rb
gdower: I generated a Person for my account to test the orcid and cached wasn't set on save, although maybe I'm supposed to manually populate it? Should I also change the User model to point orcid and wikidata_id at cached instead of identifier?

https://github.com/SpeciesFileGroup/taxonworks/blob/development/app/models/user.rb#L462

lib/export/coldp.rb
gdower: It looks like on sfg, the orcids are actually stored as either:

https://orcid.org/0000-0000-0000-0000  (205 records)
orcid.org/0000-0000-0000-0000  (11 records)

At least in the metadata model of ChecklistBank, orcid IDs are provided like 0000-0000-0000-0000: http://api.checklistbank.org/dataset/2144.json

If ChecklistBank ever adds other people resolvers, they're going to need to namespace the IDs so maybe it's better to leave it in.

PR #3489, opened 2023-07-15
Add wildcard subgenus matching for typeStatus in DwC Occurrence import

LordFlashmeow:
Subgenus may be missing from typeStatus strings, but present in TW original combination nomenclature. 

This wildcard search ensures uniqueness among the results, and will not be applied if multiple original combinations match.


app/models/dataset_record/darwin_core/occurrence.rb
```diff
@@ -827,37 +827,80 @@ def parse_location_class
...
           type_type: type_status_parsed[:type].downcase,
           protonym: original_combination_protonym
         }
       end
+
+      # Try wildcard match on subgenus if not present
+      type_name_elements = type_scientific_name.split
+      if type_name_elements.length > 1 && type_name_elements[1].first != "(" && type_name_elements[1].last != ")"
+
+        # append subgenus wildcard to genus string
```
LocoDelAssembly, line 865: Just for peace of mind can you use `Regexp.escape` on each name element before proceeding with next sentences up to `Protonym.where`?

PR #3493, opened 2023-07-16
Create combinations for synonyms with different valid rank in DwC taxon importer

LordFlashmeow:
When a subspecies name changes genus and is synonymized with its species name, no combination is created for [New genus] [species] [subspecies]. Only [New genus] [subspecies] (current placement) and [Old genus] [species] [subspecies] (original combination) exists.

Eg: _Champsomyrmex coquereli minor_ is the original combination, and a junior synonym of _Odontomachus coquereli_. _Odontomachus coquereli minor_ should exist as a combination, because _Champsomyrmex_ is a junior synonym of _Odontomachus_.

Effectively, this creates the combination as it's described in the imported file if we modify the parent & rank to match the senior synonym.

### DIFF
--- a/app/models/dataset_record/darwin_core/taxon.rb
+++ b/app/models/dataset_record/darwin_core/taxon.rb
@@ -154,17 +154,17 @@ class DatasetRecord::DarwinCore::Taxon < DatasetRecord::DarwinCore
 
             original_combination_parents.each do |ancestor|
               ancestor_protonym = ancestor[:protonym]
-              rank = ancestor[:rank]
+              ancestor_rank = ancestor[:rank]
 
               # If OC parent is combination, need to create relationship for lowest element
               if ancestor_protonym.is_a?(Combination)
                 ancestor_protonym = ancestor[:protonym].finest_protonym
               end
 
-              if (rank_in_type = ORIGINAL_COMBINATION_RANKS[rank&.downcase&.to_sym])
+              if (rank_in_type = ORIGINAL_COMBINATION_RANKS[ancestor_rank&.downcase&.to_sym])
 
                 # if the subgenus is newer than taxon_name's authorship, skip it (the name must have been classified in the subgenus later)
-                next if rank&.downcase&.to_sym == :subgenus &&
+                next if ancestor_rank&.downcase&.to_sym == :subgenus &&
                   !ancestor_protonym.year_integer.nil? &&
                   !taxon_name.year_integer.nil? &&
                   ancestor_protonym.year_integer > taxon_name.year_integer
--- a/app/models/dataset_record/darwin_core/taxon.rb
+++ b/app/models/dataset_record/darwin_core/taxon.rb
@@ -243,6 +243,32 @@ class DatasetRecord::DarwinCore::Taxon < DatasetRecord::DarwinCore
 
               taxon_name.taxon_name_relationships.find_or_initialize_by(object_taxon_name: valid_name, type: type)
 
+              if status.to_s == 'synonym'
+                # if synonym and not same rank as valid, and not original combination,
+                # create a combination with the old parent and rank
+
+                if (old_rank = get_field_value('taxonRank').downcase) != rank.downcase &&
+                  metadata['original_combination'] != get_field_value('taxonID') &&
+                  ORIGINAL_COMBINATION_RANKS.has_key?(old_rank.downcase.to_sym)
+
+                  # save taxon so we can create a combination
+                  taxon_name.save
+
+                  # stolen from combination handling portion of code
+                  parent_elements = create_parent_element_hash
+
+                  combination_attributes = { **parent_elements }
+                  combination_attributes[old_rank.downcase] = taxon_name if old_rank
+
+                  # Can't use find_or_initialize_by because of dynamic parameters, causes query to fail because ranks are not columns in db
+                  # => PG::UndefinedTable: ERROR:  missing FROM-clause entry for table "genus"
+                  # LINE 1: ..."taxon_names" WHERE "taxon_names"."type" = $1 AND "genus"."i...
+
+                  taxon_combination_name = Combination.matching_protonyms(**combination_attributes.transform_values { |v| v.id }).first
+                  taxon_combination_name = Combination.create!(combination_attributes) if taxon_combination_name.nil?
+                end
+              end
+
               # Add homonym status (if applicable)
               if status == 'homonym'
                 taxon_name.taxon_name_classifications.find_or_initialize_by(type: 'TaxonNameClassification::Iczn::Available::Invalid::Homonym')
--- a/app/models/dataset_record/darwin_core/taxon.rb
+++ b/app/models/dataset_record/darwin_core/taxon.rb
@@ -383,28 +383,7 @@ class DatasetRecord::DarwinCore::Taxon < DatasetRecord::DarwinCore
 
           else
             # parent is a protonym, look at parents in checklist to build combination
-
-            parents = [get_parent]
-
-            while (next_parent = find_by_taxonID(parents[-1].metadata['parent']))
-              parents << next_parent
-            end
-
-            # convert DatasetRecords into hash of rank, protonym pairs
-            parent_elements = parents.to_h do |p|
-              [
-                # Key is rank (as set in checklist file)
-                DatasetRecordField.where(dataset_record: p)
-                                  .at(get_field_mapping(:taxonRank))
-                  &.pick(:value)
-                  &.downcase&.to_sym,
-                # value is Protonym
-                TaxonName.find(p.metadata['imported_objects']['taxon_name']['id'])
-              ]
-
-            end
-
-            parent_elements.filter! { |p_rank, _| ORIGINAL_COMBINATION_RANKS.has_key?(p_rank) }
+            parent_elements = create_parent_element_hash
           end
 
           combination_attributes = { **parent_elements }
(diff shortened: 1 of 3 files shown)


DaveNicolson:
As a point of clarification, I wanted to ask about this... It sounds to me like this would automatically create name strings that nobody has ever used, which, if I am correct, would be counter-productive.

In the example give, at least one person considered "_Champsomyrmex coquereli minor_" to be a valid subspecies... and at least one person considered that it was not valid, but rather is a junior synonym of the species "_Odontomachus coquereli_"... From what is given, nobody ever considered it a subspecies of _O. coquereli_, or used the name "_Odontomachus coquereli minor_".

So, what is the point of doing this? Are they being tagged somehow as "synthetic"?

I am not a primary user of TaxonWorks, but I know what we at ITIS have absolutely no interest in adding phantom name strings that have never actually appeared in print, and upon discovering such a string in ITIS we would be forced to tag them as database artifacts (which hides it from most users). Better not to distribute names that have never appeared in print, in my view.

LordFlashmeow:
@DaveNicolson 
From my understanding of the importer, every name in the checklist file has been published. Ideally, every name string in the `scientificName` column will exist, either as a cached `Protonym`, cached `OriginalCombination`, or cached `Combination`.

This PR addressed the issue that a name/row marked as a synonym might not get a name string generated under specific conditions (a subsequent combination in a different genus + a different rank than the valid replacement). 

In the test case provided, _Odontomachus coquereli minor_ is a published name: https://antcat.org/catalog/442526

From the taxonomic history shown there, the protonym _minor_ was originally published as _Champsomyrmex coquereli minor_ Emery 1899. Then, it was published as _Odontomachus coquereli minor_ (Emery 1899) by Forel in 1907. Later, it was recognized as a junior synonym of _Odontomachus coquereli_ by Brown in 1976.

Because TW expects the junior synonym protonym to have the same rank as the senior synonym protonym, _Odontomachus coquereli minor_ would be inserted into the database as _Odontomachus minor_, with original combination _Champsomyrmex coquereli minor_. Thus, the name string _O. coquereli minor_ would not be created, despite being explicitly described in the checklist.


Please raise an issue / let me know if you encounter unexpected behavior with the checklist importer. I've done my best to ensure that taxonomic relationships are accurate as possible within the constraints of the DwC checklist spec, but I'm only working with one taxonomic database, so there may be edge cases not present in my dataset.

(To summarize the relationships, here's a simplified version of the checklist used in the example with only the relevant columns)

| taxonID | acceptedNameUsageID | scientificName                | taxonomicStatus      | originalNameUsageID |                                                                                                                                                                                      |
|---------|---------------------|-------------------------------|----------------------|---------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| 1       | 1                   | Odontomachus                  | valid                | 1                   |                                                                                                                                                                                      |
| 2       | 1                   | Champsomyrmex                 | synonym              | 2                   |                                                                                                                                                                                      |
| 3       | 3                   | Odontomachus coquereli        | valid                | 3                   |                                                                                                                                                                                      |
| 4       | 3                   | Odontomachus coquereli minor  | synonym              | 5                   | As a synonym, this will be elevated to species to match the senior synonm's rank.                                                                                                    |
| 5       | 3                   | Champsomyrmex coquereli minor | obsolete combination | 5                   | This is the original combination, which will be generated during the import of taxonID 3. No `Combination` will be created when this row imports, because the OC will already exist. |
|         |                     |                               |                      |                     |                                                                                                                                                                                      |

proceps:
TaxonWorks does not require creation of unpublished name string. It just strictly following the rules of ICZN nomenclature. According to ICZN, two names could be recognized as synonyms only at the same rank level. So, if somebody states that a subspecies is treated as a synonym, affectively, two separate acts are proposed (even if not stated), the subspecies is elevated to a rank of species and two species are synonymized at the same nomenclatural rank. It does not mean that we have to create artificial strings for the names.

DaveNicolson:
The Principle of Coordination (Art. 46 for species) does mean that any species-group name made available is simultaneously available at all ranks within the species group (practically speaking, species and subspecies ranks), but I have never heard of anyone using it to form & distribute combinations never seen in print. And yes, the Code's definition of "synonym" per the Glossary does specify that synonyms are names at the same rank, but there is nothing about forming novel species-group name strings by implication.

Let me provide a purely hypothetical example, since the example above turned out to have published combinations beyond those initially mentioned (thanks for clarifying that!!).

Subspecies _Aus bus cus_ Smith, 2000 is coined, and someone later decides it is not an example of the species _Aus bus_, but is rather the same taxon as was previously known as _Nus zus_ Jones, 1900, that the taxon in question is a single species without recognized subspecies, in the genus _Nus_, and that no other names are involved.

Clearly the taxon in question must now be known as _Nus zus_ Jones, 1900, since "_Nus cus_ (Smith, 2000)" would be junior. One could argue that _Aus cus_ Smith, 2000 should be included in synonymy due to Art. 46 (that species name combination is deemed to exist) and the definition of "synonym", as a matter of some kind of bookkeeping. But that's about as far as I would personally want to go based on the info above (and I would not even do that, not without assurances that doing so serves some purpose that cannot otherwise be fulfilled).

I would not want to see the potential name string "_Nus cus_ (Smith, 2000)", or the nonsense combination "_Nus bus cus_ (Smith, 1900)", as neither has been published. They certainly could have been published (e.g., if the genus placement were addressed before identifying the 2 names as conspecific), but were not, at least in this hypothetical example.

I guess I am still wondering at this point, in this "example" (a very particular circumstance, to illustrate the point), what name strings other than _Aus bus cus_ Smith, 2000 and _Nus zus_ Jones, 1900 would you want to provide in synonymy, and why? That is, for what purpose?

Sorry if I'm disrupting your efforts to move things forward (and thanks for indulging my questions), but I'm hoping to better understand what's happening and why, when/if name strings that may not have been seen in print are used/distributed.

mjy:
@DaveNicolson regardless of what's going on, big thanks for following along and chiming in, it's much appreciated.

mjy:
A couple observations,

* The conversation may be referencing seperate issues at this point?  Let's ensure that the code does what we'd logically assume it should, as a first pass.
* Checklists of names can  originate from sources that don't follow any rule (e.g. be derived from names observed on taxon determinations in collections, that may have been formed in error, etc.).  So @LordFlashmeow observation "everything in the list is published" is a safe assumption.   We are naievly importing here, whether or not the data are "true" may depend on future considerations.
* There are 3 ICZN ways to represent names in TW- Protonyms, Combinations, and Original Combinations. 
* Each of the scienticNames must be represented in one of these forms.
* Iin as much as the code @LordFlashmeow lets us represent those checklist names then it's OK.  If it is creating intermediates that are not observed in the checklist then that is a (potential) issue.

* Side note- If you want to maximally preserve the form of the name as applied to (some specimen/use in the literature), then we need to create a Combination and link that to an OTU.

proceps:
@DaveNicolson For your example, _Aus bus cus_ as a synonym of _Nus zus_, I would just create two protonyms and one synonym TaxonName relationship. '_cus_' would still be transferred to the genus _Nus_ and elevated to the rank of species automatically. You would not see _Nus zus_ combination on TaxonPages, but this a current classification of a taxon. In this combination, the name compete for the priority as a second homonym of _zus_ in the genus _Nus_. Any other _zus_ described in _Nus_ after it will be treated as a secondary homonym. So, even so no subsequen combination is published, the placement of the species into a different genus is assumed. Again, this is strictly cached fields which we use to indexing of the names and validation of nomenclatural rules. If you check the paper catalogue output, you would not see any combinations which were not published in the literature.

DaveNicolson:
Thanks for the response. So there are names being stored that are there for technical reasons. That is good to understand. Are such names visible to users? Would they be passed through to GBIF and/or COL (etc.)? I'm just trying to understand the ramifications for data consumers (including ITIS).

mjy:
Throughout TaxonWorks when you see `cached_<field>` in a model these are computed for, well, computation purposes.  We may share portions or all of these fields if the computations that go into populating the cached_field are the same as those that would be needed to compute a value to share.

rogerburks:
At least the Principle of Coordination part ought to be relevant to what I was discussing in #3495 . I don't want to hijack too much, but I just wanted to highlight that every name covered by the Code is subject to that, and that this situation results in the creation of two potentially different kinds of taxon name entities: In every case, the Principle of Coordination results in a set of simultaneously published name forms across all group ranks of the name. The "two types" comes into play here: One type of name exists in the literature, and the other type does not. Generally, when we publish a genus name, we are not especially worried about making sure that its subgenus, infragenus, etc. forms are mentioned in the literature. However, these names exist and automatically come into play in situations when another subgenus is used as valid within the genus in question. That is, we don't have a choice. If subgenus B is used as a valid child of genus A, then the nominotypical subgenus A is the only correct option to use when we want to reference the subgeneric placement of at least the type species of genus A.

proceps:
We are taking the Principle of Coordination into account. All coordinate name should share the same original source, type, spelling, etc. We add those properties from one name to another when needed/missing. And yes, nominotypical subgenus could be automatically created when any other subgenus is added to the genus. And those names are created only when needed.

rogerburks:
I understand this, although unexpected behavior of the names seems to occur. Possibly this is only due to the way that we are curating our data, however.

DaveNicolson:
I am pretty sure we are on multiple tangents at this point, and no longer focused on the issue this ticket was about, but in any case, I am still hoping for comments responsive to these questions about the stored/cached names not found in the literature:

Are such names visible to users?

Would they be passed through to GBIF and/or COL (etc.)?

I'm just trying to understand the ramifications for data consumers (including ITIS).

proceps:
@DaveNicolson I replicated your example in TW:
That what it looks like in Browse nomenclature:
image
That what it looks like when you use search box:
image
This is filter interface:
image
The last one is close to what is sent to CoL. 
In all cases, remember that each protonym has current classification and original classification. It is all depends which string we display in each particular interface. When we render a paper style catalogue, you will see something similar to the Browse nomenclature interface. My understanding, CoL should also display the synonyms in their original combination format.

DaveNicolson:
@proceps Thanks for taking the time to show some of the TW internal workings, and to everyone for comments on the same (very interesting!). My primary concerns at this point are (1) what gets shown in TW to public users (I know there's no major public web interface at this point), and especially(2) what names get sent particularly users like COL (and ITIS, if an export suitable for nearly direct import into ITIS' workbench tool). Re (2), Are you able to generate COLDP output for that simple example without sending it to CLB/GBIF/COL/Anyone, just to see what names come through in it? You're welcome to email it to me if you do! _If_ non-accepted names not found in any publication are included, are they flagged in any way so that users can exclude them if they want or need to?

proceps:
@DaveNicolson I sent you the COL output few days ago, could you please confirm that you received it.

DaveNicolson:
@proceps I can't find anything in my email, including junk folder...

PR #3503, opened 2023-07-28
Add option to enable organization determiners in import

LordFlashmeow:
I'm not 100% sure about the best wording for the setting, "Enable searching for Organization name in determinedBy field" is the best I could come up with.

I added a second setting to enable matching on alternate name to help users disambiguate multiple organizations with the same name, and fingers crossed we don't need anything more complex.

### DIFF
--- a/app/models/dataset_record/darwin_core/occurrence.rb
+++ b/app/models/dataset_record/darwin_core/occurrence.rb
@@ -397,8 +397,38 @@ class DatasetRecord::DarwinCore::Occurrence < DatasetRecord::DarwinCore
     value
   end
 
+  # Parse for names in a given field and find or create one or more Person::Unvetted (scoped to the import dataset).
+  # Also supports searching for an Organization, if the column value is an exact match for the Organization's name or alternate_name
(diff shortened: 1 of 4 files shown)


jlpereira (approved):

mjy:
@LordFlashmeow We can get this in if you can double check the one failing spec - https://github.com/SpeciesFileGroup/taxonworks/actions/runs/5686746266/job/15414071820?pr=3503#step:7:5971

app/models/dataset_record/darwin_core/occurrence.rb
```diff
@@ -553,7 +589,7 @@ def parse_occurrence_class
     # recordNumber: [Not mapped]
 
     # recordedBy: [collecting_event.collectors and collecting_event.verbatim_collectors]
-    Utilities::Hashes::set_unless_nil(res[:collecting_event], :collectors, (parse_people(:recordedBy) rescue nil))
+    Utilities::Hashes::set_unless_nil(res[:collecting_event], :collectors, (parse_people(:recordedBy, search_organizations=false) rescue nil))
```
LocoDelAssembly, line 592: `parse_people` doesn't seem to have more than one param?

PR #3523, opened 2023-08-15
3522 improve coldp export delimiter

gdower:
Sanitizes remarks fields to eliminate tabs and line breaks, and changes export file extension from csv to tsv

### DIFF
--- a/lib/export/coldp.rb
+++ b/lib/export/coldp.rb
@@ -47,6 +47,10 @@ module Export
       project_members[updated_by_id]
     end
 
+    def self.sanitize_remarks(remarks)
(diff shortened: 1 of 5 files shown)


lib/export/coldp.rb
```diff
@@ -47,11 +47,15 @@ def self.modified_by(updated_by_id, project_members)
       project_members[updated_by_id]
     end
 
+    def self.sanitize_remarks(remarks)
```
mjy, line 50: @gdower This is OK here for now, but we should move it to /lib/utilities/strings.rb

PR #3632, opened 2023-11-09
Raise better error message when typeStatus name is a homonym in importer

LordFlashmeow:
Previously, error would be "Type cannot be designated, name is not a species group name".

### DIFF
--- a/app/models/dataset_record/darwin_core/occurrence.rb
+++ b/app/models/dataset_record/darwin_core/occurrence.rb
@@ -953,11 +953,17 @@ class DatasetRecord::DarwinCore::Occurrence < DatasetRecord::DarwinCore
           type_type: type_status_parsed[:type].downcase
         }
       end
-      if (original_combination_protonym = Protonym.find_by(cached_original_combination: type_scientific_name, project_id: self.project_id))
+      original_combination_protonyms = Protonym.where(cached_original_combination: type_scientific_name, project_id: self.project_id)
+      if original_combination_protonyms.count == 1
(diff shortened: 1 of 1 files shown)


app/models/dataset_record/darwin_core/occurrence.rb
```diff
@@ -953,11 +953,17 @@ def parse_typestatus(type_status, taxon_protonym)
...
-      if (original_combination_protonym = Protonym.find_by(cached_original_combination: type_scientific_name, project_id: self.project_id))
+      original_combination_protonyms = Protonym.where(cached_original_combination: type_scientific_name, project_id: self.project_id)
+      if original_combination_protonyms.count == 1
         return {
           type_type: type_status_parsed[:type].downcase,
-          protonym: original_combination_protonym
+          protonym: original_combination_protonyms.first
         }
+      elsif original_combination_protonyms.count > 1
+        potential_protonym_strings = original_combination_protonyms.map { |proto| "[id: #{proto.id} #{proto.cached_original_combination_html}]" }.join(', ')
```
LocoDelAssembly, line 963: Does it actually render as HTML or thinks like `<i>Jivarus alienus</i>` are displayed literal instead?

mjy: We should not reference _html version, use the non in all comparisons.

LordFlashmeow: It renders the HTML in the error message:
image

PR #3649, opened 2023-11-19
Search for URL in institutionCode, better search error messages

LordFlashmeow:
Add more helpful error message when multiple repositories match the given `institutionCode` (either by acronym, URL, or name).

### DIFF
--- a/app/models/dataset_record/darwin_core/occurrence.rb
+++ b/app/models/dataset_record/darwin_core/occurrence.rb
@@ -540,15 +540,48 @@ class DatasetRecord::DarwinCore::Occurrence < DatasetRecord::DarwinCore
     # institutionCode: [repository.acronym] # TODO: Use mappings like with namespaces here as well? (Although probably attempt guessing)
     institution_code = get_field_value(:institutionCode)
     if institution_code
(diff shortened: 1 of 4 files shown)


app/models/dataset_record/darwin_core/occurrence.rb
```diff
@@ -540,15 +540,48 @@ def parse_record_level_class
...
+      acronym_repositories = Repository.where(acronym: institution_code)
+      if acronym_repositories.count == 1
+        repository = acronym_repositories.first
+      elsif acronym_repositories.count > 1
+        error_messages << "Multiple repositories with acronym #{institution_code} found."
+      else
+        error_messages << "No repositories with acronym #{institution_code} found."
+      end
+
+      if !repository && (institution_code.starts_with?('http://') || institution_code.starts_with?('https://'))
```
LocoDelAssembly, line 554: Would perhaps be best to test URL first, then acronym so you avoid hitting the database twice (first with potential URL string than cannot possibly be an acronym)?

LordFlashmeow: Agreed.

PR #3657, opened 2023-11-21
Fix get_protonym for missing intermediate ancestor

LordFlashmeow:
`nil` values for authorship details in the `where` hash were breaking the search.

This would cause the "Create if Missing" strategy to duplicate existing protonyms if it had authorship info while the new record didn't (likely because it was an intermediate name).


LocoDelAssembly:
@LordFlashmeow would be possible for you to resolve the merge conflicts?

Issue #3660, opened 2023-11-22
[Bug]: DWCA Checklist Importer: fails to import records with no scientificNameAuthorship


gdower:
### Steps to reproduce the bug

1. When clicking the import button on a DWCA record with an empty scientificNameAuthorship
2. A javascript error modal is raised with the below stack trace

```
[...]

PR #3662, opened 2023-11-22
Check for authorship presence before attempting match

LordFlashmeow:
Resolves the original issue described in #3660.

### DIFF
--- a/app/models/dataset_record/darwin_core/taxon.rb
+++ b/app/models/dataset_record/darwin_core/taxon.rb
@@ -57,23 +57,25 @@ class DatasetRecord::DarwinCore::Taxon < DatasetRecord::DarwinCore
         year = nil
 
         # split authorship into name and year
-        if nomenclature_code == :iczn
-          if (authorship_matchdata = authorship.match(/\(?(?<author>.+),? (?<year>\d{4})?\)?/))
+        if authorship.present?
+          if nomenclature_code == :iczn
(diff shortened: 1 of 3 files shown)


app/models/dataset_record/darwin_core/taxon.rb
```diff
@@ -57,23 +57,25 @@ def import(dwc_data_attributes = {})
         year = nil
 
         # split authorship into name and year
-        if nomenclature_code == :iczn
-          if (authorship_matchdata = authorship.match(/\(?(?<author>.+),? (?<year>\d{4})?\)?/))
+        if authorship.present?
+          if nomenclature_code == :iczn
+            if (authorship_matchdata = authorship.match(/\(?(?<author>.+),? (?<year>\d{4})?\)?/))
```
LocoDelAssembly, line 62: Are you specifically looking for original combination author of new combination names or you'd rather need authorship of any name? (Question is because it looks you looks for authorship in parentheses?)

LordFlashmeow: We're looking for any authorship, the parentheses are optional

LordFlashmeow: We should be leveraging `Biodiversity::Parser` authorship info if it's available, but fixing the nil error was my priority.

LocoDelAssembly: Ah sorry, missed that parens are optionally matched.

PR #3675, opened 2023-11-30
Add person/123.json `role_counts` extension to make Uniquify People task faster

LordFlashmeow:
The new `role_counts` extension for `/people/:id.json` returns information about the number of roles a Person has, grouped by type (Collector, Determiner, etc) and split by in-project usage, non-project usage, and community uses.

Sample:
```
{
  "in_project": {
    "Collector": 6
  },
  "not_in_project": {
    "Collector": 18
    "Determiner": 3
  },
  "community": {}
}
```

With this data, the Uniquify People task can be optimized, by querying for number of roles instead of getting all roles for a Person and counting them.

The only place where the actual role ids are needed is in `RoleDescription`, which renders the list of roles (for a specific type). 

Getting the list of role ids can be performed by `addSelectPerson`, which means we can move it out of `findPeople`. As a result, executing the "Selected people" action is significantly faster. 

I think it's safe to remove `roles` from the first part of `processMerge()`, since those people are being deleted anyways, and the "selected person" has its attributes re-fetched once the merge is complete.

Ideally, we could postpone the fetch for role ids until the user marks the person as "to be merged" with the checkbox in the `Match People` section, but I don't know enough about Vue to make that happen. Instead, I've left the behavior as-is (fetching roles for all matched people).

In places where the Person objects are extended with `roles`, it's faster to use `role_counts` to get role names, particularly when there are many roles (1k+).

After Gitter discussion, I also removed explicit database `exists?` query when checking if role is in project, which should make the remaining role queries much faster.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ This project <em>does not yet</em> adheres to [Semantic Versioning](https://semv
 - BiologicalAssociations as raw TaxonWorks data`/api/v1/biological_associations.csv`
 - BiologicalRelationships as raw TaxonWorks data`/api/v1/biological_relationships.csv`
 - DwC ResourceRelationship extension (preview) [#2554]
-- Taxonmy summary to CollectionObject summary report
+- Taxonomy summary to CollectionObject summary report
 - Metadata summary report from Filter BiologicalAssociations
 - Biological associations simple table preview, sortable columns [#1946]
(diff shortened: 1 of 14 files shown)


mjy:
Going to merge, pinged @jlpereira about

> Ideally, we could postpone the fetch for role ids until the user marks the person as "to be merged" with the checkbox in the Match People section, but I don't know enough about Vue to make that happen. Instead, I've left the behavior as-is (fetching roles for all matched people).

PR #3688, opened 2023-12-04
3687 dwca fix nil parent

gdower:
@LordFlashmeow, this fixes an import crash on the DwC-A checklist importer caused by a nil value for core_records[oc_index][:parent]. One of the added tests will fail because there's no missingParent error message yet.


LordFlashmeow:
Currently, it's not possible to change a parent after the import is staged. Since it's not possible to re-stage checklists, any record that encounters errors during the staging process will never be able to be imported. Does it make sense to change the error from `Not Ready` to `Failed` for these records?

gdower:
I think it might be better to aim for being able to update the IDs in table UI in the long-run and refreshing the row statuses on the interdependent rows, although I know it's currently not possible and challenging with the row interdependencies.

If the row cannot import because a dependent row is not imported, I think it might make sense to keep it in the NotReady state, but right now it doesn't provide a message that it is NotReady because of the dependent row isn't imported. If you import the test dataset to as far as you can get and click on the NotReady status (in the screenshot), there is no modal message:

image

The remaining Ready record also doesn't import and also doesn't go to Failed state, so this branch still doesn't behave right but it at least stops the importer from crashing in staging.

gdower:
I adjusted the spec to make it fail on purpose because currently the importer doesn't seem to handle cases in which: 1) the OC's parent genus record is totally missing and needs to be setup, or 2) the parentNameUsageID is missing on the OC record but the OC genus record could be looked up based on the name like in this example:

image

If the importer can handle cases like that, it might be easier than trying to communicate via the UI that a subsequent combination record is NotReady because its referenced originalNameUsageID record has a missing parent, or missing or invalid parentNameUsageID.

PR #3692, opened 2023-12-04
Handle missing originalNameUsageID in checklist import

### LINKED ISSUE #3680: [Bug]: DwC-A Checklist importer: Subsequent combinations fail to import if the originalNameUsageID is unknown
### Steps to reproduce the bug

1. When importing a DwC-A Checklist
2. If the originalNameUsageID is unknown, subsequent combinations fail to import with the error, "Could not determine if name is protonym or combination"

### Screenshot

image

### Expected behavior

It would be preferable to still import subsequent combination records in which the originalNameUsageID is unknown, and use the verbatim author field to set the bracket on the authorship like (Walsh), because after the import is completed it will be easier to filter for incomplete combination relationships with the Filter Nomenclature Task and fill in the missing original combination data with the taxon name editing tools instead of trying to correct it in the import task in which you'd have to fill in a large number of IDs. Often the names missing the originalNameUsageID are really old names in which it's hard to find/access the original genus because the literature has been long out of print, so importing the currently used subsequent combination seems more important than having the original combination data.

### Additional Screenshots

_No response_

### Environment

Sandbox

### Sandbox Used

sandbox

### Version

4a941a139

### Browser Used

Firefox

LordFlashmeow:
Resolves #3680

Protonyms without `originalNameUsageID` will not have an original combination created. Subsequent combinations may not work without it, as `originalNameUsageID` is used to determine the protonym for combinations.

**TODO**: 
- [x]  add tests

### DIFF
--- a/app/models/dataset_record/darwin_core/taxon.rb
+++ b/app/models/dataset_record/darwin_core/taxon.rb
@@ -158,7 +158,9 @@ class DatasetRecord::DarwinCore::Taxon < DatasetRecord::DarwinCore
           end
 
           # make OC relationships to OC ancestors
-          unless metadata['parent'].nil? # can't make original combination with Root or if matching pre-existing taxon name
+          # can't make original combination with Root or if matching pre-existing taxon name
+          # Do not make original combination if we assumed the value.
+          if metadata['parent'].present? && (metadata.fetch('create_original_combination', true) == true)
 
             # loop through parents of original combination based on parentNameUsageID, not TW parent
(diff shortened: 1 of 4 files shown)


gdower:
Thanks, @LordFlashmeow! With this branch merged with the latest development, the gall wasps dataset successfully imports 1740 records with only 19 not ready and 50 errored. I could add tests if you'd like?

LordFlashmeow:
@gdower Feel free to add additional tests, otherwise I think this is ready to merge.

PR #3694, opened 2023-12-05
Refactor DwC setting checkboxes into reusable component

LordFlashmeow:
Many DwC import settings are simple checkboxes that set a metadata value, and almost all the code for these components is the same, with only the setting name and description changed.

Having a generic checkbox setting component will make it much easier to add and manage import settings.


LordFlashmeow:
Implemented by ab034ea0564857e1929974f9d792272d5663f47f

PR #3695, opened 2023-12-05
Mark staged rows with errors as Failed

LordFlashmeow:
Errors that occur during staging are  because we don't support re-staging, so we should mark them as `Failed` instead of `NotReady`


LocoDelAssembly:
`Failed` is to indicate that unhandled error occurred. The ones in this situation don't appear as Not Ready in red since an error message is present? That differentiate from the others that are not ready because dependencies are not yet imported.

LordFlashmeow:
I'd like to communicate to the user that the records have encountered an error that makes them impossible to import. 

I agree that it shouldn't be `Failed`, but `Not Ready` seems to imply that they might be ready in the future. I guess it doesn't matter right now (although it might be worth revisiting if we implement restaging or dynamic dependency checking).

PR #3701, opened 2023-12-11
Fix Project.nuke deletion order and scoping

LordFlashmeow:
CEs cannot be deleted while dependent COs exist, so deleting CEs must be postponed until after the COs are deleted.

`Download` uses `default_scope` which causes `delete_all` to only delete records that haven't expired yet. `unscoped` disables the `default_scope` behavior for any class that implements it.


mjy:
Thanks @LordFlashmeow

PR #3732, opened 2023-12-20
Add caste to DwC export

LordFlashmeow:
Caste has been added to DwC (https://github.com/tdwg/dwc/issues/412), so we should support exporting a caste biocuration if it is mapped.


mjy:
@LordFlashmeow please stub a new VERSION entry for this- https://github.com/SpeciesFileGroup/taxonworks/blob/development/lib/export/dwca.rb#L14.

jlpereira (approved):

PR #3737, opened 2023-12-21
Proof of concept for exporting OTU name in DwC export

LordFlashmeow:
This is the brute-force approach, calling a method on every CO in the export. Not efficient, but very flexible/extensible.

I don't expect this to be merged, I think it's just a starting point for discussion.

- [x] Backend tests
- [x] Frontend/UI

### DIFF
--- a/app/jobs/dwca_create_download_job.rb
+++ b/app/jobs/dwca_create_download_job.rb
@@ -7,10 +7,10 @@ class DwcaCreateDownloadJob < ApplicationJob
   # take a download, and a list of scopes, and save the result to the download, that's all
   # @return
   # # TODO: handle extension scopes
-  def perform(download, core_scope: nil, extension_scopes: {biological_associations: nil, media: nil}, predicate_extensions: {}, taxon_works_extensions: {})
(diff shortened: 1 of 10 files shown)


app/jobs/dwca_create_download_job.rb
```diff
@@ -7,10 +7,10 @@ class DwcaCreateDownloadJob < ApplicationJob
   # take a download, and a list of scopes, and save the result to the download, that's all
   # @return
   # # TODO: handle extension scopes
-  def perform(download, core_scope: nil, extension_scopes: {biological_associations: nil, media: nil}, predicate_extensions: {})
+  def perform(download, core_scope: nil, extension_scopes: {biological_associations: nil, media: nil}, predicate_extensions: {}, taxon_works_extensions: {})
```
mjy, line 10: taxon_works -> taxonworks

LordFlashmeow:
How should these extensions be defined and presented to the user?

The predicates are just an API GET, but these are custom-defined so they can't be queried for in the database. Should we make an endpoint, or just define the list of available extension methods in the frontend?

For the selection UI, should the methods be displayed in a list like the predicates are? I don't know how many projects will use them, or how accessible they should be.

Issue #3739, opened 2023-12-21
dwc Importer: map dwc identificationRemarks and identificationQualifier to notes on TaxonDetermination


debpaul:
### Feature or enhancement

When I'm using the DwC Importer, I'd like to have these following properties map to the **notes** for Taxon Determination

- dwc:identificationRemarks
- dwc:identificationQualifier

1. IOW, in the DwC Importer, which "Taxon Name" provided in the dataset, will @LocoDelAssembly point these "notes" to? There's no "taxon name" column. @mjy Please clarify. Thanks!
2. Once these are mapped to Taxon Determination "notes", please what does the header look like in the DwC Importer dataset to be uploaded?

### Location

DwC Importer

### Your role

Liaison

mjy:
@LordFlashmeow @LocoDelAssembly 

* `identificationRemarks` can be mapped to a Note on TaxonDetermination.

@debpaul please add  `identificationQualifier` as its own issue, it requires future thought as to how to best map to our Confidence CVTs and doesn't really fit in TW model as a seperate propertie (field). We include this data in Otu#name so people wanting to import those assertions should do so there. I don't forsee this changing until someone formalizes Qualifiers (and also notes how they are/not the same as TaxonNameRelationship assertions).

LordFlashmeow:
> identificationRemarks can be mapped to a Note on TaxonDetermination.

I assume it should only get Notes for the latest determination, not current and historical determinations?

debpaul:
> `identificationRemarks` should already be import as a determination note, it's a bug if that's not happening.

@LordFlashmeow thank you! 
At some point maybe there's some bandwidth to try and update the fields the Occurrence and Checklist Importers support in our docs.taxonworks.org documentation. I'd be happy to add them, if I know which ones to add (that have been added since by you or @LocoDelAssembly). See [mappings](https://docs.taxonworks.org/guide/import-and-export.html#dwc-terms-mappings)

PR #3742, opened 2023-12-21
Export DwC identificationRemarks

LordFlashmeow:
Inspired by #3739, `identificationRemarks` should be generated from the notes associated with the most recent `TaxonDetermination`.


mjy:
Please add a note on the DWC VERSION constant re this change.

LordFlashmeow:
> Please add a note on the DWC VERSION constant re this change.

Is it ok to just add a comment? It feels silly to add a new version 8 hours after the last one.

mjy:
Yes, comment OK, I will update the version timestamp at the point of actual release.

Issue #3735, opened 2023-12-20
New task/workflow: Batch update collectors from verbatimCollectors or buffered_collecting_event


LordFlashmeow:
### Feature or enhancement

The DwC importer's name parsing isn't perfect, and a portion of the CEs don't have Collectors assigned.

I want to be able to set the collectors of many CEs at once. Configuring a `batch_update` endpoint is the faster approach, but a new stepwise task would be more effective in the long run.

It should support assigning multiple collectors to the events respect the order of People provided.

### Location

Filter CE or a new task

tmcelrath:
Related to #1836 as well. 

I support the creation of this. Right now I can't batch update collectors from the data I get from Bionomia because I do not have this functionality.

LordFlashmeow:
Implemented in #3745. `buffered_collecting_event` isn't supported yet, but could be in the future.

PR #3744, opened 2023-12-22
Add "Append collectors" slice to CE batch update radial

LordFlashmeow:
Related to #3735, this is the "barebones" way to assign collectors to multiple CEs at once. 

It appends collectors to the end of the existing list. If you try to add a collector that is already present in a CE, the row will error no collectors will be added to that CE.

Order of names in the role picker input is respected.

### DIFF
--- a/app/javascript/vue/components/radials/ce/components/CollectorSlice.vue
+++ b/app/javascript/vue/components/radials/ce/components/CollectorSlice.vue
@@ -14,19 +14,36 @@
       <h3>{{ count }} records will be updated</h3>
 
       <fieldset>
-        <legend>Geographic area</legend>
-        <SmartSelector
-          model="geographic_areas"
-          label="name"
-          :target="COLLECTING_EVENT"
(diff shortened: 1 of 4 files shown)


app/javascript/vue/components/radials/ce/components/CollectorSlice.vue
```diff
@@ -0,0 +1,130 @@
...
+  collecting_event: {
+    roles_attributes:
+      // remove position so roles are appended to end of list
+      collectingEvent.value.roles_attributes.map(({position, ...rest}) => rest)
+  }
+}))
+
+function updateMessage(data) {
+  const message = data.sync
+    ? `${data.updated.length} collection objects queued for updating.`
```
jlpereira, line 125: Should be "collecting events"?

LordFlashmeow: Yup, that one escaped my search and replace.

Issue #3724, opened 2023-12-18
DwC Exporter: Expose internal (non-dwc) fields for export


LordFlashmeow:
### Feature or enhancement

As someone who wants to export data out of TW, I need a way to include non-DwC fields in the DwC exporter (like `elevation_precision`).

Ideally it would use the same modal interface (and API) as for selecting predicates, so that it's clear to other users that the fields are optional and not part of the DwC spec.

### Location

DwC Export

PR #3750, opened 2023-12-27
Refactor TW extensions in DwC export to support plucking direct from tables

LordFlashmeow:
Plucking from the database is faster than calling a method on each CO when the data do not require any processing.

The change is transparent to the API consumer, all extension fields are still provided as a single list.

Included `elevation_precision` as a sample of an exposed field.

Other internal CE and CO fields can be added as needed, resolving #3724.


mjy:
@LocoDelAssembly I approve this for merge.

mjy:
@LocoDelAssembly LGTM for merge.

Issue #3366, opened 2023-03-24
Task - Nomenclature stats - combination column is always visible


proceps:
The combination column always displayed even when the checkbox is unselected.
![image](https://user-images.githubusercontent.com/5597748/227624958-7b0ae626-b53f-4464-81b1-c91d7f6a707b.png)

PR #3751, opened 2023-12-29
#3366 Respect the Combinations checkbox when displaying the Nomenclat…

kleintom:
…ure Stats table

We now display the Combinations column in the Nomenclature Stats table only when the Combinations checkbox is checked.

### DIFF
--- a/lib/queries/taxon_name/tabular.rb
+++ b/lib/queries/taxon_name/tabular.rb
@@ -307,7 +307,9 @@ module Queries
           end
         end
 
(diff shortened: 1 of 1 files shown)


mjy:
LGTM.  @kleintom did this test out OK in the UI?

jlpereira:
> LGTM. @kleintom did this test out OK in the UI?

I tested it and it seems to work fine.

mjy:
@kleintom thanks so much!

Issue #3262, opened 2022-12-21
[Bug]: Task - Browse Collection Objects: recently added 'Go' arrows do not work


AntWeb-org:
### Steps to reproduce the bug

```bash
Based on [#3229](https://github.com/SpeciesFileGroup/taxonworks/issues/3229) submission workaround.

From the BCO task, you can click the Go button to enable the Navigate popup in order to choose Previous/Next by ID, but the arrows allowing the scroll through IDs are grayed out/inoperable.

@bpescador, @chrisgrinter, @rich-keller
```

### Screenshot

![BCO_go](https://user-images.githubusercontent.com/74991451/208987730-7f297c2a-ae90-4119-9768-9cef74ca3f83.jpg)

### Expected behavior

I would like the arrows to scroll Previous/Next by ID. Perhaps similar to CSD ID by identifier. see additional screenshot

### Environment

Sandbox

### Version

build 40d115751 on 2022-12-20T13:17:02Z

### Browser Used

Chrome Version 108.0.5359.99

AntWeb-org:
![CSD_ID](https://user-images.githubusercontent.com/74991451/208989645-cc464096-e217-4395-ab5b-d30fc10e4536.jpg)

mjy:
Arrows turn on when there are options available. It appears in this case we couldn't calculate previous and next. We'll have to see why catalog-numbers (by default it starts with these) are not usable in this context.

PR #3761, opened 2024-01-05
#3262 Get navigation arrows in Browse COs task working

kleintom:
The `navigate.value.previous_by|next_by` objects get loaded with an `id` key, not `identifier`. I think the other change was just a copy/paste typo.


kleintom:
This gets the left/right arrows on the Browse Collection Objects task working, but they don't do quite what I was expecting.

What I was thinking would happen: If you click on Go and choose, say "Next without collecting event", then the next observation that doesn't have an associated collecting event gets loaded. After that I expected that the left and right arrows would continue cycling through only other observations without a collecting event.

What actually happens: the left and right arrows always only cycle through observations by id, i.e. through _all_ observations in the order they were added.

If my expected behavior is the way it's actually supposed to work then I can create a new issue if/when this one lands, or someone else can if they'd rather.

PR #3766, opened 2024-01-08
#3765 Edit the browse_annotations route and view folder to get the Br…

kleintom:
…owse Annotations task to load

This seems to be necessary since `class Tasks::BrowseAnnotationsController` was changed to `class Tasks::ObjectAnnotations::BrowseAnnotationsController`.


kleintom:
This seems to fix the issue, but my knowledge of Rails is pretty shallow so if there's a better way to do this please let me know or feel free to close this pr and take the issue.

mjy:
Thanks, as noted, this is zeitwerk related, and there are likely others issues like this.

PR #3772, opened 2024-01-10
Update text in Loans dashboard for Created in... section

kleintom:
Fixing a typo:
image

(Feel free to let me know if you'd like an associated issue for these small ones, or if I should just bundle them with some other issue, or something else.)


mjy:
Small PRs are fine.  We'll let you know if they get too big/unweildly, so you can combine if it makes your contributions (thanks for these BTW) easier, and we can find a balance.

Issue #3768, opened 2024-01-10
Browser history is broken on the Browse collection object task


kleintom:
### Steps to reproduce the bug

```bash
1. In the Browse collection task, Use the blue navigation arrows or the Go options to navigate through multiple COs. Note the url in your browser changing as you do so.
2. Now go back in your browser.

Expected results: As I click back in the browser and the url changes, the CO corresponding to the changed url is loaded.

Actual results: When I click back in the browser the url changes but the CO loaded on screen does not (until you get back to the CO url you started on).

If I'm understanding correctly this changed with the referenced commit when COs were changed to load in place in the browser as you browse (which I personally love!) instead of loading a new page each time you clicked to browse.
```

### Environment

Development (docker)

### Version

e3db6c1000ad19ce3ad7fc85a060182052236142

jlpereira:
@kleintom after merge your PR I created a composable to handle popstate event, it will automatically remove the listener when vue applications are unmounted

kleintom:
Nice! I got as far as removing the popstate handler in turbolinks:before-render, but I never thought to call unmount to do it in there, this is much cleaner and idiomatic, thanks for adding it.

PR #3777, opened 2024-01-12
Handle popState for the Browse CO task #3768

kleintom:
The trick here is that when you first load the Browse CO task, turbolinks adds a history state object (seen here from a popstate event object):
turbolinks
Apparently that state has the effect of causing a reload everytime browser history returns to the inital browse CO - so if you just loadCO on each popstate without taking care of that then everything works fine except that when history returns to that initial CO you get a loadCO that gets overrun by an App load (which loads fine in the end but causes jank and extra useless fetching). Fortunately timing is such that we're able to remove/replace that turbolinks state in the App code.

Tested on firefox and chrome.

### DIFF
--- a/app/javascript/vue/tasks/collection_objects/browse/App.vue
+++ b/app/javascript/vue/tasks/collection_objects/browse/App.vue
@@ -60,14 +60,26 @@ const collectingEvent = computed(
 const { collection_object_id: coId } = URLParamsToJSON(location.href)
 
 if (coId) {
+  // Call this for history.replaceState - otherwise turbolinks saves state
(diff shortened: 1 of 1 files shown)


kleintom:
> if you just loadCO on each popstate without taking care of that then everything works fine

(Actually not quite true, the `popstate` listener is on window, which doesn't get replaced during the reload, so each time you hit that first CO and do a reload you're adding a popstate listener in App, so now you have multiple popstate listeners, each of which does a loadCO etc. so that needs to be addressed as well if the turbolinks history state is kept in place.)

jlpereira:
@kleintom what are the steps to replicate this issue?

PR #3781, opened 2024-01-14
Lib refactor

mjy:


mjy:
Closing as out of date.  Still needs to be considered for modernizing Rails approach, but at this point deviations are such that we should just restart.

Issue #3163, opened 2022-10-28
DwC export - do not expose internal DwCOccurrence ID.


mjy:
Double check, and remove, the internal ID.  It's being exposed and changes every time we re-index.

kleintom:
Could you say a little more about the issue here? I thought you meant you didn't want the `id` column to show up in the download data when you do 'Download DwC' from Filter COs or 'Download Darwin Core Archive' from the DwC Dashboard.

PR #3789, opened 2024-01-19
#3163 Don't include the DwcOccurrence id column in DwC exports

kleintom:


mjy:
@kleintom This is by design.  The id is not persistent.  To recover records we need to reference the occurrenceID.

mjy:
@kleintom )(@*$

mjy:
@kleintom This may be introducing failures: https://github.com/SpeciesFileGroup/taxonworks/actions/runs/7587358876/job/20667528644?pr=3789#step:21:74

kleintom:
That worked out great actually, it
* caught a bug I had introduced: `DwcOccurrences.meta_fields` was opening the newly created csv file, yanking out the headers into an array, and then shifting off the first one, which used to be `id`. Now the shift is unnecessary. That fixed all but one of the failing tests (those were all checking meta_fields).
* The other fixed failing test was checking the header fields in an exported dwc csv file - the check now correctly checks that `id` is not included.
(I squashed the commits so there wouldn't be failing tests on any of them.)

Issue #3815, opened 2024-02-09
[Bug]: New Citation in Citation Annotator shows Topics component rather than Citations list


kleintom:
### Steps to reproduce the bug

```bash
1. Open any radial that shows Citations.
2. Delete all existing citations.
3. Now add a new citation.

Expected results: The new citation gets listed.
Actual results: The new citation doesn't appear, instead a Topics chooser/list shows with nothing listed.
```

### Environment

Development (docker)

### Sandbox Used

Also shows up in sandbox

### Version

08addf274e5abd25b297723f4190aa555e1569e5

jlpereira:
@mjy, if I remember correctly, this behavior was intentional. After create the citation, load the citation and display list of topics for that citation. This has changed?

kleintom:
I'm not sure my pr is what you want, but I'll add some more context as to what made it seem like a bug to me (a naive new user):
When the first citation is added, I see:
![image](https://github.com/SpeciesFileGroup/taxonworks/assets/632915/7c0b4b5e-0c88-4602-b2f7-476bd68897fe)
None of the icons to remove or edit the citation I added, nothing showing the name of the citation I added - if I now reload the page and go back to the citations radial, I see this:
![image](https://github.com/SpeciesFileGroup/taxonworks/assets/632915/b3404a6c-c466-4562-b44d-3c5e1284c0cb)
The topics section no longer appears and the list with my citation and the ability to edit or delete it appears, which is what I was (perhaps naively) expecting when the citation first got added.
Also, with the current behavior, if I start with no citations, add one, now the topics section is showing, now if I add a second citation I get the popup saying the citation was added but the display doesn't change (maybe it would if the second citation had different topics than the first?) and there's no visual indication that i have more than one citation.

And if I load the citation annotator with already existing citations, it starts off showing the list, but now if I add a new one the topics pane replaces the list and I can't see how to get back to the list except to reload the page. Maybe I'm missing something?

kleintom:
Sorry for the slow trickle as I figure some of this out, but now I notice that when I do choose the green edit icon from the list for a citation, it displays the Topics pane for that citation, but there's no indication (?) of which citation I'm editing, and no way to get out of the topics pane and back to the list. Maybe that's what this issue is actually about?
![image](https://github.com/SpeciesFileGroup/taxonworks/assets/632915/3a68d120-12fa-49b9-8582-a915b2d24c9c)

jlpereira:
[...]

PR #3816, opened 2024-02-09
#3815 Don't assign to the citation ref when a new citation is added i…

kleintom:
…n the citation annotator

The citation ref is for something else I think (related to resolving citations marked as original? Maybe a different name would help there?).


jlpereira:
@kleintom  I've talked with Matt and he told me the current behavior is expected in this context

Issue #3809, opened 2024-02-06
As an OtuPicker component user, I'd like a common interface for returns from get-item


kleintom:
### Feature or enhancement

When I use the OtuPicker component and set a `@get-item` on it, the keys of the return object depend on whether the otu that was picked in the picker was one that already existed or that the user created just now within the picker.

If the otu chosen in the picker already existed then the return object looks like this (pardon the garbage data):
![image](https://github.com/SpeciesFileGroup/taxonworks/assets/632915/ee3d8acf-9c2b-425e-b92f-191fb3931334)

If the otu chosen in the picker was one the user just created from within the picker, the return object looks like this:
![image](https://github.com/SpeciesFileGroup/taxonworks/assets/632915/45f41581-db4c-4ce4-a741-2c6cfce6680f)

The only overlapping key in the two cases is `id`, everything else would need to be checked on a case by case basis. Is it reasonable to expect the keys in the two cases to be the same, at least when their values exist?

### Location

OtuPicker

kleintom:
(There's at least one (minor) bug as a result of the non-overlap, here:
https://github.com/SpeciesFileGroup/taxonworks/blob/7356222b4de4ff0c6dc7f568ec21c999fab170ba/app/javascript/vue/tasks/loans/new/components/dateDetermination.vue#L32-L41
where the new otu case doesn't display when selected (`otuSelected` assigns from the wrong key on `$event`). I haven't checked everywhere else.)

PR #3827, opened 2024-02-14
In otuPicker don't try to assign taxon.id to a new otu when no taxon …

kleintom:
…is selected

Followup on #3809

### DIFF
--- a/app/javascript/vue/components/otu/otu_picker/otu_picker.vue
+++ b/app/javascript/vue/components/otu/otu_picker/otu_picker.vue
@@ -81,7 +81,7 @@
           @click="
             () =>
               createOtu({
-                taxonId: taxon.id,
+                taxonId: taxon ? taxon.id : undefined,
                 name: otuName
               })
           "


kleintom:
Maybe this should be a new issue (if so, feel free to ask me to file it), but I figure you're probably already aware - mentioning it just in case: I noticed that when you select or create an otu with no name using the otu picker, there's nothing displayed for the otu:
image
Would it make sense to display the taxon name or something else in that case?

jlpereira:
> Maybe this should be a new issue (if so, feel free to ask me to file it), but I figure you're probably already aware - mentioning it just in case: I noticed that when you select or create an otu with no name using the otu picker, there's nothing displayed for the otu: image Would it make sense to display the taxon name or something else in that case?

Context? What is the task? `object_label` or `object_tag` should contain the taxon name

kleintom:
Sorry, two separate issues, I think both apply anywhere the otu picker appears, for example in the Simple New Specimen task.
* The patch is for the case where you enter an unrecognized name in the otu picker and it offers to create a new otu for you, and you click Create without entering a taxon name:
image
In that case `taxon` is undefined and you get an exception.

* My other comment is just a UI question concerning the taxon name being blank - not that you're missing an object that holds the taxon name, just that that name is undefined, which means the picker displays blank, so there's no way to tell which otu you chose by looking at the UI, as in the image in my previous comment.  You can get a legitimate blank name from the picker in a couple ways:
  * if you enter the taxon of an otu that has a taxon but a blank name, for example an otu that was auto-created when you clicked on the no-associated-otu soft-validation fix on a new taxon name that didn't yet have an associated otu; or
  * if you enter the name of a taxon that matches no existing otus and the picker offers to create an otu for you via the 'Create and use OTU for taxon name ...' option:
 
image

jlpereira:
@kleintom I think I found the interface where you see this, Simple new specimen, right? The problem is not related with Otu picker. I pushed a fix for it, let me know if that works.

kleintom:
Yup, that fixed the blank name issue, thanks!

PR #3872, opened 2024-03-12
Catching some more errors

kleintom:
Commit 2 is the only one with real substance here (at least I hope) - without it the hard validation doesn't get set when creating or updating a taxon name.

p.s. I don't mind submitting these error-catching fixes, but feel free to weigh in on whether they're worth it. Personally I almost only work in docker development, so I have to deal with the red overlays, but actual users never see them, so the error-catching changes have no effect on them.


mjy:
All error catching is good.  Errors that happen "anonymously" are not OK :).

Issue #3877, opened 2024-03-14
[Bug]: Order issues with Leads


kleintom:
### Steps to reproduce the bug

```bash
Continuing the discussion from https://github.com/SpeciesFileGroup/taxonworks/commit/22a5a6c78216996f03647f6a61cb3c2fe4171071, I'll add more below. This is with the latest adjustments.
```

### Environment

Development (docker)

### Version

5cff93d3929c16f2934db34d7e0

kleintom:
It looks like there are still ordering issues with leads with the latest changes. STR for `insert_couplet`:

1) Create a new couplet (on left or right side doesn't matter) and save some distinguishing text for left and right. Note in the database that the left has `position=1` and the right has `position=2`.
2) Go up a couplet and insert a new couplet for the couplet from step 1. Now note in the database that `position` for both nodes created in step 1 is equal to 1, and `position` of both nodes created during the insert is equal to 2. Sometimes (but not always?) the newly inserted node that carries the old children appears on the wrong side of the ui (it should appear on whatever side the insert occurred on).
3) Since `position` no longer distinguishes which lead of those pairs comes first, updates can now cause the ui/display order of those pairs to change. So, e.g. edit the left lead that you created in step 1 and click Update - now the "left" lead is on the right and the "right" is on the left in the ui. (It looks like the ui order switches only when you edit/update the left lead, presumably because since `position` no longer determines order now last_updated does, or something like that?) The same switching happens if you edit the inserted pair in the same way.

In my testing so far, the `Delete both sides and reparent...` button seems to be working fine.

kleintom:
I'm all for getting this to work without using `numeric_order: true` so we don't have issues with db:create, but just to clarify, when we were using `has_closure_tree order: 'position', numeric_order: true, dont_order_roots: true` and has_closure_tree was computing `position`s for us, positions were only ever computed on child pairs (root nodes didn't have a `position` set as I recall, at any rate it definitely didn't increase or change with the number of trees in the table) and the child pairs all had `position` values of 0 or 1, so to my mind it acted as though `position` was scoped to `parent_id`, as it is now with `acts_as_list`.

kleintom:
I probably won't have time to look at this tomorrow, but if nobody gets to it earlier I should be able to try some things this weekend.

[...]

PR #3879, opened 2024-03-15
#3877 Update insert_couplet and destroy_couplet to deal with ordering…

kleintom:
… issues

This seems to work, and I've added more tests, including testing positions after `insert_couplet` and `destroy_couplet`.

Once again, as was the case when `has_closure_tree` was setting positions, many attempts that seemed perfectly reasonable (to me) failed when reparenting children with `acts_as_list`, meaning one or more child-pairs involved wound up with what seemed to me like incorrect positions given the operations performed.

For `insert_couplet` the closest I was able to get using 'natural' expected-to-work calls was to have the reparented children with correct positions but the inserted nodes with the same position (or vice versa I think). I could use builtin calls like `lead.move_higher` to fix things up in that situation without explicitly setting position values, but it would have still seemed strange since you wouldn't expect that call to be needed in the first place. So instead I went with explictly setting positions of both the inserted nodes and their new children. It's more calls than necessary (given the way things currently work), but it seems to work and I don't think it's too many extra database calls for a function that won't see much use (but would welcome other thoughts on that). It may also be more future proof, maybe.

That said, at some point I'd like to dig into why the expected calls are not working the way I'd expect and then maybe revisit this solution. If you'd rather figure things out that way now instead of using the direct position-setting calls, that's fine with me, just let me know.

New-rails-person lessons learned for me so far:
* there are more database `position`-setting calls going on than what appears in my `docker compose up` output; presumably this is `acts_as_list` making direct sql calls in addition to using builtin rails methods: that those direct sql calls don't show up in the output (I'd welcome corrections here). Lesson being you can't always just watch that output to figure out what's going on.
* you may need to reload your ActiveRecord objects after behind-the-scenes updates; otherwise things like updating position may fail if the actual position is different than what your outdated object thinks it is.
* (Don't expect the docs to always be accurate (I already knew that :).)

@mjy, since there are more tests now and it's been a number of years, I went ahead and disregarded your warning from mx this time:
https://github.com/mx3/mx/blob/e63b72db1e3cfd3f448430fc2bc07214288ab925/app/models/clave.rb#L110
I added grandkids to one of the destroy_couplet tests to try to test that warning a little more deeply and things are looking good, but if you remember any explicit things worth adding for that scenario, or you think it's wiser to do that destroy the mx way, please let me know!

### DIFF
--- a/app/models/lead.rb
+++ b/app/models/lead.rb
@@ -57,7 +57,7 @@ class Lead < ApplicationRecord
   has_many :redirecters, class_name: 'Lead', foreign_key: :redirect_id, inverse_of: :redirect, dependent: :restrict_with_error
 
   acts_as_list scope: [:parent_id, :project_id]
-  has_closure_tree order: 'position', dont_order_roots: true
+  has_closure_tree order: 'position'
 
   before_save :check_is_public
 
--- a/app/models/lead.rb
+++ b/app/models/lead.rb
@@ -99,19 +99,20 @@ class Lead < ApplicationRecord
     current_node_is_left_child = !parent_id || (parent.children[0].id == id)
 
     begin
+      texts = {
+        parented: children_present ?
+          'Child nodes are attached to this node' : 'Inserted node',
+        unparented: 'Inserted node'
+      }
+      left_text =
+        current_node_is_left_child ? texts[:parented] : texts[:unparented]
+      right_text =
+        current_node_is_left_child ? texts[:unparented] : texts[:parented]
+
       c, d = Lead.transaction do
-        texts = {
-          parented: children_present ?
-            'Child nodes are attached to this node' : 'Inserted node',
-          unparented: 'Inserted node'
-        }
-        left_text =
-          current_node_is_left_child ? texts[:parented] : texts[:unparented]
-        right_text =
-          current_node_is_left_child ? texts[:unparented] : texts[:parented]
         [
-          children.order(:position).create!(text: left_text),
-          children.order(:position).create!(text: right_text)
+          children.create!(text: left_text),
+          children.create!(text: right_text)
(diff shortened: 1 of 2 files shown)


kleintom:
Here's a version that works as expected without setting any explicit positions:
```ruby
c, d = Lead.transaction do
  if children_present
    a.update!(parent: nil)
    b.update!(parent: nil)
  end
  c = children.create!(text: left_text)
  d = children.create!(text: right_text)
  if children_present
    new_parent = current_node_is_left_child ? c : d
    a.update!(parent: new_parent)
    b.update!(parent: new_parent)
  end
  [c, d]
end
```
Still interested in looking into why more direct versions fail.

kleintom:
Better, maybe optimal:
```ruby
c, d = Lead.transaction do
  c = children.create!(text: left_text)
  d = children.create!(text: right_text)
  if children_present
    new_parent = current_node_is_left_child ? c : d
    a.update!(parent: new_parent)
    b.reload
    b.update!(parent: new_parent)
  end
  [c, d]
end
```
The `b.reload` is necessary here.

Before `c` and `d` are added to the same parent as `a` and `b`, `a` and `b` have positions 1 and 2 respectively. When `c` and `d` are added then `a, b, c, d` have positions 1, 2, 3, 4 respectively. When `a` gets reparented it's given position 1 (because it's the first child of `new_parent`) and `b, c, d` are shifted down to positions `1, 2, 3`. That all sounds good.
A theory: if `b` *doesn't* get reloaded it still thinks it has position 2, so when you do the update to reparent it as the second child of `new_parent`, the part of the code that decides if it needs to be given a new position thinks it's already set with position 2, when in reality in the database it actually has position 1, which is wrong.

PR #3882, opened 2024-03-18
Leads hub

kleintom:
Largely based on Source Hub. This is just a start, I'd welcome any thoughts!
image

Questions/TODOs:
* Do we want/need to page the list of keys?
* Include metadata section like source hub?
* Support sorting by column in the keys list (might depend on if we need to page that list)
* Make 'last updated' text nicer (maybe like /leads/list?)
* Make 'Is public' an editable checkbox in the keys list? (Is that something other tasks do?)
* Implement (some/all of these are different issues)
  * show entire key on one page, html
  * show entire key on one page, for printing
  * download markdown and/or latex versions of a key
    
Right now there's an `Options` button for each key that shows a modal with display/edit options for that key:
image
I'm not sure about the ux there, also right now it doesn't display anything not available in the radial navigator. I'm actually not sure if maybe all of the future options we're thinking of could just go on a radial like navigator or forms, and then the Options button could go away.

Right now it seems like the two 'display entire key' options could go on the radial navigator for a lead, but the download(s) should not, but I'm not sure about that. If everything can go on a radial then there's no need for the Options button, but I'm thinking if at least one of those wouldn't go on the navigator then we might as well list all of them as Options along with the download(s) option. Thoughts?


mjy:
Nice!

> Do we want/need to page the list of keys?

At the moment no. If we hit say 250+ keys we'll come back to it.

> Support sorting by column in the keys list (might depend on if we need to page that list)

Agreed. @jlpereira has default classes that make tables sortable.

> nclude metadata section like source hub?

We'll definitely want to know the root lead OTU, and possibly list a sample of OTUs treated in the key (maybe 5-6)

> Make 'last updated' text nicer (maybe like /leads/list?)

`time_ago_in_word` or somethign like that IIRC.

> Make 'Is public' an editable checkbox in the keys list? (Is that something other tasks do?)

This is a good idea.

> download markdown and/or latex versions of a key

Yes, download links from here will be good.  That can come with a new PR.  We will likely release this weekend, if we can get your basic task in (I suspect it's fine now) we can expand from there?

> I'm not sure about the ux there, also right now it doesn't display anything not available in the radial navigator. I'm actually not sure if maybe all of the future options we're thinking of could just go on a radial like navigator or forms, and then the Options button could go away.

You're right, these are navigator links at the moment.  We have a second class of radials, "object" radials".  Your ideas re future settings/options fit nicely into that concept should the need.  So, I don't believe we'll need the "Options" as you've laid them out.

mjy:
@kleintom I' inclined to merge this as is and have incremental changes added afterwards, how do you feel about that?

kleintom:
If you're happy I'm happy - feel free to remove the Options button and modal and make any other changes you see fit, or let me know if I can help with that.

kleintom:
Actually I can do that now and do the merge, unless you're already on it.

kleintom:
A couple more todos:
* add lead hub to lead navigator
* 'List of keys' on the edit key page should probably be key hub instead of /leads/list

mjy:
@kleintom as a note you can do `* [ ]` to turn bulleted lists into dynamic checkboxes, might help track the TODOs.  Merging this and will poke at the navigator TODO at least on development, leave the List for you.

kleintom:
Thanks @mjy; just a heads up that I have commits for the above TODOS, including the navigator (but no problem if you have it ready to add yourself for this release in a separate commit), I still need to finish part of the otus display, hopefully things should be ready later today, maybe this evening. You might want to change the ui/ux a little so depending on when the release is coming I wouldn't expect the new stuff to make it in this release.

Issue #3884, opened 2024-03-19
[Bug]: Last updated text for leads only refers to when the *root* node was last updated


kleintom:
### Steps to reproduce the bug

```bash
Users probably expect it to mean the last time anything in the key was last updated (not just the root node/metadata), at least that's what I would expect. I expect the same issue is happening with last_updated_by (that what we're displaying is really just "who was the *root* node last updated by?").

We're currently displaying the root's last_updated value at:
 /leads/list
 /leads/ (recents list)
 /tasks/leads/show (the list of keys)

And also the list of keys in the pr for a leads hub.
```

### Environment

Development (docker)

kleintom:
Could I get some feedback here? I'm considering two approaches:

1) Here's a query that returns:
* `updated_at` for the key (not just the root node)
* `updated_by` for the key, both id and user name
* couplets count
* otus count for the key
```ruby
roots = Lead.find_by_sql("
WITH leads_updated_at AS (
  SELECT
    roots.*,
    COUNT(DISTINCT otus_source.otu_id) AS otus_count,
    MAX(otus_source.updated_at) AS key_updated_at,
    (COUNT(otus_source.id) - 1) / 2 AS couplet_count
  FROM
    leads AS roots JOIN lead_hierarchies AS lh ON roots.id = lh.ancestor_id
      JOIN leads AS otus_source ON lh.descendant_id = otus_source.id
  WHERE roots.parent_id IS NULL
    AND roots.project_id = #{sessions_current_project_id}
  GROUP BY roots.id
)
SELECT leads_updated_at.*,
         updaters.updated_by_id AS key_updated_by_id,
         users.name AS updated_by
FROM
  leads AS updaters
    JOIN leads_updated_at
      ON leads_updated_at.key_updated_at = updaters.updated_at
    JOIN users
      ON users.id = updaters.updated_by_id
ORDER BY leads_updated_at.text
")
roots.map { |r|
  r.updated_by_id = r.key_updated_by_id
  r.updated_at = r.key_updated_at
  r
}
```
Those are all things we want when displaying a list of keys (otus count isn't necessary at this point but nice).

2) Get each of those data items one at a time, for each key at a time, using
```ruby
 def key_updated_at
    root.self_and_descendants.maximum(:updated_at)
  end

  def key_updated_by_id
    last_updated_node = Lead.find_by updated_at: key_updated_at
    last_updated_node.updated_by_id
  end
```
[...]

PR #3886, opened 2024-03-20
#3884 Update root's updated_by anytime any node in the tree is updated

kleintom:
This is so that when we display a list of keys by root nodes, the updated_last data in the root reflects the entire tree. (Of course this means now you won't always know when the metadata of a key was last updated.)

A couple notes on the location of the new callbacks:

* `self.root` is undefined in `after_destroy` and even in `before_destroy` if `before_destroy` is placed after the `has_closure_tree` declaration (suggesting whatever clears `root` on `self` happens in `has_closure_tree`'s `before_destroy`).

* `self.root` is undefined in `after_save` and all tests fail (on creating leads) unless `after_save` is placed after `has_closure_tree` (suggesting that `root` is set on a lead in `has_closure_tree`'s `after_save`).

### DIFF
--- a/app/models/lead.rb
+++ b/app/models/lead.rb
@@ -56,10 +56,15 @@ class Lead < ApplicationRecord
   belongs_to :redirect, class_name: 'Lead'
   has_many :redirecters, class_name: 'Lead', foreign_key: :redirect_id, inverse_of: :redirect, dependent: :restrict_with_error
 
+  # This needs to go before has_closure_tree.
+  before_destroy :set_root_updated_on_destroy, unless: -> { parent_id.nil? }
+
   acts_as_list scope: [:parent_id, :project_id]
   has_closure_tree order: 'position'
 
   before_save :check_is_public
+  # This needs to go after has_closure_tree.
+  after_save :set_root_updated, unless: -> { parent_id.nil? }
 
   validate :root_has_title
   validate :link_out_has_protocol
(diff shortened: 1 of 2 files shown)


kleintom:
> Sorry, I should have chimed in on this earlier.

No problem.

> there are no cascading updates like this intended

I'm fine with making the changes, but just to make sure I'm thinking of this right: that will make displaying a list of all keys with housekeeping info considerably more expensive, right? We'll have to query every lead node in the database every time. 

Paging would help, but not in the case of the `Recent updates` list at `/leads/`, where again we'd need to query every lead node in the database to figure out the n most recently edited leads.

Also there will be no record of an update if a couplet was deleted from a key.

Does that sound okay, or any other thoughts?

mjy:
> that will make displaying a list of all keys with housekeeping info considerably more expensive, right? We'll have to query every lead node in the database every time.

Yes and no.  ` p.root_taxon_name.self_and_descendants.maximum(:updated_at)` is 1/2 second for > 84k descendants, our keys will not be anywhere close to that size.

> Also there will be no record of an update if a couplet was deleted from a key.

Understood.  If we want to capture this we'll want to add a new attribute on Lead.  This brings in the concept of Versioning etc., which is good to think about, but I'm personally not super concerned about it at the moment, and if we do want to introduce this concept I'd like to do it with a generalized approach/concern that could be re-used.

PR #3887, opened 2024-03-21
Add volumes/ and /.ruby-lsp to .gitignore and remove some duplicate e…

kleintom:
…ntries

volumes/ is used by Docker so shouldn't be tracked; I'm not seeing any files in it even after `docker-compose up`, which I guess is why it hasn't been an issue in the past, but it still shows up in `git ls-files --others --directory --exclude-standard` which I'd like to use to chown directories created as root by `bin/rails generate ...` inside the docker image.

/.ruby-lsp is used by the ruby-lsp vscode plugin.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,8 @@ config/initializers/vendor/better_errors.rb
 /tmp
 /coverage
 /.ruby-gemset
+/.ruby-lsp
+/volumes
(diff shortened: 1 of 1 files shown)


mjy:
@LocoDelAssembly please merge if you concurr.

PR #3892, opened 2024-03-22
Remove acts as list from Lead

kleintom:
It turned out that `append_sibling` was what worked again here for reparenting (I'm personally starting to believe there's a bug in some of the other reparenting techniques regarding ordering).

### DIFF
--- a/app/models/lead.rb
+++ b/app/models/lead.rb
@@ -122,13 +122,18 @@ class Lead < ApplicationRecord
       c = Lead.create!(text: left_text, parent: self)
       d = Lead.create!(text: right_text, parent: self)
 
-      o.each do |i|
-        if p == :left || p == :root
-          i.update!(parent: c)
-        else
-          i.update!(parent: d)
-        end
+      new_parent = (p == :left || p == :root) ? c : d
+      last_sibling = nil
 
+      # !! The more obvious version using add_child is actually more error
+      # prone than using add_sibling.
+      o.each_with_index do |c, i|
+        if i == 0
+          new_parent.add_child c
+        else
+          last_sibling.append_sibling c
+        end
+        last_sibling = c
       end
     else
       c = Lead.create!(text: t1, parent: self)
--- a/app/models/lead.rb
+++ b/app/models/lead.rb
@@ -50,11 +50,6 @@ class Lead < ApplicationRecord
 
   has_closure_tree order: 'position', numeric_order: true, dont_order_roots: true , dependent: nil
 
-  # New records added at bottom (default)
-  # !! Note there is some oddness with update(parent: obj) vs. update(parent_id: object.id),
-  # if your function is not working as expected try the opposite.
-  acts_as_list scope: [:parent_id, :project_id]
-
   belongs_to :parent, class_name: 'Lead'
 
   belongs_to :otu, inverse_of: :leads
@@ -110,7 +105,7 @@ class Lead < ApplicationRecord
   end
 
   def insert_couplet
-    a,b,c,d = nil, nil, nil, nil
+    c,d = nil, nil
 
     p = node_position
 
@@ -166,20 +161,16 @@ class Lead < ApplicationRecord
 
         i = has_kids.children
 
+        # !! Use append_sibling, which is numeric_order-aware, to reparent
+        # the children of has_kids to self.
+        last_sibling = children[-1]
         i.each do |z|
-          # reload is required for pass, it seems we need to know the sibiling state
-          z.reload.update!(parent_id: has_kids.parent_id)
+          last_sibling.append_sibling z
+          last_sibling = z
         end
 
-        # `position` looks OK here
         has_kids.destroy!
-        # `position` looks OK here
         no_kids.destroy!
-        # !! `position` NOT OK here
-
-        # Give up as to why destroy fails to re-index. Force re-index of `position`
-        children.reload.reverse.map(&:save)  # reload is required
-
       else # Neither side has children
         Lead.transaction do
           a.destroy!
--- a/app/models/lead.rb
+++ b/app/models/lead.rb
@@ -160,8 +160,9 @@ class Lead < ApplicationRecord
 
         i = has_kids.children
 
-        # !! Use append_sibling, which is numeric_order-aware, to reparent
-        # the children of has_kids to self.
+        # Reparent the children of has_kids to self.
+        # !! The obvious solution using add_child is actually more error-prone
+        # than using add_sibling.
         last_sibling = children[-1]
         i.each do |z|
           last_sibling.append_sibling z
--- a/app/models/lead.rb
+++ b/app/models/lead.rb
@@ -112,8 +112,7 @@ class Lead < ApplicationRecord
     t1 = 'Inserted node'
     t2 = 'Child nodes are attached to this node'
 
-    # !! This reload is key to have specs pass
-    if children.reload.any?
+    if children.any?
       o = children.to_a
 
       left_text = (p == :left || p == :root) ? t2 : t1
--- a/spec/models/lead_spec.rb
+++ b/spec/models/lead_spec.rb
@@ -34,7 +34,7 @@ RSpec.describe Lead, type: :model do
 
   specify '#insert_couplet between leads inserts with positions' do
     FactoryBot.create(:valid_lead, parent: lead, text: 'bottom')
-    lead.insert_couplet
+    lead.reload.insert_couplet
     a = lead.children.reload.order(:position)
 
     expect(a.first.children.size).to eq(1)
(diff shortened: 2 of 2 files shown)


kleintom:
> I'm personally starting to believe there's a bug in some of the other reparenting techniques regarding ordering

Or that the `add_child` way of reparenting, for example, doesn't work the way I thought it did. `add_child` is documented as a way of reparenting nodes. I've been assuming that when you reparent a node it gets added to the end of the list of the current children. That might not be the case.

If I add the following to the `multiple leads` section of `lead_spec.rb` (thanks for adding the `draw` method, it's really great!):
```rb
specify "add_child doesn't work with ordering when reparenting" do
  lr.add_child r

  Lead.draw(lr.reload)
end
```
I get:
```
lr root 0 [.parent: l]
 lrl left 0 [.parent: lr]
 ll middle 1 [.parent: lr]
 lrr right 2 [.parent: lr]
```
as expected.

If I do
```rb
specify "add_child doesn't work with ordering when reparenting" do
  lr.add_child ll

  Lead.draw(lr.reload)
end
```
instead I get:
```
lr root 0 [.parent: l]
 lrl left 0 [.parent: lr]
 ll middle 1 [.parent: lr]
 lrr right 2 [.parent: lr]
```
where the new child was placed in the middle instead of the end.

My interpretation of that is that reparenting inserts a new child based on its current `position`, not necessarily to the end of the list: in the first case `r` had position 1 and it was added after the existing child with position 1, in the second case `ll` had position 0 and was added after the existing child with position 0 (and then the higher children had their positions bumped).

I guess `add_sibling`, for example, is the more explicit way of saying where the reparented child should be placed in the current `children`'s list.

kleintom:
You get the same results using `r.update! parent: lr` and `ll.update! parent: lr` instead. That definitely changes things for me :/

kleintom:
The failure of the following algorithm for `destroy_couplet`, versions of which I've tried mutliple times, now makes sense:

Given
```
          r
        /   \
       a     b
     /  \
    c    d
```
If I want to delete a,b and reparent c,d on r, I might try:
```
r.add_child c
r.add_child d
```
and expect the children of r to be `a,b,c,d` at that point, in that order.
Then I would simply
```
a.destroy!
b.destroy!
```
and be done.

What actually happens is
```
r.add_child c
```
r's children are now `a,c,b` with positions `0,1,2` because `c` had position 0 and so got inserted after `a`, which already had position 0 (and then `b`'s position got bumped to make room for `c`).
Now `d` has position 0 because it lost its only sibling, so when we do
```
r.add_child d
```
`d` gets added after the child of `r` with position 0 and now `r`'s child list is `a,d,c,b` with positions `0,1,2,3`.
Now the destroys of `a` and `b` work as expected and we're left with `r`'s children in the order `d,c` instead of `c,d`.

kleintom:
> My interpretation of that is that reparenting inserts a new child based on its current `position`, not necessarily to the end of the list

Yeah, that's looking to me like the way it works.
`after_save` calls `rebuild`:
https://github.com/ClosureTree/closure_tree/blob/af8572c460a5a5f69be19051d7ae702d0d37ebb1/lib/closure_tree/hierarchy_maintenance.rb#L39

`rebuild` first calls `_ct_reorder_prior_siblings_if_parent_changed` and then calls `_ct_reorder_siblings`.

Both of those functions call `_ct.reorder_with_parent_id`, the first with the previous value of the new child's parent id, the second with the current/new value of the new child's parent id.

Here's `reorder_with_parent_id` for postgresql: https://github.com/ClosureTree/closure_tree/blob/af8572c460a5a5f69be19051d7ae702d0d37ebb1/lib/closure_tree/numeric_order_support.rb#L34-L52
(As I'm sure will be obvious to many of you but was not at first to me) it assigns new positions sequentially based on the order of the current rows by position values.

So if current children have positions `0,1` for example and a new child just got added with position 0, the new child will get ordered as a second 0 (since it's newer), the current positions will get ordered as `0,0,1`, and the newly assigned positions will be `0,1,2` based on sequentially numbering those current positions (nice! but corrections welcome if I got anything wrong).

kleintom:
When I restart docker, db/schema.rb gets adjusted like this:
```
     t.index ["parent_id"], name: "index_leads_on_parent_id"
-    t.index ["position"], name: "index_leads_on_position"
     t.index ["project_id"], name: "index_leads_on_project_id"
     t.index ["redirect_id"], name: "index_leads_on_redirect_id"
-    t.index ["text"], name: "index_leads_on_text"
     t.index ["updated_by_id"], name: "index_leads_on_updated_by_id"
```
and now one of the insert_couplet tests fails.
I don't understand:
* why those two indices were removed (are they somehow automatic now?)
* why removing an index would cause a test to fail

But I'm investigating now.

kleintom:
If I stash the db/schema.rb changes then all the tests pass again. I had thought schema was a representation of the db, not a determiner?

kleintom:
Okay, apparently the *test* database *is* recreated from `schema.rb`. Still don't know why removing an index would cause a test to fail (or why those indices are being removed).

mjy:
@kleintom https://github.com/brendon/acts_as_list/issues?q=is%3Aissue+is%3Aopen+index

There are multiple issues there that reflect what you are seeing I think.

mjy:
@kleintom ignore previous, of course we are no longer using acts_as_list!

 You can always destroy schema.rb and re-migrate in case there were interactions between previous gem inclusions.

kleintom:
I pushed a commit to use `append_sibling` in `insert_couplet` instead of updating parents. I'll copy the commit message here - the cause of that failing test is speculative, but I do feel better about this version even if the particular reason for the error turns out to be wrong, as I explain in the message. It is slightly less intuitive and obvious to use append_sibling, so see what you think, if you prefer the old way (and you're not concerned about that test failing without those indices) I'm fine with that.

Here's the commit message:
To be honest this came about a bit by accident: I restarted docker and it came up with the following db/schema.rb changes:
```    
         t.index ["parent_id"], name: "index_leads_on_parent_id"
    -    t.index ["position"], name: "index_leads_on_position"
         t.index ["project_id"], name: "index_leads_on_project_id"
         t.index ["redirect_id"], name: "index_leads_on_redirect_id"
    -    t.index ["text"], name: "index_leads_on_text"
         t.index ["updated_by_id"], name: "index_leads_on_updated_by_id"
```
   
With those changes, the test 're-attached child lead positions are correct after #insert_couplet (right)' was failing consistently (but only if run with all other tests, if you ran it on its own it consistently passed).
    
I don't know why any of the above was occurring, but I do believe the changes here using add_sibling instead of add_child/updating-parent are more explicitly correct. In the previous version with reparenting, on the second child in the o.each block, we're adding a child with position 0 (because it's now the first child after one child was already reparented) to the new parent which already has a child with position 0. That could be interpreted as asking for trouble.
    
The new child being reparented should get ordered after the child that has already been reparented since it was updated most recently (with its new parent_id) - at least I've been assuming most-recently-changed is how that implicit order works (see reference below) - but maybe without an index on position that order is no longer consistent and the second child's 0-position winds up with it ordered before the original child's 0-position. (Why the same doesn't happen when doing an insert on a left node - where the same 0-positions setup occurs - I have no idead. Why it fails when all tests are run but not when just the failing test runs alone, I don't know.)
    
The ordering issue I'm referring to is the 'ORDER BY POSITION' here: https://github.com/ClosureTree/closure_tree/blob/af8572c460a5a5f69be19051d7ae702d0d37ebb1/lib/closure_tree/numeric_order_support.rb#L45
    When you've added a second child with position 0, as we were doing here, that code doesn't work right unless the most recently added 0-position-child gets ordered last by that order clause.
    
So in some sense the old code was maybe fine as long as those indices existed - which they definitely should - but I guess my feeling is that this version is still more explicitly correct nonetheless (since it doesn't make any assumptions about how the database is ordering two children with the same position [if that was actually the issue...]).

kleintom:
> You can always destroy schema.rb and re-migrate in case there were interactions between previous gem inclusions.

I tried that: deleted `schema.rb` and then did `bin/rails db:migrate` from the app shell. It recreated the same `schema.rb` with the missing lead indices.

kleintom:
I should have said with the new commit all tests pass with either version of the schema.

mjy:
@kleintom Thanks again for all the exploration and clarificaiton.  I'm going to pull this in for now so that we can anticipate a relelase if things align.

kleintom:
> deleted `schema.rb` and then did `bin/rails db:migrate` from the app shell. It recreated the same `schema.rb` with the missing lead indices.

If I do `bin/rails db:migrate:reset` instead then the missing indices reappear (I'm sure that's what you meant). Thanks for the lead!

kleintom:
The schema issue seems to have been an issue with an old database dump I've been reloading for a month or so; I think what might have happened is I dumped that db when the leads migration existed but didn't yet have indices set on those two fields (parent and text). Now I reload that db dump, which says it has already run the migration that (now) provides those indices, but in actuality it wasn't dumped with the version of that migration that provided them.

I'm still not sure how I've not run into that failing test before now in that case... Probably because I'm constantly reverting unwanted schema.rb changes. I'm willing to call this a happy accident :-) (Sorry for all the extra email chatter on this.)

PR #3895, opened 2024-03-26
Key hub improvements

kleintom:
Here's a screenshot of how it looks now:
image

Otus and citations for a key are listed below that key's row in the keys table. The root otu and the root citations of each key are loaded right away; there's a button to load any other otus from a given key - in the screenshot that button has been clicked on the second and third keys already.

Like I said earlier, thoughts on both the ui and the ux welcome, and I don't have any expectations that this will make it into the current release but if you want to cherry-pick anything here feel free.

### DIFF
--- a/app/controllers/leads_controller.rb
+++ b/app/controllers/leads_controller.rb
@@ -2,7 +2,7 @@ class LeadsController < ApplicationController
   include DataControllerConfiguration::ProjectDataControllerConfiguration
   before_action :set_lead, only: %i[
     edit create_for_edit update destroy show show_all show_all_print all_texts
-    destroy_couplet insert_couplet delete_couplet duplicate update_meta ]
+    destroy_couplet insert_couplet delete_couplet duplicate update_meta otus]
 
   # GET /leads
   # GET /leads.json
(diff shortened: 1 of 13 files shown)


mjy:
@kleintom I like the concepts.  Some observations: 

* [ ] Do we need radial annotators beside each citation?  The should be redundant with the in-line rightmost annotator if they are referencing leads.
* [ ] Let's not use underline here as we don't have this pattern elsewhere and in general underline suggests a link (we use blue text for links).
* [ ] Perhaps you could add a row with 2 spanning columns, one for the OTUs, and one for the citaitons, this might make the table feel a little more cohesive (nitpicking here)
* [ ] Hyperlink the text (title) of the key to goto the browse/show task

kleintom:
> * Do we need radial annotators beside each citation?

They're citation annotators, but I don't have a feel for how useful that actually is, I'll remove them.

kleintom:
> * (nitpicking here)

Nitpicks welcome, here's a new screenshot, see what you think:
image

mjy:
@kleintom merging this as is to sneak it in, you can PR against development again please.

PR #3899, opened 2024-03-29
Key hub updates and updates to updated_by/at handling for keys

kleintom:
There are two issues resolved here:
* 3884, show updated-by and -at for the whole key, not just the root node
* The last two TODOs from @mjy on the keys hub: move Is Public to the last column and display number of otus in the key. Given work done for the updated_ issues I was able to add the otus count on load instead of only after actually loading/displaying all of the otus (which is why the two issues are being submitted together).

It took me some time to figure out some technical issues concerning the new query for root leads - see the commit message on the next to last commit (feel free to skip if it's all old news).

In short, the issue is that the query is a `.joins`. A simplified example is `q = T.joins('JOIN U on T.id = U.id).select('U.*')`. This is a contrived example, but two of the issues with this query are that, even though you selected the `U` side:
* If you do `q.order(:text)` where both `T` and `U` have a text column, you get ordered by `T.text`, not `U.text`
* If you do `q.pluck(:text)` you get texts from the `T` side of the relation, not the `U` side.
There are ways to call both order and pluck to get what you want, but it requires knowledge of how the query was created and (in my case) the alias used for the second table `U`. if you return this query from a function, users will have no idea of the issue. That's the current state of these commits. I'm not sure if this is a common kind of issue, or if there's a standard way to fix it, or if it's just the way it is. Thanks for taking a look!

### DIFF
--- a/app/controllers/leads_controller.rb
+++ b/app/controllers/leads_controller.rb
@@ -10,17 +10,11 @@ class LeadsController < ApplicationController
     respond_to do |format|
       format.html {
         one_week_ago = Time.now.utc.to_date - 7
-        recents = []
-        roots = Lead.roots_with_data(sessions_current_project_id)
-        roots.each do |k|
-          if k.key_updated_at > one_week_ago
-            recents.push k
-          end
-        end
-
-        recents.sort! { |a, b| b[:key_updated_at] <=> a[:key_updated_at] }
-
-        @recent_objects = recents.take(10)
+        @recent_objects = Lead
+          .roots_with_data(sessions_current_project_id)
+          .where('key_updated_at > ?', one_week_ago)
+          .reorder(key_updated_at: :desc)
+          .limit(10)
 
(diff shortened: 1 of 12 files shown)


app/controllers/leads_controller.rb
```diff
@@ -9,22 +9,34 @@ class LeadsController < ApplicationController
   def index
     respond_to do |format|
       format.html {
-        @recent_objects = Lead.recent_from_project_id(sessions_current_project_id).where('parent_id is null').order(updated_at: :desc).limit(10)
+        one_week_ago = Time.now.utc.to_date - 7
```
mjy, line 12: @kleintom you have a native relation returned from `roots_with_data`.  You should be able to chain addition .where, limit, order conditions directly on to that rather than dropping to Ruby?

kleintom: Yup, thanks.

PR #3945, opened 2024-05-10
Adds an extended BA endpoint (closes #3944)

### LINKED ISSUE #3944: As an R user, I need an extended biological associations endpoint
### Feature or enhancement

As an R user, I need an extended biological associations endpoint that extends the simple endpoint with identifiers on all objects and taxon_name_id's.

https://github.com/SpeciesFileGroup/Rtaxonworks/issues/4

### Location

/api/v1/biological_associations/extended

### Screenshot, napkin sketch of interface, or conceptual description

_No response_

### Your role

_No response_

gdower:


mjy:
@gdower Thanks.  I think we should add a couple of fields and clarify a couple things.  `object_id` and `subject_id` can be for different types, right?  so we need `object_type` and `subject_type` perhaps?

gdower:
@mjy, good point. I added those fields.

PR #3958, opened 2024-05-29
A couple commits around import attributes

kleintom:
I'm not sure if the original AddInternalPredicate.vue behavior was intentional? If so I'll resubmit.

I can add some tests for the import attribute filtering, just let me know (https://github.com/SpeciesFileGroup/taxonworks/blob/development/spec/lib/queries/concerns/data_attributes_spec.rb).


kleintom:
Contrary to my commit message, when you try to create a source from an unknown bibtex type, the error message letting you know does appear, it's just that the exception is uncaught.

mjy:
Thanks @kleintom this looks like some good catches.

PR #3960, opened 2024-05-30
Matrix view ordering update and nexus export changes

kleintom:
The main thing here is a couple fixes/changes to make nexus export of OMs more usable/nexus-standardized:
* fix an issue where single word taxon names were being exported as ''
* deal with the discrepancy between the way TW handles gap states (with a gap character state specific to each descriptor that uses a gap state) and the way nexus handles gap states (with a global symbol, which these commits assume/assign to '-', and which does not contribute to sequential numbering of character state labels for a given descriptor). The way gap states are currently exported confuses Mesquite (see the commit message). [Note that the export also/still assumes that the state labels given to non-gap states in TW are numbered sequentially starting at 0.]


app/views/observation_matrices/export/nexus/_nexus.html.erb
```diff
@@ -20,12 +20,13 @@ Begin characters;
 Title qualitative_characters;
 Dimensions Nchar=<%= descriptors.size -%>;
 Format
-<%= as_file ? '  ' : nb -%>datatype = STANDARD  
-<%= as_file ? '  ' : nb -%>missing = ?  
-<%= as_file ? '  ' : nb -%>symbols = "0 1 2 3 4 5 6 7 8 9 A B C D E F G H I J K L M N P Q R S U V W X Y Z !"; 
+<%= as_file ? '  ' : nb -%>datatype = STANDARD
+<%= as_file ? '  ' : nb -%>gap = -
+<%= as_file ? '  ' : nb -%>missing = ?
+<%= as_file ? '  ' : nb -%>symbols = "0 1 2 3 4 5 6 7 8 9 A B C D E F G H I J K L M N P Q R S U V W X Y Z !";
```
mjy, line 26: Side note: we could improve the `symbols` line by referencing the observed states in the data, this would be a fairly big improvement.

app/views/observation_matrices/export/nexus/_nexus.html.erb
```diff
@@ -20,12 +20,13 @@ Begin characters;
...
 Charstatelabels
 <% descriptors.each_with_index do |d, i| -%>
-  <%= as_file ? '  ' : nb -%><%= "#{i+1} #{descriptor_matrix_label(d).html_safe} /".html_safe %> <%= d.character_states.order(:position).collect{|cs| "'#{cs.name}'".html_safe}.join(' ').html_safe if d.qualitative? -%>,
+  <%= as_file ? '  ' : nb -%><%= "#{i+1} #{descriptor_matrix_label(d).html_safe} /".html_safe %> <%= d.character_states.order(:position).filter_map{|cs| "'#{cs.name}'".html_safe if cs.name.downcase != 'gap'}.join(' ').html_safe if d.qualitative? -%>,
```
mjy, line 29: I feel we could add a method to CharacterState `is_gap?` that predicts whether the character is a gap?

PR #3962, opened 2024-05-31
Add gap detection for Qualitative OM stuff and compute Symbols list for nexus export

kleintom:
Should a gap state be 'required' to have label '-'?


mjy:
> Should a gap state be 'required' to have label '-'?

Techically no.  Practically I think we assume this utill we hear otherwise via a real-life use case.

PR #3965, opened 2024-06-03
Fix 500 error on /tasks/administrator/project_classification

kleintom:
I don't know what's going on with that 'chartjs' parameter, but this seems to fix it...

The same `javascript_include_tag` at /administration/user_activity seems to not cause any issue; I removed it here since it doesn't add anything to the include that shows up in the source as far as I can tell. Again, not sure if that's the right thing to do or not.


kleintom:
It seems any logged in user can view /tasks/administrator/project_classification

PR #3987, opened 2024-07-09
Fix to put cursor in input on DestroyConfirmation modal

kleintom:
Was throwing a this.$refs.inputtext nil error before.


jlpereira:
Good catch!

PR #4006, opened 2024-08-14
Rails 7.2 initial setup (unofficial postgis adapter)

LocoDelAssembly:
Do not merge right away. We likely merge no earlier than official support from `activerecord-postgis-adapter`.
~~Also watch `activerecord-nulldb-adapter`, using unofficial because original one limits activerecord version to `<7.2`~~


LocoDelAssembly:
> Do not merge right away. We likely merge no earlier than official support from `activerecord-postgis-adapter`. Also watch `activerecord-nulldb-adapter`, using unofficial because original one limits activerecord version to `<7.2`

postgis-adapter now back to official. Not sure about nulldb, I don't see it in Gemfile/Gemfile.lock.

PR #4120, opened 2024-11-19
Instructions for importing a data download in a Docker Compose dev env

neolefty:
Documentation only.

* Are these instructions correct?
    * They are tested, but is there a better way?
* Is this a good (or only) place to put the documentation?

### DIFF
--- a/app/views/tasks/projects/data/index.html.erb
+++ b/app/views/tasks/projects/data/index.html.erb
@@ -8,6 +8,14 @@
     <h3> Export SQL </h3>
     <p><em>Generate a downloadable copy of the database (PostgreSQL dump) with all data referenced in this project. Includes Community data like Sources, People, and Repositories.</em></p>
     <p> Restorable with <b>psql -U :username -d :database -f dump.sql</b>. Requires database to be created without tables (<b>rails db:create</b>) </p>
+    <p>To import into a <code>docker compose</code> development environment:</p>
+    <ol>
+      <li>Download and unzip the file. In this example, assume that creates a local file <code>unzip/dump.sql</code></li>
+      <li>Mount the unzipped file in the <code>db</code> container. For example add a line to <code>docker-compose.yml</code> inside <code>services.app.db.volumes</code> like this: <code>- ./unzip:/mnt/unzip</code> </li>
+      <li>Drop existing DB: <code>docker compose exec app bundle exec rails db:drop</code></li>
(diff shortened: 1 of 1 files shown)


LocoDelAssembly:
`docker compose exec db psql -U postgres -d taxonworks_development -f /mnt/unzip/dump.sql`

I believe you can just send through stdin and save yourself from adding another volume:

`docker compose exec db psql -U postgres -d taxonworks_development -f - < local/path/dump.sql`

[https://github.com/SpeciesFileGroup/install_taxonworks/tree/master/development/docker#restore-a-database-dump](Here) there is an example with `pg_restore` (doesn't work with `sql` files)

neolefty:
> `docker compose exec db psql -U postgres -d taxonworks_development -f /mnt/unzip/dump.sql`
> 
> I believe you can just send through stdin and save yourself from adding another volume:
> 
> `docker compose exec db psql -U postgres -d taxonworks_development -f - < local/path/dump.sql`
> 
> [https://github.com/SpeciesFileGroup/install_taxonworks/tree/master/development/docker#restore-a-database-dump](Here) there is an example with `pg_restore` (doesn't work with `sql` files)

Thanks for the suggestion — that would be better.

However, I tried that but didn't get it to work:

```shell
$ echo "select count(id) from users;" > foo.sql
$ docker compose exec db psql -U postgres -d taxonworks_development -f - < foo.sql
WARN[0000] /Users/wbbaker/source/taxonworks/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion 
the input device is not a TTY
```

LocoDelAssembly:
As discussed in https://github.com/SpeciesFileGroup/taxonworks_doc/pull/94#issuecomment-2512331198 perhaps such warning should also appear here. Text might become too big already and perhaps we should consider linking to docs for ways on restoring dumps instead of keep expanding.

Other than that, the instructions in this PR are correct. Perhaps we can merge, then refactor to link to docs with more use cases and extended explanation.

mjy:
> Perhaps we can merge, then refactor to link to docs with more use cases and extended explanation.

Agreed.  We'll merge this *after* 0.46.0 release. 

We need some cannonical way to store linkages between app and documentation endpoints, maybe a config file that lets us introduce help tags pointing to help keywords, perhaps something likeK

```yml
# config/initializers/help_index.yml
---
help_keyword: 
  url: 
keyword_keyword2: 
  url:
```

`<%= help_link(help_keyword) %>`

or 

`/help/:keyword` to get the JSON response to inject the link.

PR #4125, opened 2024-11-21
Modernize docker compose usage & language

neolefty:
I think the old versions of `docker-compose` have percolated out of dev environments by now, since v2 was released in 2020. See https://docs.docker.com/compose/releases/migrate

* `docker-compose` –> `docker compose`
* `docker-compose.yml` –> `compose.yml`
* Remove deprecated "version" tag from `compose.yml` (it is causing warnings in the dev env)

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -1,7 +1,6 @@
 #
-# docker-compose targets development only
+# docker compose targets development only
 #
-version: '2'
 services:
   app:
     build:
@@ -30,6 +29,7 @@ services:
       - '0.0.0.0:15432:5432'
     volumes:
       - pg12:/var/lib/postgresql/data
+      - ./unzip:/mnt/unzip
   webpack:
     build:
       context: .
--- a/.github/docker-tests/compose.yml
+++ b/.github/docker-tests/compose.yml
@@ -1,4 +1,3 @@
-version: "2.1"
 services:
   taxonworks:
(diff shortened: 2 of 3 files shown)


mjy:
@LocoDelAssembly You mentioned a concern with the change in chat I think?

LocoDelAssembly:
Missed this one. File rename and removing version seems compatible with the docker version that comes with Ubuntu 22.04 LTS (still supported). However I'd remove the unzip mount point as it is not necessary.

The version number was 2 because I think were using something that was removed in later versions, something related to health checks or detecting when the container is ready, but not currently needed since startup script in app container waits for db to come up before letting TW run.

mjy:
@neolefty Did you see? V (and comments)?

> However I'd remove the unzip mount point as it is not necessary.

compose.yml
```diff
@@ -30,6 +29,7 @@ services:
       - '0.0.0.0:15432:5432'
     volumes:
       - pg12:/var/lib/postgresql/data
+      - ./unzip:/mnt/unzip
```
neolefty:
```suggestion
```

neolefty:
> @neolefty Did you see? V (and comments)?
> 
> > However I'd remove the unzip mount point as it is not necessary.

Done! Sorry I forgot about this until now.

PR #4144, opened 2024-12-09
Multifurcating key and usability additions

kleintom:
The main change is to make keys multifurcating instead of just dichotomous (with UI now using "Multifurcating" instead of "Dichotomous").

To support that, added two new actions:
* add_lead, which adds a new lead to the children of a given lead
* swap_lead, which changes the place of two adjacent sibling leads

Other changes:
* added the ability to insert a new couplet on root (per Davide's request)
* added the ability to insert a key (per Davide's request): currently we dupe the key to be inserted, and then append the duped root to the children of the lead inserted on
* added a delete button on leads in the UI, which will delete the entire subtree at that lead (is the warning big enough?)
* I updated the UI in 'Show key' to make it more clear where individual leads start and stop (for the case when there are enough leads to require more than one row of leads); changes or suggestions welcome!

When I started this work I thought couplet referred only to the case where there were exactly two options, so I changed a lot of wording to "option set" instead of "couplet" - but then I saw https://en.wikipedia.org/wiki/Single-access_key which implies the term 'couplet' can be applied to any number of leads, so I changed everything back to 'couplet'. I think I prefer that for clarity/common use (even though internally I still have a strong feeling that couplet should match with 2). Preference?


mjy:
>  "couplet"

Thanks for the link.  I've struggled with nomenclature of keys too.  I'm wondering if that article thought of > 2, couplet/ing really means two in most cases. That said, I think for now it's fine to leave it as couplet.  Perhaps we don't really need to name the set-entity at all in the UI...

app/controllers/leads_controller.rb
```diff
@@ -33,14 +34,14 @@ def list
       roots_with_data(sessions_current_project_id).page(params[:page])
   end
 
-  # GET /leads/1/all_texts.json
-  def all_texts
+  # GET /leads/1/redirect_option_texts.json
+  def redirect_option_texts
```
mjy, line 38: This lodgic should move to the model I think.

app/controllers/leads_controller.rb
```diff
@@ -189,37 +187,61 @@ def destroy_couplet
...
         format.json {
           render json: @lead.errors, status: :unprocessable_entity
         }
       end
     end
   end
 
-  # POST /leads/1/duplicate
+  # POST /leads/1/add_lead.json
+  def add_lead
```
mjy, line 207: This seems like a clone of `def create`?

app/controllers/leads_controller.rb
```diff
@@ -189,37 +187,61 @@ def destroy_couplet
...
+    parent = Lead.find(params[:id])
+    @lead = parent.children.create!
+  end
+
+  # POST /leads/1/duplicate.html
   def duplicate
```
mjy, line 213: Can you clarify if this is intended to duplicate children?  If not then it feels like it's possibly doable with client side sending `def create` a filtered set of parameters?

app/controllers/leads_controller.rb
```diff
@@ -189,37 +187,61 @@ def destroy_couplet
...
+      @lead.errors.add(:insert_key, "failed: '#{rv}'")
+      render json: @lead.errors, status: :unprocessable_entity
+      return
+    end
+
+    head :no_content
+  end
+
+  # PATCH /leads/1/update_meta.json
   def update_meta
```
mjy, line 245: This is redundant with `def update` and a JSON formatted call I think.

app/controllers/leads_controller.rb
```diff
@@ -230,15 +252,25 @@ def update_meta
...
+    end
+
+    @leads = swap_result[:leads]
+    @positions = swap_result[:positions]
+    @futures = @leads.map(&:future)
+  end
+
   # GET /leads/1/otus.json
   def otus
     leads_list = @lead.self_and_descendants.where.not(otu_id: nil).includes(:otu)
```
mjy, line 271: Can you invert this to somethign like `@otus = Otu.joins(:leads).<in key scope>`? We'll need a dynamic scope like that for other uses I suspect.

app/javascript/vue/tasks/leads/hub/app.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <h1>Dichotomous Keys Hub</h1>
+  <h1>Multifurcating Keys Hub</h1>
```
mjy, line 2: I think we're going to use "Keys hub" and "Interactive keys hub" as the differentiation.

app/javascript/vue/tasks/leads/hub/components/KeysList.vue
```diff
@@ -109,7 +109,7 @@
           :href="RouteNames.NewLead"
           data-turbolinks="false"
         >
-          New dichotomous key
+          New multifurcating key
```
mjy, line 112: See comment above.

app/javascript/vue/tasks/leads/new_lead/app.vue
```diff
@@ -6,7 +6,7 @@
       :href="RouteNames.LeadsHub"
       data-turbolinks="false"
     >
-      Dichotomous Keys Hub
+      Multifurcating Keys Hub
```
mjy, line 9: "Keys Hub"

app/models/lead.rb
```diff
@@ -231,9 +242,34 @@ def all_children_standard_key(node = self, result = [], depth = 0) # couplets be
     result
   end
 
+  # Return value always lists self first.
```
mjy, line 245: There is a position-based paradigm/pattern for updating positions that might be cleaner.  In essence you POST to somethign like a `reorder` action the ids in order, in an array. THen you loop that array and simply update the position (without validation callbacks etc).  However, I'm not completely sure that meets the need in this method.

config/interface/hub/user_tasks.yml
```diff
@@ -868,19 +868,19 @@ debug_geographic_item_task:
   description: 'Report metadat about a geographic item.'
 new_lead_task:
   hub: true
-  name: 'New dichotomous key'
+  name: 'New multifurcating key'
```
mjy, line 871: See above, I think we can just do "New key", and "New interactive key", and folks will get familiar with that pattern.

mjy (commented):
@kleintom Thanks.  Some smaller points to consider before I spend some time on the UI.  Thanks again for all your work, this is a lot.

In general keep trying to think about using RESTful requests to update the data whenever possible.  I know there is a lot of swapping/cloning type things going on here, which don't fit too nicely into that pattern, but if you can break down operations to a single Lead then use create/update/new.  I'm not positive how many opportunities for this refactoring there are, maybe 1 or two though.

app/controllers/leads_controller.rb
kleintom: Yes, this duplicates all children of the lead it acts on.

mjy:
@kleintom I'm going to merge to development as is, thanks very much for all this work.   Should be in sandboxes shortly.

PR #4156, opened 2024-12-17
Key print fixups

kleintom:


mjy:
Wonderful, thanks for the refactor @kleintom!

PR #4169, opened 2025-01-16
Scaffold Sound and Conveyance

mjy:
Sound branch.  For test purposes only.


kleintom:
Very neat, I'm really looking forward to seeing what unexpected sounds get added! I'm sure you're probably aware of some of these already, but for what it's worth:

### General
* The combination of:
  * Sound autocomplete doesn't search on filename, and
  * when you create a sound from an annotator, you can't set a name

means that currently you can't autocomplete on sounds created from annotators (without first setting the name by some other means), which was unexpected.
If `name` remains unrequired for sounds, does it perhaps make sense to be able to filter on filename in addition to name?

### 500s
* Select edit in the header bar from conveyances/1 or from a conveyances radial navigator
* conveyances/list (select 'list' from /conveyances)
* select 'related' from a sound navigator
* from filter sounds, choose 'confidence' and select a confidence from either of the batch annotators (works fine on individual sound annotators)
* add a sound row to a matrix (I presume that's what triggers this), then choose 'interactive key' from the matrix navigator

### Conveyances in radial annotators
* Is it intended that more than one conveyance can use the same sound file? Currently I don't see how to do that: you can only select a new file when creating a conveyance from annotators.
* Deleting conveyances from the radial updates the ui but doesn't actually do the delete (when you reload they're back)
* In edit/move_to there's a 'taxon name' option (which fails) - should that be field occurrence instead?

### Collecting events
In gitter you mentioned annotating collecting events with sounds, but I'm not seeing how, there's no conveyances option in annotators (CE also doesn't show up as a 'move to' option).

### Filter sounds
* add filter for blank name?

### Browse sound
* Should the `sound_id` query param be set on selection from search box? Possibly related:
  * If I search for a sound from the Browse Sounds search box, its conveyances don't get loaded - I have to go to a different page and select 'Browse sounds' from the navigator radial for that sound to see the conveyances in browse
  * If I search for one sound and load it, then search for a different sound and load it, the sound doesn't update.
* add a sound navigator radial in addition to the annotator?
* would it be helpful for conveyance objects displayed in the Conveyances panel to link to their object or have radials?

### Matrices (cool that adding this was so easy!)
* Add a sound row: the green edit circle is a broken link

mjy:
Amazing, thanks @kleintom.  I was wondering who was hitting things thoroughly last nite, seeing email raise messages from different places.

kleintom:
In matrix edit, the pinboard tab for new rows doesn't show pinned sounds:
image

kleintom:
This is looking great!
* I'm getting a vue exception when I send sounds to unify from Filter Sounds
* I'm getting duplicate result rows from the 'Conveyance object type' facet in Filter Sounds (with my toy data where I'm putting the same sound on multiple conveyances)
* (still getting the 500 on trying to use an Interactive Key with a sound row)

jlpereira:
@kleintom Thank you so much for taking the time to do all this testing.

@proceps could you check interactive keys lib? I think it's expecting an otu id for sounds, not sure how you want to handle this

kleintom:
Not that people should really be doing it this way, but FYI I just tried New from a Conveyance navigator radial and got a 500.

mjy:
500s are never OK.  The only time we don't "fix" them in code is when they are because of (now) invalid data.  

* We should redirect_to the new_sound_task from those endpoints.
* There is also a 'NO_NEW_LINK' or somethign like that constant in one of the helpers, navigation maybe?

PR #4184, opened 2025-02-05
Cached map issues

kleintom:


app/models/otu/maps.rb
```diff
@@ -12,10 +12,9 @@ module Otu::Maps
   #       Should, in theory, speed-up higher level maps
   #
   def create_cached_map(cached_map_type = 'CachedMapItem::WebLevel1', force = false)
-
     if force
       cached_map = cached_maps.where(cached_map_type:)
-      cached_map&.destroy
+      cached_map&.destroy_all
```
mjy, line 17: I think this might be a problem with the relationship, Otu map relationship should be `has_one` ... maybe

mjy:
Thanks again @kleintom

PR #4185, opened 2025-02-05
Make FieldOccurrences destroyable

kleintom:
Currently destroy reports success but actually silently fails, in TaxonDetermination#prevent_if_required.

This works, but I confess this isn't the way I would have done it, and I don't understand why "my way" doesn't work. [Additional unaddressed issue in this submitted patch: this doesn't surface errors from destruction of associated objects, only errors directly on FO.]

I would have put `include Shared::TaxonDeterminationRequired` *after* `include Shared::BiologicalExtensions`.

There are two issues between those includes:
* each defines #requires_taxon_determination?, which needs to be true for FO - BE makes it false, TDR makes it true. Including BE last as in this patch requires that the function is defined to be true in the FO model itself (as was already the case), if it were included first then FO wouldn't need to repeat that function (since TDR would give the right version). That all makes sense to me.
* the order of the includes seems to have an effect on the order in which:
  * `before_destroy` in TDR, and
  * `before_destroy` in TaxonDetermination (a has_many/has_one relation in FO) get run. That I don't understand, I would have thought the before_destroy in FO would always get run before the before_destroy in its associated taxon_determinations, but here it seems to depend on whether the association or the TDR before_destroy gets included first (??). Shrug (would need to read the rails source to figure it out I guess, but please let me know if anybody knows what's going on).


kleintom:
@mjy, I assume this is what you intended `Shared::TaxonDeterminationRequired` for, but if not feel free to take this one (or let it wait).

kleintom:
On second thought, maybe hold off on this one, I think there's more work to do here.

mjy:
Agreed, this is tricky, stuff.  Haven't looked hard at what you are suggesting here yet.

> Field occurrence was not destroyed, Cannot delete record because dependent biological associations exist

This is desired.

kleintom:
Aside from not understanding
> * the order of the includes seems to have an effect on the order in which [two different before_destroys run]

this is looking good to me.

mjy:
@kleintom Please sync with dev when you get a chance.

mjy:
@kleintom I think the include order matters because we must have the before_destroy that flags that it's *OK* to destroy in place before the other destroy related hooks.

Thanks for finding and fixing this.

PR #4194, opened 2025-02-10
Taxon Names With/Without-gender filter

kleintom:


lib/queries/taxon_name/filter.rb
```diff
@@ -562,6 +573,38 @@ def leaves_facet
...
+        return nil if latinized.nil?
+
+        tnc = ::TaxonNameClassification.arel_table
+        if latinized == true
+          # Note the query here does not restrict to genus/species groups - a
+          # genus whose rank is changed to subfamily can retain its gender,
+          # e.g., and we want to include those here.
+          ::TaxonName.where(
+            ::TaxonNameClassification.where(
+              tnc[:taxon_name_id].eq(::TaxonName.arel_table[:id]).and(
```
mjy, line 587: @kleintom When we originally coded out this aproach I leaned heavily on arel, I've since moved somewhat away from that, for better or worse.  This is fine (because of specs). Do check to see if there are some generic methods in superclasses (there used to be `def table` which gives you the `TaxonName.arel_table`, for example).

The `.with()` paradigm makes it significantly easier to join individual queries as well.

There are certainly other considerations I've realized in the Filter framework that we could discuss in person perhaps, things to think about in terms of general optimizations.

lib/queries/taxon_name/filter.rb
```diff
@@ -562,6 +573,38 @@ def leaves_facet
...
+        else
+          ::TaxonName
+            .where(
+              "rank_class ILIKE '%speciesgroup%' OR \
+               rank_class ILIKE '%genusgroup%'"
+            )
+            .where.not(
+              ::TaxonNameClassification.where(
+                tnc[:taxon_name_id].eq(::TaxonName.arel_table[:id]).and(
+                  tnc[:type].matches('%latinized%')
```
mjy, line 601: If possible see if we can reference a constant that has the pertinent `type`s (`config/initializers/constants...`).  This lets us manage names == meaning nastiness in one place in the long run.  I.e. I *think* we should try to unify the general code pattern to avoid these matches.  We have this pattern in many places, it should(?) be improved(?) elsewhere too.

I'd not be too worried about such (again, tests are great) but since you're coming in a little more fresh to the nomenclatural side of things it might be an opportunity to get us all to be a little more consistent.

kleintom: > If possible see if we can reference a constant that has the pertinent `type`s (`config/initializers/constants...`). This lets us manage names == meaning nastiness in one place in the long run. I.e. I _think_ we should try to unify the general code pattern to avoid these matches.

That feels better to me, thanks for these pointers. To be honest I'm still frequently at the 'what's the right syntax for this query in rails (can I write this in rails)'/'I'm not sure how this rails/arel code works without to_sql'ing it' stage, so any thoughts on these things are always welcome.

kleintom:
gak, I didn't do the same for speciesgroup/genusgroup, I'll submit again :-/

mjy:
Thanks.

PR #4195, opened 2025-02-10
Add TNRelationship subject/object filters to TN filter, other TN filter fixes

kleintom:


kleintom:
Marking as a draft just so I can add some notes later today.

mjy:
@kleintom Again, thanks much.

PR #4196, opened 2025-02-10
Edit taxon name fixups

kleintom:
STR commit 1: create a synonym where the old name has children, then go to manage synonymy in Edit Taxon Name and click green edit button --> bad link

STR commit 2: go to a taxon with a relationship, goto relationships section, click edit green --> in development mode in the console see [Vue warn]: Invalid prop: type check failed for prop "objectLists". Expected Array, got Object  
  at <CommonList key=1 object-lists=

### DIFF
--- a/app/javascript/vue/tasks/nomenclature/new_taxon_name/components/manageSynonym.vue
+++ b/app/javascript/vue/tasks/nomenclature/new_taxon_name/components/manageSynonym.vue
@@ -136,6 +136,7 @@
 
 <script>
 import { GetterNames } from '../store/getters/getters'
+import { RouteNames } from '@/routes/routes'
 import { TaxonName } from '@/routes/endpoints'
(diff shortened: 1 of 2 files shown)


app/javascript/vue/tasks/nomenclature/new_taxon_name/components/manageSynonym.vue
```diff
@@ -206,7 +206,9 @@ export default {
 
     loadTaxon(id) {
       if (window.confirm('Are you sure you want to load this taxon name?')) {
-        window.open(`/tasks/nomenclature/new_taxon_name/${id}`, `_self`)
+        window.open(
+          `/tasks/nomenclature/new_taxon_name?taxon_name_id=${id}`, `_self`
```
jlpereira, line 210: Thanks for fix this Tom, maybe it's a good time to replace it with RouteNames.NewTaxonName

PR #4202, opened 2025-02-13
FO identifier and other changes

kleintom:
This is mostly changes to the Catalog Number panel (not currently in use) in New FO, but also:
* require `total` to be > 0 when not absent (already required to be == 0 when absent).
* show items in the Recents modal as clickable with pointer css

### DIFF
--- a/app/javascript/vue/tasks/field_occurrences/new/components/Recent.vue
+++ b/app/javascript/vue/tasks/field_occurrences/new/components/Recent.vue
@@ -20,19 +20,30 @@
         <thead>
           <tr>
             <th>Total</th>
+            <th />
           </tr>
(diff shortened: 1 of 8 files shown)


app/javascript/vue/tasks/field_occurrences/new/components/Recent.vue
jlpereira, line 1: @kleintom maybe it's a good time to change this to be consistent like the other tasks, instead make the row clickable, lets add a button with a new column for it:

```vue
<VBtn
  circle
  color="primary"
  @click="() => selectFO(item)"
>
  <VIcon
    name="pencil"
    x-small
  />
</VBtn>
```

PR #4210, opened 2025-02-18
SimpleMappr fixups

kleintom:
The main fix here is to the zip call: with existing code the number of points associated with the first CO in the COs list becomes the max number of points displayed for any CO in the list.

The other commits handle some edge cases associated with SimpleMappr behavior:
* make sure every CO has a header name, otherwise SimpleMappr (rightfully) gets confused
* if there's only one CO - so only one column of data sent to SimpleMappr - then SimpleMappr parses wrong and errors out
* SimpleMappr won't process more than 13 columns in a file (extra columns are silently dropped). (It may process more if you add extra columns by hand.)
* Columns (COs) with no geographic data don't appear in any way in the output of SimpleMappr, so I've dropped them in the TW UI output as well to give more room for 13 populated COS (I'm not sure if this is a "good" thing or not).

### DIFF
--- a/app/helpers/vendor/simple_mappr_helper.rb
+++ b/app/helpers/vendor/simple_mappr_helper.rb
@@ -5,14 +5,14 @@ module Vendor::SimpleMapprHelper
     # TODO: Add support for FieldOccurrence, or by Otu Filter
 
     if params[:collection_object_query]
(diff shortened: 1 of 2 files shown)


mjy:
Thanks @kleintom

PR #4224, opened 2025-02-27
Support (multiply encoded) brackets in venn query strings

kleintom:
For example, pasting something like this as the 'b' query string (which I copied from 'JSON response'):
```
http://localhost:3000/collection_objects/filter.json?per=50&collecting_event_id%5B%5D=87629&extend%5B%5D=dwc_occurrence&extend%5B%5D=repository&extend%5B%5D=current_repository&extend%5B%5D=collecting_event&extend%5B%5D=taxon_determinations&extend%5B%5D=identifiers&exclude%5B%5D=object_labels&page=1
```

I think there are two issues:
* venn doesn't support inputting query strings that already include encoded brackets (because they get encoded again on their way to the server(?) and ::Rack::Utils.parse... functions only unencode once)
* venn doesn't support any kind of brackets in its 'b' query, encoded or not --> use ::Rack::Utils.parse_nested_query instead of ::Rack::Utils.parse_query (more details in the commit message)

### DIFF
--- a/lib/queries/query/filter.rb
+++ b/lib/queries/query/filter.rb
@@ -316,7 +316,6 @@ module Queries
 
     def venn_mode
       v = @venn_mode.to_s.downcase.to_sym
-      v  = :ab if v.blank?
       if [:a, :ab, :b].include?(v)
         v
       else
@@ -765,7 +764,7 @@ module Queries
         q = referenced_klass.all
       end
 
-      if venn && !api
+      if venn_mode && venn && !api
         q = apply_venn(q)
       end
 
--- a/lib/queries/query/filter.rb
+++ b/lib/queries/query/filter.rb
@@ -707,8 +707,18 @@ module Queries
     end
 
     def venn_query
-      u = ::Addressable::URI.parse(venn)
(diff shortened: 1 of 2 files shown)


kleintom:
(Note it may look like I rewrote/repurposed an existing spec, which technically I did, but the one I rewrote is a duplicate of the previous spec.)

mjy:
@kleintom Thanks. Can you comment on the max param you use here, i.e. what's it doing, and cross reference this PR to the issue on Venn?

kleintom:
> Can you comment on the max param you use here, i.e. what's it doing

Yeah, I added that late, it just always wigs me out a little to rely on calling a non-local function to avoid an infinite loop/denial-of-service type situation there, even though it seems pretty unlikely. I'm fine with removing it or adding a comment, let me know.

>and cross reference this PR to the issue on Venn?

It is, on the first commit, but following @jlpereira's pattern I started adding the reference to the end of the commit title instead of the front, where github expands it and then hides the rest of the title. I guess this way you have to expand every message to make sure the ref wasn't forgotten... !

mjy:
Thanks for clarification on both fronts, glad this made it in.

PR #4232, opened 2025-03-10
Add `project_scope_tag_id` scope-to-tag functionality to some api otu queries

kleintom:


kleintom:
In progress - there may still be some additions.

app/helpers/otus/catalog_helper.rb
```diff
@@ -65,9 +65,17 @@ def otu_descendants_and_synonyms(
...
-    max_descendants_depth: Float::INFINITY
+    max_descendants_depth: Float::INFINITY,
+    project_scope_tag_id: nil
   )
-    s = Otu.where(taxon_name_id: otu.taxon_name_id).where.not(id: otu.id).where.not(name: (otu.name.presence)).to_a
+    q = Otu.where(taxon_name_id: otu.taxon_name_id).where.not(id: otu.id).where.not(name: (otu.name.presence))
+
+    if project_scope_tag_id.present?
+      q = q.joins(:tags)
+        .where("tags.tag_object_type = 'Otu' AND tags.keyword_id = #{project_scope_tag_id}")
```
kleintom, line 75: I don't need `tags.tag_object_type = 'Otu'` here, it's covered by `joins(:tags)`

app/helpers/otus/catalog_helper.rb
```diff
@@ -82,11 +90,22 @@ def otu_descendants_and_synonyms(
...
+      descendants = otu.taxon_name.descendants
+        .where(parent: otu.taxon_name).that_is_valid
+
+      if project_scope_tag_id
+        # The project scope tag has the property that any ancestor of a tagged
+        # otu is also tagged, so this is legit.
+        descendants = descendants
+          .joins(:otus)
+          .joins("JOIN tags ON tags.tag_object_type = 'Otu' AND tags.tag_object_id = otus.id")
+          .where("tags.keyword_id = #{project_scope_tag_id}")
```
mjy, line 107: `where(:tags: {keyword_id: project_scope_tag_id}`

app/helpers/otus/catalog_helper.rb
```diff
@@ -82,11 +90,22 @@ def otu_descendants_and_synonyms(
...
     if otu.taxon_name
-      descendants = otu.taxon_name.descendants.where(parent: otu.taxon_name).that_is_valid
+      descendants = otu.taxon_name.descendants
+        .where(parent: otu.taxon_name).that_is_valid
+
+      if project_scope_tag_id
+        # The project scope tag has the property that any ancestor of a tagged
+        # otu is also tagged, so this is legit.
+        descendants = descendants
+          .joins(:otus)
```
mjy, line 105: `.joins(otus: [:tags])` probably

app/helpers/otus/catalog_helper.rb
```diff
@@ -101,9 +120,12 @@ def otu_descendants_and_synonyms(
         end
         data[:leaf_node] = data[:descendants].empty?
       else
-        data[:leaf_node] = descendants.where.not(id: similar_otus).none?
```
mjy, line 104: see about adding `.unscope(:order)` on call using `descendant` etc. it can save a lot of time`

lib/queries/otu/autocomplete.rb
```diff
@@ -26,6 +26,10 @@ class Autocomplete < Query::Autocomplete
       #   if 'true' then only #name = query_string results are returned (no fuzzy matching)
       attr_accessor :exact
 
+      # @return Integer, nil
+      # Only return otus tagged with this tag id
+      attr_accessor :project_scope_tag_id
```
mjy, line 31: I'm worried about the scope in various other places, there are various other filter results that need this scoping that are used.  Before we get too far I'd like to enumerate where else we'd need this tag param to ensure consistency across the site.  Remember too that for basic OTUs this works fine, but as as soon as we add all the filters/search that others want to use this will be quite difficult to maintain without a lot of thought.

app/helpers/otus/catalog_helper.rb
```diff
@@ -82,11 +90,22 @@ def otu_descendants_and_synonyms(
...
+             # TODO: ignores project_scope_tag_id (will require a GA instead).
              common_names: (common_names ? otu_inventory_common_names(otu, language_alpha2) : []),
              descendants: []}
 
     if otu.taxon_name
-      descendants = otu.taxon_name.descendants.where(parent: otu.taxon_name).that_is_valid
+      descendants = otu.taxon_name.descendants
+        .where(parent: otu.taxon_name).that_is_valid
+
+      if project_scope_tag_id
```
mjy, line 101: @kleintom Internal to catalog, rather than a tag Id. I think we should try to use an Otu scope as the join, that would let us pre-define any set of OTUs.  The concern would be performance, but let's experiment with it.

External to Catalog their might be a configure wrapper, that handles creating the OTU scopes.

So in the tag case the `otu_bounding_scope= Otu.joins(:tags).where(tags: {keyword_id: 123})`

kleintom:
I'm going to close this one, I think everyone agrees we'd like a more general approach here.

PR #4240, opened 2025-03-13
Make DwcOccurrence and DWCO hooks FieldOccurrence-aware

kleintom:
I'm not expecting this in the next release either way, but it would be nice to add one more fix for an existing issue here - the STR here also apply in the FO instead of CO case now:
* create a CO with a CE, save them; note the DwcOccurrence item that is created
* add a georef to the CE
* see the georef data has not been added to the CO's DWCO item.

The DWCO for the CO in this case is intended to be updated by this Georef dwco hook:
https://github.com/SpeciesFileGroup/taxonworks/blob/2ae53cf17bec59e7235dd74b6957ca8d05b00f81/app/models/concerns/shared/dwc_occurrence_hooks.rb#L19

My best guess at this point is that the reason it fails in this case is:
* if you nested-save/update twice on a create, the second save will have its `saved_changes` (and `previous_changes`) empty/`{}` (according to chatgpt... fits with what I'm seeing, I think)
* if you create a georef using nested attributes for its geographic_item, as we do, then two saves occur *on the georef* during create

As near as I can tell, this is something rails is doing (not our code): i removed *all* includes from Georef and GeographicItem, removed all validations, removed acts_as_list from georef, removed all callbacks, and then created a new georef in the console with/out a nested geographic_item, with only an after_save on georef that I added to log the value of `saved_changes`:
* Without a nested geographic_item:
```
taxon-works(dev)> Georeference.create!(collecting_event_id: CollectingEvent.first.id, type: "Georeference::Leaflet", position: 56, created_by_id: 154, updated_by_id: 154, project_id: 13, geographic_item_id: GeographicItem.last.id)
  CollectingEvent Load (0.8ms)  SELECT "collecting_events".* FROM "collecting_events" ORDER BY "collecting_events"."id" ASC LIMIT $1  [["LIMIT", 1]]
  GeographicItem Load (0.5ms)  SELECT "geographic_items".* FROM "geographic_items" ORDER BY "geographic_items"."id" DESC LIMIT $1  [["LIMIT", 1]]
  TRANSACTION (0.4ms)  BEGIN
[...]

PR #4243, opened 2025-03-14
Refactor DWCO hooks callbacks for create/delete handling

kleintom:
@mjy see what you think here please when you get a chance - maybe you'll see some better alternatives. I'll want to add some specs in any case before merging.

There were two issues:
* `after_save_commit` wasn't getting called on Georef creates, apparently because georef create nested geo_items, and in that case there's a nested save during create which causes saved_changes to be {} in after_save_commit (more details at #4240)
  * Downside to this fix: we now update dwcos (potentially expensive) even if there were no saved_changes

* the hook wasn't working on deletes: in after_destroy the record is already gone, so dwc_occurrences doesn't find what you want it to. (And if the record was still in the db, so dwc_occurrences finds what you want it to, then you update dwcos as though your record that's about to disappear still exists!).
  * Downside to this fix: if the destroy was done in a transaction that fails then we've marked some dwcos for update that shouldn't have been so marked. (Seems fine to me.)

Then there was another issue: apparently joins in a scope break update_all, so the previous version was updating _every_ DwcOccurrence. This version seems safe.

### DIFF
--- a/app/models/concerns/shared/dwc_occurrence_hooks.rb
+++ b/app/models/concerns/shared/dwc_occurrence_hooks.rb
@@ -16,18 +16,19 @@ module Shared::DwcOccurrenceHooks
     #   See also Shared::IsDwcOccurrence
     attr_accessor :no_dwc_occurrence
 
-    after_save_commit :update_dwc_occurrence, if: :saved_changes?, unless: :no_dwc_occurrence
-    after_destroy :update_dwc_occurrence
+    after_save :process_dwc_occurrences, if: :saved_changes?, unless: :no_dwc_occurrence
+    after_destroy :process_dwc_occurrences
 
-    def update_dwc_occurrence
+    def process_dwc_occurrences
       t = dwc_occurrences.count
-      q = dwc_occurrences.unscope(:select).select('dwc_occurrences.id', 'occurrenceID', :dwc_occurrence_object_type, :dwc_occurrence_object_id, :is_flagged_for_rebuild)
 
       begin
         # If the scope is returning every object at this point (arbitrarily > 5k records), then the scope is badly coded.
         raise TaxonWorks::Error if t > 5000 && (t == DwcOccurrence.where(project_id:).count)
 
         if t < 100
+          q = dwc_occurrences.reselect('dwc_occurrences.id', :occurrenceID, :dwc_occurrence_object_type, :dwc_occurrence_object_id, :is_flagged_for_rebuild)
+
           q.find_each do |d|
             d.dwc_occurrence_object.set_dwc_occurrence
           end
--- a/app/models/concerns/shared/dwc_occurrence_hooks.rb
+++ b/app/models/concerns/shared/dwc_occurrence_hooks.rb
@@ -16,10 +16,10 @@ module Shared::DwcOccurrenceHooks
     #   See also Shared::IsDwcOccurrence
     attr_accessor :no_dwc_occurrence
 
-    after_save :process_dwc_occurrences, if: :saved_changes?, unless: :no_dwc_occurrence
+    after_save_commit :update_dwc_occurrence, unless: :no_dwc_occurrence
     around_destroy :process_destroy
 
-    def process_dwc_occurrences
+    def update_dwc_occurrence
       t = dwc_occurrences.count
 
       begin
@@ -66,11 +66,11 @@ module Shared::DwcOccurrenceHooks
           :dwc_occurrence_object_type, :dwc_occurrence_object_id).to_a
       else
         ::DwcOccurrence
-            .where(id: dwc_occurrences.reselect(:id))
-            .update_all(is_flagged_for_rebuild: true)
+          .where(id: dwc_occurrences.reselect(:id))
+          .update_all(is_flagged_for_rebuild: true)
       end
 
-      yield
+      yield # destroy
 
       if to_be_processed_now.any?
         to_be_processed_now.each { |d|
--- a/app/models/concerns/shared/dwc_occurrence_hooks.rb
+++ b/app/models/concerns/shared/dwc_occurrence_hooks.rb
@@ -17,7 +17,7 @@ module Shared::DwcOccurrenceHooks
     attr_accessor :no_dwc_occurrence
 
     after_save :process_dwc_occurrences, if: :saved_changes?, unless: :no_dwc_occurrence
-    after_destroy :process_dwc_occurrences
+    around_destroy :process_destroy
 
     def process_dwc_occurrences
       t = dwc_occurrences.count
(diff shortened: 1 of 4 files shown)


mjy:
Thanks @kleintom 

It feels like that if Georeferences are being created in a nested manner they could call up to the CE to trigger a rebuild at the CE level? 

I'm not sure, but It feels like we can solve this with a more general pattern specific to all nested attributes.  Perhaps exploring has_nested_attributes hooks (maybe you did this).

One thing certain- we definitely need specs with nested pattern to cover this.

```
ce.update!(georeferences_attributes: [{}])...
```

app/models/concerns/shared/dwc_occurrence_hooks.rb
```diff
@@ -16,24 +16,53 @@ module Shared::DwcOccurrenceHooks
...
+          q = dwc_occurrences.reselect('dwc_occurrences.id', :occurrenceID, :dwc_occurrence_object_type, :dwc_occurrence_object_id, :is_flagged_for_rebuild)
+
           q.find_each do |d|
             d.dwc_occurrence_object.set_dwc_occurrence
           end
-        else
-          dwc_occurrences.update_all(is_flagged_for_rebuild: true) # Quickly mark all records requiring rebuild
-          ::DwcOccurrenceRefreshJob.perform_later(project_id:, user_id: Current.user_id)
+        else # flag_only || t >= 100
+          # dwc_occurrences.update_all(...) can update *every* dwco depending on how
```
mjy, line 57: The hope is that given the right callback (when object is not destroyed) this never happens.  If there are specific ones doing this let's identify them.

kleintom: This `update_all` issue was surprising to me and I might not have been clear on what I was seeing - here's an example from my current local db:
```sql
taxon-works(dev)> Georeference.first.dwc_occurrences.count
  Georeference Load (0.8ms)  SELECT "georeferences".* FROM "georeferences" ORDER BY "georeferences"."id" ASC LIMIT $1  [["LIMIT", 1]]
  DwcOccurrence Count (3.6ms)  SELECT COUNT(*) FROM ( (SELECT DISTINCT "dwc_occurrences".* FROM "dwc_occurrences" JOIN collection_objects co on dwc_occurrence_object_id = co.id AND dwc_occurrence_object_type = 'CollectionObject' JOIN georeferences g on co.collecting_event_id = g.collecting_event_id WHERE "g"."id" = 4) UNION (SELECT DISTINCT "dwc_occurrences".* FROM "dwc_occurrences" JOIN field_occurrences fo on dwc_occurrence_object_id = fo.id AND dwc_occurrence_object_type = 'FieldOccurrence' JOIN georeferences g on fo.collecting_event_id = g.collecting_event_id WHERE "g"."id" = 4)) as dwc_occurrences
=> 31
taxon-works(dev)> Georeference.first.dwc_occurrences.update_all(is_flagged_for_rebuild: true)
  Georeference Load (0.3ms)  SELECT "georeferences".* FROM "georeferences" ORDER BY "georeferences"."id" ASC LIMIT $1  [["LIMIT", 1]]
  DwcOccurrence Update All (1415.3ms)  UPDATE "dwc_occurrences" SET "is_flagged_for_rebuild" = $1  [["is_flagged_for_rebuild", true]]
=> 69271
taxon-works(dev)> ::DwcOccurrence
taxon-works(dev)> .where(id: Georeference.first.dwc_occurrences.reselect(:id))
taxon-works(dev)>             .update_all(is_flagged_for_rebuild: true)
  Georeference Load (0.5ms)  SELECT "georeferences".* FROM "georeferences" ORDER BY "georeferences"."id" ASC LIMIT $1  [["LIMIT", 1]]
  DwcOccurrence Update All (13.9ms)  UPDATE "dwc_occurrences" SET "is_flagged_for_rebuild" = $1 WHERE "dwc_occurrences"."id" IN (SELECT "dwc_occurrences"."id" FROM ( (SELECT DISTINCT "dwc_occurrences".* FROM "dwc_occurrences" JOIN collection_objects co on dwc_occurrence_object_id = co.id AND dwc_occurrence_object_type = 'CollectionObject' JOIN georeferences g on co.collecting_event_id = g.collecting_event_id WHERE "g"."id" = 4) UNION (SELECT DISTINCT "dwc_occurrences".* FROM "dwc_occurrences" JOIN field_occurrences fo on dwc_occurrence_object_id = fo.id AND dwc_occurrence_object_type = 'FieldOccurrence' JOIN georeferences g on fo.collecting_event_id = g.collecting_event_id WHERE "g"."id" = 4)) as dwc_occurrences)  [["is_flagged_for_rebuild", true]]
=> 31
```
My take has been that there's nothing wrong per se with the `Georeference.first.dwc_occurrences` callback scope (31 is the expected number in this case), except that it's incompatible in some (as-yet-unknown-to-me) way with `update_all`. Are you saying you want to investigate that incompatibility and see if we can make all of these dwc_occurrences scopes compatible with update_all (instead of just working around the issue, as I did here)? I'm fine with doing that, just want to make sure I'm on the same page, thanks.

p.s. I'll be pushing some other updates on this soon as well.

mjy:
>  Are you saying you want to investigate that incompatibility and see if we can make all of these dwc_occurrences scopes compatible with update_all

Yes, in part, definitely this.

mjy:
@kleintom That Georeference `dwc_occurrences` is definitely messed up, exploring it too

mjy:
For one thing it shouldn't be `update_all`, but `update_columns`.

mjy:
@kleintom Clues: https://github.com/rails/rails/issues/44861#issuecomment-1114221816

mjy:
@kleintom we should use `execute` and just raw SQL to fix the issue and update all the recors in question, update_all is going to be a headache.

mjy:
@kleintom I pushed a work-around to test to development.  Turns out wrapping to `in_batches` might work as it basically gets back to the id based approach

kleintom:
@mjy I force pushed but left in one of my earlier commits switching to after_save instead of after_save_commit, in order to have the commit message pros/cons of each (as I was seeing them).

A couple followups I'm seeing, which you may already know about:
* I left an xspec in to test that DwCOs are getting updated on annotation updates, like Protocols, Notes, Citations, etc. (they aren't currently, at least the few I looked at).
* In one of the commit messages I mention the thought of preferring after_save_commit to after_save because it runs later as a callback, meaning it catches more data updated by other callbacks. But it still doesn't run after other after_commit callbacks, so there are still issues with, e.g.
  * add a TD, now notice the TD's name in TW is misspelled, so correct it --> the cached TN values are updated *after* the dwco after_save_commit, so in the dwco specificEpithet is updated, e.g., but scientificName (cached) is not. (A bit contrived, but still something to worry about I think?)

mjy:
Thanks @kleintom.  Once tests green I'm going to merge this.  I'm on a branch moving away from boolean rebuild flag to a UUID identified set of records, to facilitate Community object updates.  Should be a straightforward refactor.

PR #4253, opened 2025-03-24
Fix navbar links in Browse Asserted Distributions task

kleintom:
STR:
* Find an otu with siblings and/or a parent
* Load that otu in the 'Browse asserted distributions' task
See 'Parent', 'Previous', and/or 'Next' tags, but no links for those tags.


kleintom:
There were two issues here:
* the server was sending otu tags, the ui was expecting labels - I went with labels
* the link to go with the labels had '/index' at its end where it wasn't needed

PR #4294, opened 2025-04-21
4287 lead otus list

kleintom:
@proceps, I think this is ready for feedback, could you give it a try when you get a chance? Please re-request things that you requested during the hackathon, if I didn't add them it may just be because I forgot.

The way things currently work:
* the only way to create a key with an otu list is to do it from Filter Otus, where there's an option under the Radial Matrix menu (maybe that should be Radial Keys now, or something). You can't add otu lists to keys that weren't started that way.
* all of the existing key operations should work on a key that has otu lists. When you delete couplets or more generally subtrees, their leaf otus get bundled and added to either the parent of the delete or a new placeholder lead.

Technical points:
* Only leaf leads keep a list of their otus, leads with children know nothing.
"Issues":
* there's no way to tell if a key is a lead_items key other than checking every leaf
* that means we do extra work to check for lead_items in some "regular key" calls, even when we're on a key that has no lead_items and never will.
I don't currently have a solution for that (aside from putting a bool on every lead); suggestions welcome.
* you won't see the link to view the print version of your key with otu-list info unless you're on a couplet with a leaf node

Really technical point:
* The only gotcha I'm aware of is that if you insert a new couplet between existing couplets (or if you insert a key), you're never presented with the option to add otus to the new inserted lead that has no children (because it doesn't know it's in a lead_items key, and it would be expensive (well, painful at least) to check on every render).

TODO:
* specs, more cleanup
* maybe an option to remove all otu lists, once your key is "finished"


proceps:
1. I enter new lead, add text1 and text2, after selection of a single OTU, the text disappeared on both sides (it should stay)
2. I can select a particular OTU on one side only. Expected behavior: by default all OTUs are on the right side. I click the box on the left size, the OTU is moved from the right to left, when I click on the OTU on the right after that, it should be selected on both sides (the same OTU can go both ways), if two are selected, I should be able to deselect either of those two
3. The future couplets are displayed on the right lead, I would prefer to see a complete key on the top part (in the Previous leads frame), probably as a scrollable frame. And preferably as a dichotomous key with lead number (as in printed output, but this could be implemented later).
4. When a single OTU left, should it be automatically set as a lead OTU?
5. update the button names, OTU should be capitalized.
6. The upper section of the interface takes too much space. The title and the buttons need some better layout
7. Create and edit the next couplet and Insect a couplet buttons could also go one line up.

kleintom:
@proceps could you take another look? Thanks.

> 1. I enter new lead, add text1 and text2, after selection of a single OTU, the text disappeared on both sides (it should stay)

Nice catch, this should be fixed now.
>     2. I can select a particular OTU on one side only. Expected behavior: by default all OTUs are on the right side. I click the box on the left size, the OTU is moved from the right to left, when I click on the OTU on the right after that, it should be selected on both sides (the same OTU can go both ways), if two are selected, I should be able to deselect either of those two

image

Updated this so you can add an otu to any number of leads: the green circle you should think of as a (horizontal) radio select, where selecting an otu on one lead deselects it from all others; the green box you should think of as a (horizontal) checkbox select, where you can use it to add to your selection without affecting any others already selected. When you have an otu selected on more than one lead, a red button appears to deselect.


>     3. The future couplets are displayed on the right lead, I would prefer to see a complete key on the top part (in the Previous leads frame), probably as a scrollable frame. And preferably as a dichotomous key with lead number (as in printed output, but this could be implemented later).

I added the print key below the Previous leads frame.
> 
>     4. When a single OTU left, should it be automatically set as a lead OTU?

If I'm starting on my first couplet and all of the otus are on the right, and then I add one to the left, I don't want to automatically set it as the lead otu because I might be (probably will) add others as well to the left.
The only way to get exactly one otu left in the list is to create a new couplet from a lead where the otu was selected but other otus were not. Then that otu starts on the right lead of the new couplet, so should it get auto-added to the right lead? Probably left lead would be more likely, but I feel like that's too much to make that decision for people.

> 
>     5. update the button names, OTU should be capitalized.
Updated, thanks.

proceps:
Add an OTU option is not available before initial list of OTUs added to the Key, There should be a way to populate the list manually (similar to an Image matrix, there are several options: manually one by one, export from an existing key or matrix)

kleintom:
> Add an OTU option is not available before initial list of OTUs added to the Key, There should be a way to populate the list manually (similar to an Image matrix, there are several options: manually one by one, export from an existing key or matrix)

I added a button from Edit key to be able to start an otus list, but it's only available if one of the root children doesn't have any children. That's because we need a place to put the otus you're adding. (This should maybe require that *none* of the root children has any children, because those that do are never going to get list otus you're adding. We could instead allow a new list to be started from *anywhere* in the key, but that's more complex and for a later issue I think, if it's desired.) Given that restriction, I haven't added the ability to add otus from filter otu to an existing key, since the existing key would basically have to be new anyway.

PR #4299, opened 2025-04-24
Fix fetching of FO TDs from FO radial form

kleintom:
STR: Create an FO in the New FO task, then click on radial forms and select TDs --> 422

### DIFF
--- a/config/routes/data.rb
+++ b/config/routes/data.rb
@@ -334,6 +334,9 @@ end
 
 resources :field_occurrences do
   concerns [:data_routes]
+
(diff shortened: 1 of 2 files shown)


kleintom:
Sub-resource here needed because the TD slice uses annotatorExtend.js, which does:
https://github.com/SpeciesFileGroup/taxonworks/blob/96dc3273d67e7d022245a09faf80185e00c981ac/app/javascript/vue/components/radials/object/components/annotatorExtend.js#L42-L48

Hard to keep track of all of these!

PR #4304, opened 2025-04-25
Make sure all container labels are strings, not numbers

kleintom:
Downstream code assumes they're strings.

STR:
1) Create a CO without a dwc_catalog_number or dwc_scientific_name (in my case toy test data).
2) Add a container item to that CO so that both are now in a container.
3) Load 'New container' for either of those COs
Get a javascript exception on String#shorten trying to shorten the label for either CO, which is an id number, not a string.


kleintom:
> If you want to hold off on this a bit more I think I might have another container-related commit to add.

Discussed my other issue on gitter, so that's covered, I don't think I have anything more to add here now.

PR #4305, opened 2025-04-25
Add common names as an option to OTU autocomplete

neolefty:
Add an optional parameter `include_common_name` to the OTU autocomplete endpoint `/api/v1/otus/autocomplete`.

How it looks in TaxonPages:

image

### DIFF
--- a/app/helpers/otus_helper.rb
+++ b/app/helpers/otus_helper.rb
@@ -1,8 +1,8 @@
 module OtusHelper
 
-  def otu_tag(otu)
+  def otu_tag(otu, include_common_names = false)
     return nil if otu.nil?
-    a = otu_tag_elements(otu)
+    a = otu_tag_elements(otu, include_common_names)
     a.push taxon_name_type_short_tag(otu.taxon_name)
     content_tag(:span, a.compact.join(' ').html_safe, class: :otu_tag)
   end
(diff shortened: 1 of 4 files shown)


mjy:
@neolefty Thanks for this.  I've looked it over but don't have the time to think about the consequences until after TWT.  

I will say that we don't want to parameterize `otu_tag`, it opens up too much of a can of worms.  That needs to be the internal method that largely "just works" to identify the object that is in question.  It has to be a jack of all, and the same as every other <object>_tag.

I'll also note that we have a new auto-select paradigm coming, slowly, it has provisions for adapting responses that may better accomodate this.

So- please let us stew on this for a bit.

app/helpers/otus_helper.rb
```diff
@@ -1,8 +1,8 @@
 module OtusHelper
 
-  def otu_tag(otu)
+  def otu_tag(otu, include_common_names = false)
```
neolefty, line 3: @mjy writes:

> I will say that we don't want to parameterize otu_tag, it opens up too much of a can of worms. That needs to be the internal method that largely "just works" to identify the object that is in question. It has to be a jack of all, and the same as every other _tag.

I can understand that — to avoid breaking things, I've been adding optional params whose default values leave the behavior unchanged, especially with components in TaxonPages. But I can see how the warts accumulate that way. Each one probably deserves a discussion with someone who has a broader view of the architecture.

neolefty: Perhaps rather than leaving this PR hanging we should close it as rejected @mjy ?

PR #4311, opened 2025-04-30
Fixes syntax in initialize.rake

gdower:
I think coloring used to work without Rainbow() but no longer.


gdower:
The connection_config also changed so I updated Support::Database to the new method:

ActiveRecord::Base.connection_db_config.configuration_hash

PR #4326, opened 2025-05-05
Slow cached map item translates

kleintom:
This fixes the gazetteer branch being 3x slower than old development on recreating the cached map index. On OSF project locally it's now:
gaz branch: 1h2m
old dev branch: 1h20m

Side note: I noticed that most of the time creating translations for ADs for OSF (at least the version of OSF I have) is spent on just _one_ geographic_item: North America (61718). Checking for intersection with 3,669 ne_states GAs takes a _long_ time in that case. No fix for that yet.

### DIFF
--- a/lib/tasks/maintenance/cached/maps.rake
+++ b/lib/tasks/maintenance/cached/maps.rake
@@ -103,7 +103,7 @@ namespace :tw do
             h = CachedMapItem.cached_map_name_hierarchy(o.geographic_item_id)
 
             z = CachedMapItem.where(geographic_item_id: o.geographic_item_id)
-              .where.not(untranslated: true)
+              .where('untranslated IS NULL OR untranslated = FALSE')
 
             #puts 'Size: ' + z.size.to_s
 
--- a/lib/tasks/maintenance/cached/maps.rake
+++ b/lib/tasks/maintenance/cached/maps.rake
@@ -105,7 +105,7 @@ namespace :tw do
             z = CachedMapItem.where(geographic_item_id: o.geographic_item_id)
               .where.not(untranslated: true)
 
-            puts 'Size: ' + z.size.to_s
+            #puts 'Size: ' + z.size.to_s
 
             z.update_all(
               level0_geographic_name: h[:country],
@@ -113,8 +113,8 @@ namespace :tw do
               level2_geographic_name: h[:county]
             )
 
-            puts o.geographic_item_id
-            puts h
+            #puts o.geographic_item_id
+            #puts h
           end
           puts 'Done labelling cached map items.'
         end
@@ -277,7 +277,7 @@ namespace :tw do
           begin
             #  print "#{id}: "
             t = CachedMapItem.translate_geographic_item_id(
-              geographic_item_id, geographic_area_based, ['ne_states']
+              geographic_item_id, geographic_area_based, false, ['ne_states']
             )
             # if t.present?
             #   print t.join(', ')
--- a/lib/tasks/maintenance/cached/maps.rake
+++ b/lib/tasks/maintenance/cached/maps.rake
@@ -255,15 +255,19 @@ namespace :tw do
           ids_in__ga.sort!
           ids_in__gz.sort!
 
-          Parallel.each(ids_in__ga, progress: 'build_cached_map_item_translations', in_processes: cached_rebuild_processes ) do |id|
+          Parallel.each(ids_in__ga, progress: 'build_cached_map_item_translations GA', in_processes: cached_rebuild_processes ) do |id|
             reconnected ||= CachedMapItemTranslation.connection.reconnect! || true
             process_asserted_distribution_translation(id, true)
           end
 
-          Parallel.each(ids_in__gz, progress: 'build_cached_map_item_translations', in_processes: cached_rebuild_processes ) do |id|
+          puts 'Geographic Area-based Asserted Distributions done.'
+
+          Parallel.each(ids_in__gz, progress: 'build_cached_map_item_translations GZ', in_processes: cached_rebuild_processes ) do |id|
             reconnected ||= CachedMapItemTranslation.connection.reconnect! || true
             process_asserted_distribution_translation(id, false)
           end
+
+          puts 'Gazetteer-based Asserted Distributions done.'
 
           puts 'Done.'
         end
--- a/lib/tasks/maintenance/cached/maps.rake
+++ b/lib/tasks/maintenance/cached/maps.rake
@@ -103,7 +103,7 @@ namespace :tw do
             h = CachedMapItem.cached_map_name_hierarchy(o.geographic_item_id)
 
             z = CachedMapItem.where(geographic_item_id: o.geographic_item_id)
-              .where('untranslated IS NULL OR untranslated = FALSE')
+              .where(untranslated: [nil, false])
 
             #puts 'Size: ' + z.size.to_s
 
--- a/app/controllers/tasks/geographic_areas/usage_controller.rb
+++ b/app/controllers/tasks/geographic_areas/usage_controller.rb
@@ -4,7 +4,7 @@ class Tasks::GeographicAreas::UsageController < ApplicationController
   # GET
   def index
     if params[:geographic_area_id].present?
-      @geogrpahic_areas = GeographicArea.where(id: params[:geographic_area_id])
+      @geographic_areas = GeographicArea.where(id: params[:geographic_area_id])
     else
       @geographic_areas = GeographicArea.all
     end
--- a/app/models/cached_map_item.rb
+++ b/app/models/cached_map_item.rb
@@ -191,6 +191,10 @@ class CachedMapItem < ApplicationRecord
   #   area (and so has a chance of already being associated with a cached map
   #   item/translation)
   #
+  # @param search_existing_translates [Boolean]
+  #   true if existing cached_map_item_translations should be checked for a
+  #   match.
+  #
   # @param data_origin Array, String
   #   like `ne_states` or ['ne_states, 'ne_countries']
   #
@@ -199,7 +203,8 @@ class CachedMapItem < ApplicationRecord
   #   Typical use, do not apply for Georeferences, apply -10km for AssertedDistributions
   #
   def self.translate_geographic_item_id(
-    geographic_item_id, geographic_area_based, data_origin = nil, buffer = nil
+    geographic_item_id, geographic_area_based,
+    search_existing_translates = true, data_origin = nil, buffer = nil
   )
     return nil if data_origin.blank?
 
@@ -208,6 +213,13 @@ class CachedMapItem < ApplicationRecord
     a = nil
 
     b = buffer
+
+    if search_existing_translates
+      a = translate_by_geographic_item_translation(
+        geographic_item_id, cached_map_type
+      )
+      return a if a.present?
+    end
 
     # All these methods depend on "prior knowledge" (not spatial calculations)
     if geographic_area_based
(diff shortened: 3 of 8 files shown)


mjy:
> North America (61718). Checking for intersection with 3,669 ne_states GAs takes a long time in that case. 

Maybe we should just hard code/exception that?  Though translations for non Gaz should be almost "fixed", and we're not that worried about time on those.

lib/tasks/maintenance/cached/maps.rake
```diff
@@ -103,18 +103,18 @@ namespace :tw do
             h = CachedMapItem.cached_map_name_hierarchy(o.geographic_item_id)
 
             z = CachedMapItem.where(geographic_item_id: o.geographic_item_id)
-              .where.not(untranslated: true)
+              .where('untranslated IS NULL OR untranslated = FALSE')
```
LocoDelAssembly, line 106: Per change does `.where(untranslated: [nil, false])` provide similar behavior?

```
CachedMapItem.where(geographic_item_id: 3).where(untranslated: [nil, false]).to_sql
 => "SELECT \"cached_map_items\".* FROM \"cached_map_items\" WHERE \"cached_map_items\".\"geographic_item_id\" = 3 AND (\"cached_map_items\".\"untranslated\" = FALSE OR \"cached_map_items\".\"untranslated\" IS NULL)"
 ```

Comment (kleintom) [reply lib/tasks/maintenance/cached/maps.rake:106]: Yah, that should be the same, thanks - I'll change.

**Comment by LocoDelAssembly** [comment]
```
3.3.7 :023 > GeographicItem
3.3.7 :024 >       .joins(:geographic_areas_geographic_items)
3.3.7 :025 >       .where(geographic_areas_geographic_items: { data_origin: })
3.3.7 :026 > .merge(GeographicItem.intersecting(:multi_polygon, geographic_item_id)).select(:id).explain
  GeographicItem Load (1648.3ms)  SELECT "geographic_items"."id" FROM "geographic_items" INNER JOIN "geographic_areas_geographic_items" ON "geographic_areas_geographic_items"."geographic_item_id" = "geographic_items"."id" WHERE "geographic_areas_geographic_items"."data_origin" = $1 AND ST_Intersects(CASE ST_GeometryType(("geographic_items"."geography")::geometry) WHEN 'ST_MultiPolygon' THEN "geographic_items"."geography" ELSE NULL END, (SELECT "geographic_items"."geography" FROM "geographic_items" WHERE "geographic_items"."id" = 1))  [["data_origin", "ne_countries"]]
 => 
EXPLAIN SELECT "geographic_items"."id" FROM "geographic_items" INNER JOIN "geographic_areas_geographic_items" ON "geographic_areas_geographic_items"."geographic_item_id" = "geographic_items"."id" WHERE "geographic_areas_geographic_items"."data_origin" = $1 AND ST_Intersects(CASE ST_GeometryType(("geographic_items"."geography")::geometry) WHEN 'ST_MultiPolygon' THEN "geographic_items"."geography" ELSE NULL END, (SELECT "geographic_items"."geography" FROM "geographic_items" WHERE "geographic_items"."id" = 1)) [["data_origin", "ne_countries"]]
                                                                        QUERY PLAN
----------------------------------------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=9.15..5304.08 rows=1 width=4)
   InitPlan 1 (returns $0)
     ->  Index Scan using geographic_items_pkey on geographic_items geographic_items_1  (cost=0.42..8.44 rows=1 width=7468)
           Index Cond: (id = 1)
   ->  Index Scan using index_geographic_areas_geographic_items_on_data_origin on geographic_areas_geographic_items  (cost=0.29..131.27 rows=252 width=4)
         Index Cond: ((data_origin)::text = 'ne_countries'::text)
   ->  Index Scan using geographic_items_pkey on geographic_items  (cost=0.42..20.48 rows=1 width=4)
         Index Cond: (id = geographic_areas_geographic_items.geographic_item_id)
         Filter: st_intersects(CASE st_geometrytype((geography)::geometry) WHEN 'ST_MultiPolygon'::text THEN geography ELSE NULL::geography END, $0)
```
_(replaced `pluck` with `select` so I can execute `explain`)_

Why is `CASE` needed when comparing interaction and why is limited to MultiPolygon only?

Full SQL code:
```sql
SELECT "geographic_items"."id"
FROM "geographic_items"
  INNER JOIN "geographic_areas_geographic_items" ON "geographic_areas_geographic_items"."geographic_item_id" = "geographic_items"."id"
WHERE "geographic_areas_geographic_items"."data_origin" = $1
  AND ST_Intersects(
    CASE
      ST_GeometryType(("geographic_items"."geography")::geometry)
      WHEN 'ST_MultiPolygon' THEN "geographic_items"."geography"
      ELSE NULL
    END,
    (
      SELECT "geographic_items"."geography"
      FROM "geographic_items"
      WHERE "geographic_items"."id" = 1
    )
  )
  ```

All above is equivalent to calling `CachedMapItem.translate_by_spatial_overlap(1, 'ne_countries', nil)` (or close).

Not totally sure, but it looks like it is not actually using spatial indexes, it is scanning over other indicies and still computing intersection one by one.

**Comment by kleintom** [comment]
> Why is `CASE` needed when comparing interaction and why is limited to MultiPolygon only?

Here's pre-gaz code: https://github.com/SpeciesFileGroup/taxonworks/blob/c46bf0fd808170facdedabad1461b6f262f56c49/app/models/cached_map_item.rb#L172-L176

I think the idea was that GA GIs are always multi-polygons, and the query is "get all GA ne_state GIs that intersect geographic_item_id", so the CASE was restricting to multi-polygon GIs to reduce work (I guess) since it knew what we wanted had to be a multi-polygon.

The version you reference here - my first gaz version - was written when I was less comfortable with what I was doing and just aimed to replicate the previous code. Technically I did but for whatever reason the execution of the query changed and made it much slower - my latest version in this PR removes the multi_polygon CASE because it ensures that the intersection only ranges over the GA GIs of type ne_states, which are all automatically multi_polygon.

**Comment by LocoDelAssembly** [comment]
Ah I see:
```ruby
taxon-works(prod)> GeographicItem.group(:type).count
=> {"GeographicItem::MultiPolygon"=>41462, "GeographicItem::Point"=>125651, "GeographicItem::Polygon"=>7559}
taxon-works(prod)> GeographicItem.joins(:geographic_areas_geographic_items).group(:type).count
=> {"GeographicItem::MultiPolygon"=>41462}
```
So indeed only MultiPolygon when tied to a geographic area.

Sorry, forgot to run experiments on this PR rather than `development` branch :facepalm:

LocoDelAssembly:
With same example it runs a bit more than 10 times faster:

```
3.3.7 :018 > GeographicItem
3.3.7 :019 >       .with(b: GeographicItem
3.3.7 :020 >         .joins(:geographic_areas_geographic_items)
3.3.7 :021 >         .where(geographic_areas_geographic_items: { data_origin: })
3.3.7 :022 >       )
3.3.7 :023 >       .from('b')
3.3.7 :024 >       .select('b.*')
3.3.7 :025 >       .where(
3.3.7 :026 >         'ST_Intersects(b.geography, (SELECT geography FROM geographic_items ' \
3.3.7 :027 >         'WHERE geographic_items.id = ?))', geographic_item_id
3.3.7 :028 >       ).select(:id).explain
  GeographicItem Load (166.4ms)  WITH "b" AS (SELECT "geographic_items".* FROM "geographic_items" INNER JOIN "geographic_areas_geographic_items" ON "geographic_areas_geographic_items"."geographic_item_id" = "geographic_items"."id" WHERE "geographic_areas_geographic_items"."data_origin" = $1) SELECT b.*, "id" FROM b WHERE (ST_Intersects(b.geography, (SELECT geography FROM geographic_items WHERE geographic_items.id = $2)))  [["data_origin", "ne_countries"], [nil, 1]]
 => 
EXPLAIN WITH "b" AS (SELECT "geographic_items".* FROM "geographic_items" INNER JOIN "geographic_areas_geographic_items" ON "geographic_areas_geographic_items"."geographic_item_id" = "geographic_items"."id" WHERE "geographic_areas_geographic_items"."data_origin" = $1) SELECT b.*, "id" FROM b WHERE (ST_Intersects(b.geography, (SELECT geography FROM geographic_items WHERE geographic_items.id = $2))) [["data_origin", "ne_countries"], [nil, 1]]
                                                                        QUERY PLAN
----------------------------------------------------------------------------------------------------------------------------------------------------------
 Hash Join  (cost=294.02..425.67 rows=1 width=39258)
   Hash Cond: (geographic_areas_geographic_items.geographic_item_id = geographic_items.id)
   InitPlan 1 (returns $0)
     ->  Index Scan using geographic_items_pkey on geographic_items geographic_items_1  (cost=0.42..8.44 rows=1 width=7468)
           Index Cond: (id = 1)
   ->  Index Scan using index_geographic_areas_geographic_items_on_data_origin on geographic_areas_geographic_items  (cost=0.29..131.27 rows=252 width=4)
         Index Cond: ((data_origin)::text = 'ne_countries'::text)
   ->  Hash  (cost=285.08..285.08 rows=17 width=39254)
         ->  Index Scan using index_geographic_items_on_geography on geographic_items  (cost=0.28..285.08 rows=17 width=39254)
               Index Cond: (geography && $0)
               Filter: st_intersects(geography, $0)
(11 rows)
 
3.3.7 :029 > CachedMapItem.translate_by_spatial_overlap(1, 'ne_countries', nil)
  GeographicItem Pluck (165.2ms)  WITH "b" AS (SELECT "geographic_items".* FROM "geographic_items" INNER JOIN "geographic_areas_geographic_items" ON "geographic_areas_geographic_items"."geographic_item_id" = "geographic_items"."id" WHERE "geographic_areas_geographic_items"."data_origin" = $1) SELECT "id" FROM b WHERE (ST_Intersects(b.geography, (SELECT geography FROM geographic_items WHERE geographic_items.id = $2)))  [["data_origin", "ne_countries"], [nil, 1]]
 => [41, 144, 179, 186, 194, 1, 30432] 
```

An alternative but without any speed improvement:
```ruby
3.3.7 :042 > GeographicItem.
3.3.7 :043 >   joins(:geographic_areas_geographic_items).
3.3.7 :044 >   where(geographic_areas_geographic_items: { data_origin: }).
3.3.7 :045 > where('ST_Intersects(geography, (SELECT geography FROM geographic_items WHERE geographic_items.id = ?))', geographic_item_id).pluck(:id)
  GeographicItem Pluck (169.0ms)  SELECT "geographic_items"."id" FROM "geographic_items" INNER JOIN "geographic_areas_geographic_items" ON "geographic_areas_geographic_items"."geographic_item_id" = "geographic_items"."id" WHERE "geographic_areas_geographic_items"."data_origin" = $1 AND (ST_Intersects(geography, (SELECT geography FROM geographic_items WHERE geographic_items.id = $2)))  [["data_origin", "ne_countries"], [nil, 1]]
 => [41, 144, 179, 186, 194, 1, 30432] 
```

I have also experimented with removing `pluck` and using `a.select(:id)` on second query when buffer is present. Took almost a minute just to compute `explain`, so looks it is better to do in two steps.
```ruby
  def self.translate_by_spatial_overlap(geographic_item_id, data_origin, buffer)
    return [] if geographic_item_id.blank?

    # !! Assumes all GeographicArea shapes were loaded to multi_polygon
    # (pre-adapts us to a single geometry field type), however be
    # aware of this assumption

    # This is a fast first pass, pure intersection
    a = GeographicItem
        .joins(:geographic_areas_geographic_items)
        .where(geographic_areas_geographic_items: { data_origin: })
        .where('ST_Intersects(geography, (SELECT geography FROM geographic_items WHERE geographic_items.id = ?))', geographic_item_id)

    return a.pluck(:id) if buffer.nil?

    # Refine the pass by smoothing using buffer/st_within
    return GeographicItem
      .where(id: a.select(:id))
      .where(
        GeographicItem.st_buffer_st_within_sql(geographic_item_id, 0.0, buffer)
      )
      .pluck(:id)
  end
```

Do you agree on merging @mjy?

PR #4328, opened 2025-05-05
Remove old GI centroid index

kleintom:
Creating the new index took 1m17s on my computer.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.2].define(version: 2025_03_21_032314) do
+ActiveRecord::Schema[7.2].define(version: 2025_05_05_151422) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "btree_gin"
   enable_extension "fuzzystrmatch"
@@ -18,6 +18,7 @@ ActiveRecord::Schema[7.2].define(version: 2025_03_21_032314) do
   enable_extension "pg_trgm"
   enable_extension "plpgsql"
   enable_extension "postgis"
+  enable_extension "postgis_raster"
   enable_extension "tablefunc"
 
   create_table "active_storage_attachments", force: :cascade do |t|
@@ -722,8 +723,8 @@ ActiveRecord::Schema[7.2].define(version: 2025_03_21_032314) do
   end
 
   create_table "documentation", id: :serial, force: :cascade do |t|
-    t.string "documentation_object_type", null: false
     t.integer "documentation_object_id", null: false
+    t.string "documentation_object_type", null: false
     t.integer "document_id", null: false
     t.integer "project_id", null: false
     t.integer "created_by_id", null: false
@@ -741,7 +742,7 @@ ActiveRecord::Schema[7.2].define(version: 2025_03_21_032314) do
   create_table "documents", id: :serial, force: :cascade do |t|
     t.string "document_file_file_name", null: false
     t.string "document_file_content_type", null: false
-    t.bigint "document_file_file_size", null: false
+    t.integer "document_file_file_size", null: false
     t.datetime "document_file_updated_at", precision: nil, null: false
     t.integer "project_id", null: false
     t.integer "created_by_id", null: false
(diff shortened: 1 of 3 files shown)


LocoDelAssembly:
Looks this index might cause some issues: https://github.com/SpeciesFileGroup/taxonworks/actions/runs/14840581728/job/41662316989

```
   1) Gis::GeoJSON outputting GeoJSON 'Feature's geographic_item that the geographic_item type 'geometry_collection' produce GeoJSON
     Failure/Error:
       object = GeographicItem.create!(geography: 'GEOMETRYCOLLECTION( POLYGON((0.0 0.0 0.0, ' \
                                                             '10.0 0.0 0.0, 10.0 10.0 0.0, ' \
                                                             '0.0 10.0 0.0, 0.0 0.0 0.0)), POINT(10 10 0)) ')
     
     ActiveRecord::StatementInvalid:
       PG::InternalError: ERROR:  ST_Centroid(geography) unhandled geography type
     # ./spec/lib/gis/geo_json_spec.rb:169:in `block (4 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # PG::InternalError:
     #   ERROR:  ST_Centroid(geography) unhandled geography type
     #   ./spec/lib/gis/geo_json_spec.rb:169:in `block (4 levels) in <top (required)>'
```

mjy:
We may have to just remove it.  I'm not sure what benefit we were getting from it, perhaps from distance calculations.

kleintom:
ST_Centroid is supposed to work on geography since 2.4 per https://postgis.net/docs/ST_Centroid.html, shrug. I'm not surprised centroid on geometry is faster than centroid on geography, but it's over 100x faster! so probably better to just operate on geometry anyway.

mjy:
Is the roblem specifically with GeometryCollection or it's equivalent in Geographaphy?

kleintom:
> Is the problem specifically with GeometryCollection or it's equivalent in Geographaphy?

That spec was added 7 years ago
https://github.com/SpeciesFileGroup/taxonworks/blame/8ae209916c0a02bcdd35e466485eed2fc57a8a9a/spec/lib/gis/geo_json_spec.rb#L172-L185

and the existing centroid index was added in 2021 (db/migrate/20210617203831_add_centroid_index_to_geographic_item.rb) so my guess would be that the issue is with doing st_centroid on a geometry_collection _as geography_, not with geometry_collection in general.

They give an image example for centroid of a geometry_collection and explain how it works for geometry_collection at https://postgis.net/docs/ST_Centroid.html.

Let me know if you'd like me to dig into the source to make sure, I'm happy to do so.

kleintom:
Yah, looks like just geometry_collection is unsupported for St_Centroid of a geography:
https://git.osgeo.org/gitea/postgis/postgis/src/commit/935e23f3c42b24ed9074e77e5473ae18ab8eedbe/postgis/geography_centroid.c#L161-L164

and everything is supported for St_Centroid of a geometry:
pg calls GEOS (I think...):
https://git.osgeo.org/gitea/postgis/postgis/src/commit/935e23f3c42b24ed9074e77e5473ae18ab8eedbe/liblwgeom/lwgeom_geos.c#L936-L956
and GEOS supports centroids of geometry collections (I think):
https://github.com/libgeos/geos/blob/6f4a7396481ab28dd240cc7b1511949f1892ac2a/src/algorithm/Centroid.cpp#L71-L93

LocoDelAssembly:
Given we don't seem to have not even a single instance of geometry_collection in production, could we perhaps just drop support for it and instead upon arrival transform those into multiple GeographicItem instances? (Only possible if we don't have models with `has_one` relationship to GeographicItem that geometry_collection would be a reasonable type to use)

mjy:
> Given we don't seem to have not even a single instance of geometry_collection in production, could we perhaps just drop support for it 

We can do this until there is compelling demand- there might be down the road with Gaz, but I'm happy to drop.

kleintom:
My initial reaction is to prefer to keep geometry_collection, but I agree there are pluses and minuses.

Another possibility: keep the has_one geographic_item on Gazetteer but reject (frontend and/or backend) trying to save a Gaz that woud create a GeometryCollection. Personally I think I'd prefer that (although I'd kind of like to see something in the current code broken before we disallow GeometryCollection).

> Only possible if we don't have models with `has_one` relationship to GeographicItem that geometry_collection would be a reasonable type to use

Gazetteer may currently qualify: https://github.com/SpeciesFileGroup/taxonworks/blob/8ae209916c0a02bcdd35e466485eed2fc57a8a9a/app/models/gazetteer.rb#L40

If someone creates a new Gaz by drawing a polygon and a line, that creates a geometry collection (and of course other combinations would as well). You know better than me if that would be a reasonable thing to want to do.

> Given we don't seem to have not even a single instance of geometry_collection in production
Is there currently any way for a user to create one? The georef leaflet UI doesn't support it, but I guess you could create one by entering WKT for a georef (how often is wkt on georef used?).

kleintom:
Are you thinking of http://localhost:3000/geographic_areas/by_lat_long?latitude=12&longitude=14 ?  It uses this: https://github.com/SpeciesFileGroup/taxonworks/blob/9b1bddde0204a50598711b0510c1187a38cf6929/app/models/geographic_area.rb#L281-L300 which doesn't use centroids.

kleintom:
> It uses this

(Note to self: that looks overly complicated (that's on me), and not sure why we're checking polygons on GA GIs.)

db/schema.rb
```diff
@@ -10,14 +10,15 @@
...
-ActiveRecord::Schema[7.2].define(version: 2025_03_21_032314) do
+ActiveRecord::Schema[7.2].define(version: 2025_05_05_151422) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "btree_gin"
   enable_extension "fuzzystrmatch"
   enable_extension "hstore"
   enable_extension "pg_trgm"
   enable_extension "plpgsql"
   enable_extension "postgis"
+  enable_extension "postgis_raster"
```
kleintom, line 21: haha, look who's back

PR #4346, opened 2025-05-13
Filter roll call

kleintom:
See what you think, I'm open to other approaches or holding off on this. The tldr: ~2x speed up on (some) spatial filters.

There are two patterns in use for checking if a query's params generate any facets other than the default project_id_facet: the `# Turn project_id off and check for a truly empty query` pattern and the `Filter#only_project?` call.

The issue with both of those patterns is that they both create every facet to see which are not nil. (Some of the) spatial filters do almost all of their work *before* returning their facet scope, so in effect when the above pattern is used it doubles the time spent on spatial queries.
See for example https://github.com/SpeciesFileGroup/taxonworks/blob/e1ff19a25bb138fb3cdac0ad9cc551ba38bed2fd/lib/queries/collecting_event/filter.rb#L253-L260

This pr adds a `@roll_call` member variable to Filter that lets filter facets know they're just being polled to see if they're nil or not, allowing them to just return a scope.none value instead of doing all of their normal work. For facets that don't do much work to create their scope, they just ignore this and do what they've always done. In the spatial scopes though this make the facet run ~2x as fast.

Filters that benefit (for spatial facets): CO, CE, AD, Otu, BA, FO
Tasks that benefit (on spatial filters): Batch update Confidence and Protocol Relationships on filters, since they check for `only_project?` on their filter_query before running.

A couple timing examples:
* Filter Biological Associations on the NZ Stewart Island Gaz (with 38,055 exterior vertices) went from 4.39 minutes to 2m15s.
* Filter Collection Objects on the Fiordland Gaz (with 99,556 exterior vertices) went from 2.10 to 1.02m

### DIFF
--- a/lib/queries/asserted_distribution/filter.rb
+++ b/lib/queries/asserted_distribution/filter.rb
@@ -192,6 +192,7 @@ module Queries
       # Shape is a Hash in GeoJSON format
       def geo_json_facet
         return nil if geo_json.nil?
+        return ::AssertedDistribution.none if roll_call
 
         i = spatial_query
         return nil if i.nil?
@@ -213,7 +214,6 @@ module Queries
             ::AssertedDistribution
               .where( asserted_distribution_shape: gz_ids )
               .where( asserted_distribution_shape_type: 'Gazetteer')
-
           )
 
       end
@@ -263,6 +263,7 @@ module Queries
         return nil if geo_shape_id.empty? || geo_shape_type.empty? ||
           # TODO: this should raise an error(?)
           geo_shape_id.length != geo_shape_type.length
+        return ::AssertedDistribution.none if roll_call
 
         geographic_area_shapes, gazetteer_shapes = shapes_for_geo_mode
 
--- a/lib/queries/asserted_distribution/filter.rb
+++ b/lib/queries/asserted_distribution/filter.rb
@@ -277,11 +277,9 @@ module Queries
 
         if geo_mode == true # spatial
           i = ::Queries.union(::GeographicItem, [a,b])
-          wkt_shape =
-            ::Queries::GeographicItem.st_union(i)
-              .to_a.first['st_union'].to_s
+          u = ::Queries::GeographicItem.st_union_text(i).to_a.first
 
-          return from_wkt(wkt_shape)
+          return from_wkt(u['st_astext'])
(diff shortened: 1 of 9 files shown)


kleintom:
I think there's room for more improvement here: https://github.com/SpeciesFileGroup/taxonworks/blob/e1ff19a25bb138fb3cdac0ad9cc551ba38bed2fd/lib/queries/collecting_event/filter.rb#L333-L337 I'll check tomorrow

kleintom:
> I think there's room for more improvement here

It turned out this is where the vast majority of the slowness was hiding:
Compare https://github.com/SpeciesFileGroup/taxonworks/blob/4114378452175a5670842f1315ce3b4c7dff64c2/lib/queries/collecting_event/filter.rb#L253-L259
and
https://github.com/SpeciesFileGroup/taxonworks/blob/4114378452175a5670842f1315ce3b4c7dff64c2/lib/queries/collecting_event/filter.rb#L334-L338
We compute a union of shapes, instantiate it with `to_a` - which converts the union shape to RGeo - immediately convert the rgeo shape to a string, pass that string to `from_wkt` which immediately parses the string into RGeo so it can determine its geometry type, then ditch the rgeo shape again. For shapes with lots of vertices, that was a major time sink. (The situation was the same pre-gaz, though the code changed a little.)

New numbers, with both the roll_call fix and the wkt fix:
Filter CO on Fiordland Gaz (99k external vertices): 2.4m on dev, now 1.6 seconds and then 1.6s again because we do a separate repeated query for data attributes, so ~3s total (!!)
Filter CO on Northern America GA (90k external vertices): 1.8m on dev, now 3.7+2.2=5.9s

Filter BA on Stewart Island Gaz: 4.6m on dev, now 6s
Filter BA on Northern America: 11m on dev, now 43s

(My interpretation is that Fiordland has more vertices than Northern America, but Northern America covers a much much larger area, so many more subshapes intercept it: the spatial processing on NA is a larger portion of the total run time compared to Fiordland.)

kleintom:
D'oh, I had thought there were more of these and searched, but I had put my search term in the wrong box :see_no_evil:

mjy:
Sounds great, and doesn't suprise me that this existed.  Will merge on green.

Issue #4333, opened 2025-05-07
[Bug]: 500 on Sound metadata whose ActiveStorage has been removed


kleintom:
### Steps to reproduce the bug

```bash
Discussion in followup comment.
1. Create some sounds.
2. As a TW instance maintainer, delete your ActiveStorage folder in your TW instance directory.
3. Now do anything that tries to load any of those sounds, like try to add a conveyance which tries to get
   metadata on recent sounds.
Get a 500.
```

### Expected behavior

I'm not sure what expected behavior should be... This is really an admin-level problem I think.

### Environment

Development (native)

kleintom:
I think this is a bug in some form, probably not the form I put it in.
I believe what actually happened in my case is I downloaded a project that had sounds in it, then I imported that project into my local TW db, and (I'm guessing) the sounds didn't come with the project, or at least weren't reloaded in their original ActiveStorage locations, and so now anytime you try to touch those sounds in the re-imported project you get an error.
I think we're in the same situation with images right? And we just tolerate images that don't exist at their expected location - presumably we should do the same here?

kleintom:
@mjy I'm not sure if the previous imported-project situation applies to your demo experience (I've forgotten what you said that 500 said).

LocoDelAssembly:
We might have forgotten to properly update project download? The new models likely have been added to the export automatically, but the tables that active_storage added to the system perhaps not? (And of top of that, perhaps it is not resilient to missing files as you mention)

LocoDelAssembly:
Tables are included. So we need to guard against missing files only.

mjy:
> And we just tolerate images that don't exist at their expected location - presumably we should do the same here?

Yes. this is the case.  This year I'd like to start looking at functions that offline-sync binary resources as we've discussed internally (but not here).

PR #4347, opened 2025-05-14
Tolerate missing Sound file #4333

kleintom:
This can happen when you import an exported project (or delete your active storage folder).

### DIFF
--- a/app/helpers/conveyances_helper.rb
+++ b/app/helpers/conveyances_helper.rb
@@ -15,11 +15,23 @@ module ConveyancesHelper
       object.conveyances.collect{|c|
         content_tag(:div) do
           safe_join([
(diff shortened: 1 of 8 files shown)


mjy:
@kleintom In general we don't want to tolerate bad data in the UI. Thinking to advance the concept we have for Images to accomodate this idea, can we scope  this handling to a development or sandbox environment specifically, but continue to raise nastily in production?

kleintom:
> can we scope this handling to a development or sandbox environment specifically, but continue to raise nastily in production?

I wasn't seeing how it's done for images, is this okay for Sounds? I'm checking `Rails.env.production?`, though I couldn't run with that locally because the pg adapter doesn't like it.

PR #4357, opened 2025-05-20
Turn off pp_sql on to_sql, except for logging

kleintom:


kleintom:
I don't think the failure here is to_sql related, there's no exception (the spec query returns results, just not the expected results). Plus it passes for me locally. Re-try?

PR #4358, opened 2025-05-20
Copy georef error attrs on Set Georeference filter CE operation

kleintom:
Note also this fixes copying only the selected georefs from the UI:
image

There are other georef attributes that could be copied as well which I left off, e.g., `api_request`, `is_undefined_z`, `attribute_year_georefenced`, etc.

Also we still don't display error regions in the Set Georeferences map (I didn't look into that).

### DIFF
--- a/app/controllers/collecting_events_controller.rb
+++ b/app/controllers/collecting_events_controller.rb
@@ -295,7 +295,8 @@ class CollectingEventsController < ApplicationController
       :group, :member, :formation, :lithology, :max_ma, :min_ma,
       :end_date_year, :verbatim_habitat, :field_notes, :verbatim_datum,
       :verbatim_elevation, :meta_prioritize_geographic_area,
-      georeferences_attributes: [:type, :geographic_item_id], # batch add only use right now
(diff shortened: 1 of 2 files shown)


jlpereira (approved):

mjy:
> Also we still don't display error regions in the Set Georeferences map (I didn't look into that).

This is known.  It's OK for now.

We likely should copy these attributes:
```
 geographic_item_id       | integer                     |           | not null | 
 error_radius             | numeric                     |           |          | 
 error_depth              | numeric                     |           |          | 
 error_geographic_item_id | integer                     |           |          | 
 type                     | character varying           |           | not null | 
 is_public                | boolean                     |           | not null | false
 api_request              | character varying           |           |          | 
 year_georeferenced       | integer                     |           |          | 
 month_georeferenced      | integer                     |           |          | 
 day_georeferenced        | integer                     |           |          | 
```

These may have no meaning, conceptual only, and can likely be removed at some point until we need them:

 is_undefined_z           | boolean                     |           |          | 
 is_median_z              | boolean                     |           |          |

jlpereira:
Thanks @kleintom

PR #4376, opened 2025-05-23
Switch to pp_sql-logging-on only in dev console; never in logs

kleintom:
I discovered that pp_sql is active during background jobs because I had a job that timed out on

```
Could not log "sql.active_record" event. AnbtSql::FormatterException: execution expired (Delayed::Worker.max_run_time is only 14400 seconds) ["/home/klein/.rbenv/versions/3.4.4/lib/ruby/gems/3.4.0/gems/anbt-sql-formatter-0.1.2/lib/anbt-sql-formatter/coarse-tokenizer.rb:141:in 'String#[]'", "/home/klein/.rbenv/versions/3.4.4/lib/ruby/gems/3.4.0/gems/anbt-sql-formatter-0.1.2/lib/anbt-sql-formatter/coarse-tokenizer.rb:141:in 'CoarseTokenizer#shift_to_buf'", "/home/klein/.rbenv/versions/3.4.4/lib/ruby/gems/3.4.0/gems/anbt-sql-formatter-0.1.2/lib/anbt-sql-formatter/coarse-tokenizer.rb:129:in 'CoarseTokenizer#tokenize'", "/home/klein/.rbenv/versions/3.4.4/lib/ruby/gems/3.4.0/gems/anbt-sql-formatter-0.1.2/lib/anbt-sql-formatter/parser.rb:283:in 'AnbtSql::Parser#parse'", "/home/klein/.rbenv/versions/3.4.4/lib/ruby/gems/3.4.0/gems/anbt-sql-formatter-0.1.2/lib/anbt-sql-formatter/formatter.rb:55:in 'AnbtSql::Formatter#format'", "/home/klein/.rbenv/versions/3.4.4/lib/ruby/gems/3.4.0/gems/pp_sql-2.1.0/lib/pp_sql.rb:56:in 'PpSql::LogSubscriberPrettyPrint#sql'", "/home/klein/.rbenv/versions/3.4.4/lib/ruby/gems/3.4.0/gems/activesupport-7.2.2.1/lib/active_support/subscriber.rb:138:in 'ActiveSupport::Subscriber#call'" ...
```

AnbtSql is pp_sql's parser(?) - the background job was processing large shapes, like 1,000,000+ vertex shapes, it looks like it timed out tokenizing one of those (it never got past the first shape). Without pp the job completed.

Another similar experience I had the other day was filtering COs in Brazil, the query took about 6 minutes, I think because it was logging Brazil's wkt shape, probably multiple times. Without pp that query takes about a second.

### DIFF
--- a/config/initializers/vendor/pp_sql.rb
+++ b/config/initializers/vendor/pp_sql.rb
@@ -1,4 +1,17 @@
 if Rails.env.development? && defined?(Rails::Console)
   require 'pp_sql'
-  PpSql.add_rails_logger_formatting = true
+
+  if ENV['RAILS_PP_SQL_CONSOLE_LOGGING']
+    PpSql.add_rails_logger_formatting = true
+  end
+
+  def pp(bool)
+    if bool == true || bool == false
+      puts "PpSql logging is now set to #{bool}."
+    else
+      puts "Boolean input value required."
+    end
+
+    PpSql.add_rails_logger_formatting = bool
+  end
 end
--- a/config/initializers/vendor/pp_sql.rb
+++ b/config/initializers/vendor/pp_sql.rb
@@ -1,7 +1,7 @@
 if Rails.env.development? && defined?(Rails::Console)
   require 'pp_sql'
 
-  if ENV['RAILS_PP_SQL_CONSOLE_LOGGING']
+  if ENV['TW_PP_SQL_CONSOLE_LOGGING']
     PpSql.add_rails_logger_formatting = true
   end
 
--- a/config/initializers/vendor/pp_sql.rb
+++ b/config/initializers/vendor/pp_sql.rb
@@ -1,0 +1,4 @@
+if Rails.env.development? && defined?(Rails::Console)
+  require 'pp_sql'
+  PpSql.add_rails_logger_formatting = true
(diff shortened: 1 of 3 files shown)


kleintom:
You can of course still run into the same problem in the console (pp taking a long time to parse wkt shapes); with this patch to avoid that you'd have to manually turn pp_sql logging off in the console with `PpSql.add_rails_logger_formatting = false`. I tested that that works, you can turn it on and off manually that way.

mjy:
@kleintom Perhaps we need to eliminate it altogether and turn it on with an ENV variable as needed in development OR a console command?

config/initializers/vendor/pp_sql.rb
```diff
@@ -0,0 +1,17 @@
+if Rails.env.development? && defined?(Rails::Console)
+  require 'pp_sql'
+
+  if ENV['RAILS_PP_SQL_CONSOLE_LOGGING']
```
mjy, line 4: Maybe 'TW_<something>' ?

kleintom: I thought of that too after I pushed of course, thanks for catching it.

PR #4380, opened 2025-05-27
Remove old geographic_item shape columns

kleintom:


kleintom:
I had two different errors (on two migrations on dbs in different states) on creating a georef after this migration that disappeared after I reloaded the page - is it possible production users could run into that, or will they have to re-login or something after an update?

PR #4384, opened 2025-05-29
Fix duplicate filter results on match_identifiers

kleintom:
This fixes the duplicate results but now, in the case where there *are* duplicate results, query.pluck(:id) is broken (i.e. returns unexpected results) because the solution to the duplicate results was to add .select('DISTINCT ...'), whereas .pluck (from what I'm told) works by undoing the existing select and adding a new one.

Could do
    .select("#{table.name}.*, sid.s")
    .group("#{table.name}.id", 'sid.s')
    .order('sid.s')
    but that "breaks" .size (and .count) on the result set

Could we instead just add .distinct to the end of the query? Maybe?? The query has a select across a join because we need to preserve order of the results by match array order on identifiers. AFAICT .distinct in that case can be flaky in rails, in a way I don't understand well enough to use it.

On a similar note - which would be a better way to fix this bug if possible - see https://github.com/SpeciesFileGroup/taxonworks/blob/e1d952e23fd34f93fc620b0b7ecd2733196e2f12/lib/queries/query/filter.rb#L838-L839

Longer-winded and perhaps slower, but also perhaps most reliable and least gotcha-y: 
deduped = target_query.joins(ids that match the input match_ids).distinct 
ordered = target_query.joins(ids that match the input match_ids, along with the index of the id that matches).sort(index of id that matches).where(id: deduped)


kleintom:
I've marked as draft to solicit feedback here. As I note in my commit message, several of the solutions I tried "worked" in the sense that the result set was correct, but returned a query with some unexpected behavior, e.g. pluck on it might not return what you expect, or size might not return what you expect.

Can anyone comment on the extent to which this is expected/tolerated/... in the TW codebase? On one hand I think in terms of wanting to return a black-box scope that can be chained in whatever way the receiver needs ... in practice there often seem to be caveats to that. Comments?

Right now the 'pluck sometimes broken' solution is posted, but I'm actually leaning toward the last solution mentioned, which seems (maybe?) slower but less prone to surprising gotchas.

kleintom:
I should have written this down at the start:
STR
1) Add two identifiers to a CO, both with the same identifier value and namespace, but one a Catalog Number and one a Record Number.
2) Add your identifer, e.g. `ASDF 1`, to the match-identifiers facet of Filter COs. Use the `Sort as listed` option.
3) See the filter result includes a duplicate, one each of your CO being returned for the catatlog number and the record number.

The problem is that with `Sort as listed` checked, we try to return results in batches ordered in the same order as the identifiers list. In this example that means we match 'ASDF 1' against all Identifiers having that cached value, and then join that to COs, which in this case gives two COs.

So that matching result needs to be deduplicated.

kleintom:
Here's another example, from antweb: https://sfg.taxonworks.org/tasks/collecting_events/browse?collecting_event_id=562666
image
From the UI it appears that there are 3 identical Identifiers on this CE, but if you check the Network tab of devtools, you see:
* 1 has `identifier` value `22559` and non-virtual namespace id 917 with verbatim_short_name `BLF`
* 1 has `identifier` value `BLF 22559` and virtual namespace id 1014
* 1 has `identifier` value `BLF 22559` and virtual namespace id 1016

I think Brian has explained elsewhere that this was the result of multiple imports? These are perfectly valid data as far as I can tell, but if the verbatim short name 'BLF 22559' is entered in the Identifier facet of Collecting Event, then we're going to find that CE three times (one for each of the identifiers whose short name matches `BLF 22559`) and need to just return that result once instead of three times.

kleintom:
Should be ready to go but marking as draft for now since we'll need to "fix" some production identifiers before this can land.

kleintom:
> * Local identifiers must be unique on cached/type/identifier_object_type (Global and Unknown already have stronger unique on cached==identifier)
> 
> * Over all identifier types, cached is unique per object (can only have one identifier on CO id=1 with cached value 'ASDF 1234').

A couple of possible issues with these (examples from AntWeb):
* When importing from DwC, COs get a `Identifier::Local::Import::Dwc` of the form `occurrenceID:<occurrenceID>`, where `<occurrenceID>` is the value of that column from the import file. If those IDs come from database table ids, then we could easily have multiple imports of different datasets producing COs with the same `Identifier::Local::Import::Dwc` values, which would not be allowed by the above validations. (If needed, I guess the fix would be to add some unique-to-project namespace prefix to your `occurrenceID` column values.)
https://sfg.taxonworks.org/tasks/collection_objects/filter?per=50&identifier=occurrenceID%3ACASENT0151595-D01&extend%5B%5D=dwc_occurrence&extend%5B%5D=repository&extend%5B%5D=current_repository&extend%5B%5D=collecting_event&extend%5B%5D=taxon_determinations&extend%5B%5D=identifiers&extend%5B%5D=container_item&extend%5B%5D=container&exclude%5B%5D=object_labels&page=1&paginate=true
* they would not allow a given CE to have the same record number and catalog number, e.g.: https://sfg.taxonworks.org/collection_objects/2066290 My understanding is that ideally that shouldn't occur with best practice, but sometimes it does? Do we really want to disallow it?

Issue #4390, opened 2025-05-30
Ignore paging/per in venn queries


kleintom:
### Feature or enhancement

From Brian Fisher on AntWeb, discussing in chat:
STR:
A query: https://sfg.taxonworks.org/tasks/collecting_events/filter?per=50&data_attribute_wildcard_pair%5B%5D=3625%3AMadagascar&extend%5B%5D=roles&page=1&paginate=true
B query: https://sfg.taxonworks.org/tasks/collecting_events/filter?per=50&geo_shape_id%5B%5D=2421&geo_shape_type%5B%5D=Gazetteer&geo_mode=true&extend%5B%5D=roles&page=1&paginate=true

What I (think I) was seeing:

> the DA query is your A query (with 50,624 results), the geo query is your B query (with 49980 results). If I load your A query, paste your B query in its venn facet, and select Exclude as the option, I get 50574 results. 50,624 - 50,574 = 50, so the venn is removing 50 results from the B query. I thinks that's becasue the B query has per=50 (and maybe because of paginate=true and/or page=1, I'm not so familiar with all of that). If I remove per=50 and paginate=true and page=1 (those last two may not need to be removed) from the B query and rerun the venn I get 669 records. 

Here's the venn query with `per` on B: https://sfg.taxonworks.org/tasks/collecting_events/filter?per=50&data_attribute_wildcard_pair%5B%5D=3625%3AMadagascar&page=1&venn=https%3A%2F%2Fsfg.taxonworks.org%2Ftasks%2Fcollecting_events%2Ffilter%3Fper%3D50%26geo_shape_id%25255B%25255D%3D2421%26geo_shape_type%25255B%25255D%3DGazetteer%26geo_mode%3Dtrue%26page%3D1%26extend%25255B%25255D%3Droles%26paginate%3Dtrue&extend%5B%5D=roles&paginate=true

[...]

PR #4392, opened 2025-06-02
Remove pagination params on venn_query #4390

kleintom:
This removes them on the :b query; the :a query hasn't had them applied when venn processing occurs; the pagination params of the :a query are added to the venn'ed query after venn processing.

### DIFF
--- a/app/javascript/vue/components/Filter/Facets/shared/FacetDiffMode.vue
+++ b/app/javascript/vue/components/Filter/Facets/shared/FacetDiffMode.vue
@@ -46,9 +46,9 @@
         <input
           type="checkbox"
           :value="true"
-          v-model="params.venn_b_one_page"
+          v-model="params.venn_ignore_pagination"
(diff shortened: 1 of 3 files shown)


mjy:
Discussion resolved to an option to let user remove outermost pagination. If server side then `venn_ignore_pagination` param perhaps, I think client side processing likely not the approach.

kleintom:
I added a checkbox for paging:
image

Further suggestions on wording/default value/etc welcome.

mjy:
@kleintom Can you use the Orange help and/or tooltips to expand the help to shorten the UI load?  I.e. replace the label with "Ignore pagination ", and put that as a help element.

kleintom:
Added orange help - cool! I hadn't realized how easy that is, I really like it.
Also flipped the new venn param to `venn_ignore_paging` and made the default false as discussed on gitter.
image

Issue #4188, opened 2025-02-05
Filter CO- Batch Update- Catalog Number Namespace


lvhart2:
### Feature or enhancement

In Filter CO task, I would like to be able to batch update the Catalog Number Namespace. Currently, this is not possible through the Radial CO button options. 
In INHS Insect Collection project, I have a list of 324 West Virginia Plecoptera CO's that need to be batch updated from **INHS namespace** to **INHS Insect Collection namespace**. It takes too long to edit these manually. 

### Location

Filter collection objects

### Screenshot, napkin sketch of interface, or conceptual description

![Image](https://github.com/user-attachments/assets/05e61e78-c152-4f35-aa07-6d2348039ee8)

### Your role

Graduate Research Assistant

PR #4396, opened 2025-06-03
Make BatchByFilterScope controller concern DataControllerConfiguratio…

kleintom:
…n #4188

It's just a little awkward here since `DataControllerConfiguration::ProjectDataControllerConfiguration` also includes `DataControllerConfiguration`, but Ruby protects us there from double include, and I think in general this is the right thing for BatchByFilterScope to do.


mjy:
@kleintom Going to merge this, but see possible improvement re macro version downstream.

PR #4397, opened 2025-06-03
Fix cached_numeric_identifier creation in the virtual-namespace case

kleintom:
When the namespace of a local identifier is virtual, the `identifier` field may be saved prefixed with its namespace, e.g. `identifier == 'ASDF1234'`.

The migration here took about 3 minutes on 3i for me.

### DIFF
--- a/db/migrate/20250602194448_update_local_identifiers_cached_numeric.rb
+++ b/db/migrate/20250602194448_update_local_identifiers_cached_numeric.rb
@@ -1,0 +1,11 @@
+class UpdateLocalIdentifiersCachedNumeric < ActiveRecord::Migration[7.2]
+  def change
+    Identifier::Local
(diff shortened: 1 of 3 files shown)


db/migrate/20250602194448_update_local_identifiers_cached_numeric.rb
```diff
@@ -0,0 +1,8 @@
+class UpdateLocalIdentifiersCachedNumeric < ActiveRecord::Migration[7.2]
+  def change
+    Identifier::Local.find_each do |i|
```
mjy: @kleintom I think we can be more efficient here?  Does this only apply to a subclass of identifiers, or identifier with a particular namespace?  I.e. `Idenfiier...join(:namespace).where(namespace: { is_virtual: true})...`

PR #4409, opened 2025-06-06
Batch Data Attribute refactor

kleintom:
Ready for review but this may need some discussion, so marking as draft for now.
Changes:
* Left and right annotators offer the same options
* The options have changed from combined-create-update (left) and create (right) to Add, Remove, Replace - the same options offered for Confidences and Protocols.
* Operations are now using batch_by_filter_scope


mjy:
@kleintom can you screenshot the UI now if you get a chance?

kleintom:
image
image

On replace you can only replaces value, not predicate and value.

PR #4412, opened 2025-06-06
Taxon Name Relationships filter

kleintom:
image

* You can send results between TN and TNR, but in both directions it links on both sides of relationships
* Almost surely there will be other options to make this more useful, I just added some basic facets that seemed useful
* @jlpereira I think this is the first filter with no left side annotators (it only has right side Citations and Notes), so there's a black hole where the left annotator would appear: 
image
I wasn't sure what to do there (just add another filter prop to hide that annotator?)


mjy:
Very nice @kleintom looks great.  Let's make sure the CHANGELOG also mentions the generic verifiers facet, that can go in other filters too.

PR #4416, opened 2025-06-08
Make more attribute filter concern queries distinct

kleintom:


lib/queries/concerns/citations.rb
```diff
@@ -75,7 +75,7 @@ def origin_citation_facet
     return nil if origin_citation.nil?
 
     if origin_citation
-      referenced_klass.joins(:origin_citation)
+      referenced_klass.joins(:origin_citation).distinct
```
kleintom: This one probably isn't needed since a given citation_object can only have one origin citation.

kleintom: Removed.

PR #4421, opened 2025-06-09
4410 Create Asserted Distributions for Otu filter result

kleintom:
@jlpereira @mjy Ready for feedback/review here. As my commits reflect, it took me a minute to move away from trying to do this with QueryBatchRequest - I don't think the creates we're doing here fit that pattern.

Client side is still using UpdateBatch and PreviewBatch, we may want to adjust some of the wording that provides from 'update'-centric to 'create'-centric, e.g. the submit button is 'Update' and the confirm string is 'CHANGE'.


app/controllers/asserted_distributions_controller.rb
```diff
@@ -126,6 +126,20 @@ def batch_update
     end
   end
 
+  def batch_template_create
+    if r = AssertedDistribution.batch_template_create(
```
mjy, line 130: I like the "template" convention.

app/models/asserted_distribution.rb
```diff
@@ -175,6 +175,73 @@ def self.batch_update(params)
     query_batch_update(request)
   end
 
+  def self.batch_template_create(params)
```
mjy, line 178: I feel ikely ultimately we should move these batch related code to their own <something>Batch modules., then use constants for the cutoff values etc. For later...

Issue #4425, opened 2025-06-10
[Bug]: Errors on People Filter to linker radial tasks (people vs. person confusion)


kleintom:
### Steps to reproduce the bug

```bash
1. Go to Filter People, check some results.
2. If you use the linker to send to Summary or Synchronize you get a 500, I think because the query
sent uses `people` instead of `person`.
E.g. the summary link is `tasks/people/summary?person_id[]=1&person_id[]=2`
and the synchronize link is `field_synchronize?people_query[person_id][]=1&people_query[person_id][]=2`

Oddly, the Multi-Update attributes link works with people: multi_update?people_query[person_id][]=1&people_query[person_id][]=2 (it's fetching data using `Person` though, somehow).
3. 
...
```

### Environment

Development (native)

kleintom:
@jlpereira People.js here seems suspicious? https://github.com/SpeciesFileGroup/taxonworks/tree/development/app/javascript/vue/components/radials/linker/links All the others are singular.

Issue #4007, opened 2024-08-14
Send Filtered Results to Project Vocabulary


tmcelrath:
### Feature or enhancement

When I am looking at a set of results from a filter,
Then I want to send that via radial linker to project vocabulary,
Then I can look in the set of results for groups of things to act on.

### Location

Filters/Radials/Project Vocabulary

### Your role

Collections manager

PR #4426, opened 2025-06-10
Filter results to Project Vocabulary

kleintom:
This is https://github.com/SpeciesFileGroup/taxonworks/issues/4007

Filter people --> Project Vocab is currently affected by #4425 
I also noted an issue in setParam.js in the code, which currently uses a workaround.


lib/vocabulary.rb
```diff
@@ -2,27 +2,33 @@ module Vocabulary
...
+    if klass.new.attributes.symbolize_keys.keys.include?(:project_id)
+      words = words.where(project_id:)
+    end
+
     if [:string, :text].include?(t)
 
       c = "COUNT(\"#{attribute}\")" # double quote to handle things like `group`
 
-      words = words.where("\"#{attribute}\" like ?", "%#{begins_with}") if begins_with.presence
+      words = words.where("\"#{attribute}\" like ?", "#{begins_with}%") if begins_with.presence
```
mjy, line 31: Good catch.

PR #4427, opened 2025-06-10
Data attribute type validation

kleintom:


app/models/data_attribute.rb
```diff
@@ -37,7 +37,11 @@ class DataAttribute < ApplicationRecord
 
   belongs_to :predicate, foreign_key: 'controlled_vocabulary_term_id', class_name: 'Predicate', inverse_of: :internal_attributes
 
-  validates_presence_of :type, :value
+  validates :value, presence: true
+
+  validates :type, inclusion: {
```
mjy, line 42: I'm not sure we need this.  Try making `DataAttribute` an abstract class and see if it does the same thing.

kleintom: Ah, yah, I had forgotten about that. Unfortunately it seems that once you mark an ActiveRecord class as abstract, you can no longer use it for querying, data storage, etc. (per llm, which matches the errors I'm seeing):
image

mjy: OK, thanks for experimenting.

PR #4432, opened 2025-06-12
Add `include_common_names` param to otu api_autocomplete to return common names

kleintom:
Based on @neolefty's #4305; suggestions/feedback welcome!
image

### DIFF
--- a/app/helpers/otus_helper.rb
+++ b/app/helpers/otus_helper.rb
@@ -23,14 +23,51 @@ module OtusHelper
   end
 
   # Used exclusively in /api/v1/otus/autocomplete
-  def otu_extended_autocomplete_tag(target)
+  def otu_extended_autocomplete_tag(target, otu, term, include_common_names)
     if target.kind_of?(Otu)
-      otu_tag(target)
+      t = otu_tag(target)
     else # TaxonName
       a = [ tag.span( full_taxon_name_tag(target).html_safe, class: :otu_tag_taxon_name, title: target.id) ]
       a.push taxon_name_type_short_tag(target)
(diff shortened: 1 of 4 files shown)


kleintom:
Let me know (if this looks good) if we'd like to add a language parameter as well.

mjy:
@kleintom We can do the language param as a seperate issues. 

We should use concepts from https://guides.rubyonrails.org/i18n.html wherever possible.

At a global level:
* We should infer the language from the localization coming from the browser if possible to auto-set it.  
* We can overide browser localization with a param we "recognize" globally too.

app/helpers/otus_helper.rb
```diff
@@ -23,14 +23,51 @@ def otu_tag_elements(otu)
   end
 
   # Used exclusively in /api/v1/otus/autocomplete
-  def otu_extended_autocomplete_tag(target)
+  def otu_extended_autocomplete_tag(target, otu, term, include_common_names)
```
mjy: @kleintom I'd like to refactor this.  the last param should be an array of common names to include.  This lets us find all common names for a set of OTUs *before* we get here in a single query, removing a lot of n+1 queries.

lib/queries/otu/autocomplete.rb
```diff
@@ -160,14 +168,14 @@ def autocomplete_taxon_name_extended
 
         otus = scope_autocomplete(otus).includes(:taxon_name)
 
-        otus
+        return include_common_names ? otus.includes(:common_names) : otus
```
mjy: Maybe this is `eager_load` instead of `includes`.

PR #4455, opened 2025-06-23
4025 asserted distribution refactor

kleintom:
Updates:
* ADs are now polymorphic on object
  * Allowed object types: Otu, BiologicalAssociation, BiologicalAssociationsGraph, Conveyance (on Otu only), Depiction (on Otu only), Observation (on Otu only)
* API updates:
  * ADs now export an `asserted_distribution_object` that has a `type` attribute
  * BiologicalAssociationsGraph now has index and show endpoints
* Filter updates:
  * Filter on each of the AD object types that has a filter now sends to AD and vice versa
  * Each of those filters has spatial facets that include AD results
    * BA ADs match on BiologicalAssociationsGraphs, BiologicalAssociations, and subject/object of BiologicalAssociations
* Task updates:
  * New AD, Browse AD, Filter AD should fully support all new types
  * Batch load AD has not been tested (multiple ongoing issues there I suspect)
  * Basic Endemism task may need changes on shape matching (the previous Gaz version was not quite written correctly, though it's possible no one ever had an issue) - the current version matches Exact, I'm not sure if that's what's desired or not (I would have thought spatial)

TODO (maybe in followup):
* Add Filter <AD object type> -> Batch create AD for all AD object types, not just Otu
  * In the otu version this branch fixes an issue @proceps noticed: we shouldn't create a new AD if we can match an existing one (just create a new citation in that case)


kleintom:
This is ready for feedback now.
A couple more notes:
* I had to allow wrapping on VSwitch to display all the types allowed for AD objects now:
image
that may change other existing switches as well (José maybe you know how to fix or want to do something different here; feel free to push here if that's easiest)
* the pin icon in that selector still only keys off of pinboard Otus; I'm not sure what to do there now, maybe just remove it?
* Observation tags/labels now show the observation value and object in addition to the descriptor now that we need to be able to autocomplete observations:
image

mjy:
> Observation tags/labels now show the observation value and object in addition to the descriptor now that we need to be able to autocomplete observations:

We need a general rule in ARCHITECTURE about object_tag vs. autocomplete tags.  There is some info there, but will have to look to see if it covers this choice.

kleintom:
> > Observation tags/labels now show the observation value and object in addition to the descriptor now that we need to be able to autocomplete observations:
> 
> We need a general rule in ARCHITECTURE about object_tag vs. autocomplete tags. There is some info there, but will have to look to see if it covers this choice.

More documentation here might be helpful, but in this case, thinking back more, I think I made the change to tag and label because it wasn't making sense to specify an Observation just by its descriptor and value. It looks like the existing use case (before this pr) for observation_tag/label is in Associated objects of show pages for Matrix items:
image
New version is:
image

app/javascript/vue/components/ui/VSwitch.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="switch-radio">
+  <div :class="{ 'switch-radio': true, 'wrap-switch': options.length > 4 }">
```
jlpereira: I think this is a good idea when switch has many options. Perhaps this could be configurable by prop? Maybe in some scenarios we won't want to break the switch. So, we could add two props `wrap` that will allow wrap tabs and another one to set the max tabs before wrap, like you have now, for example: 0 = always wrap

or you can just use "wrap" for both things, if is passed as prop without assign value, then it will wrap everything, is you set `wrap="4"` then add wrap class when you have 4 or more options

PR #4463, opened 2025-07-01
Filter image attribution facets

kleintom:
@jlpereira Any feedback on the new facets before I merge?
image

### DIFF
--- a/app/javascript/vue/tasks/containers/new_container/components/ContainerSize.vue
+++ b/app/javascript/vue/tasks/containers/new_container/components/ContainerSize.vue
@@ -11,8 +11,10 @@
           type="number"
           :disabled="disabled"
           v-between-numbers="[1, 100]"
-          v-model="size[axis]"
-          @change="() => emit('change', true)"
+          @change="(event) => {
+            size[axis] = Number(event.target.value)
+            emit('change', true)
+          }"
         />
       </div>
     </div>


app/javascript/vue/components/Filter/Facets/shared/FacetYears.vue
```diff
@@ -0,0 +1,91 @@
...
+    <div class="margin-medium-bottom">
+      Or
+    </div>
+
+    <fieldset>
+      <legend>Range</legend>
+      <div class="field">
+        <label>After:</label>
+        <br />
+        <input
```
jlpereira, line 23: import { vBetweenNumbers } from '@/directives/betweenNumbers'

We got a directive for this, you can use `v-between-numbers="[minYear, maxYear]"`

It will force the user to set a number in the range

app/javascript/vue/components/Filter/Facets/shared/FacetYears.vue
```diff
@@ -0,0 +1,91 @@
...
+          :min="minYear"
+          :max="maxYear"
+          v-model="params[paramAfter]"
+        />
+      </div>
+      <div class="field">
+        <label>Prior to:</label>
+        <br />
+        <input
+          type="number"
```
jlpereira, line 33: Same this

jlpereira:
@kleintom Everything looks great. I left you a related comment about using a directive.

Just as an aside, I saw that in two facets you pass the parameter name as a prop. This seems fine to me if you think the filters will use another parameter for `license` or `source_id`. Otherwise, setting it internally will save you having to define it in each filter or modify it later if it changes. What you could do in this case is define a default value for those props and only set it if you plan to use a different one. These facets will probably only be used here and in one or two other filters, so I don't bother changing it.

PR #4469, opened 2025-07-08
4464 refactor spatial search to not go through wkt

kleintom:
* `filter.rb`s changed: AD, OTU, CE; all other spatial searches go through those in some way.

* I'm instantiating arrays of (potentially) tens of thousands of AD ids here, because it's (seems to be) faster than a monolithic query: is that bad?

* on dev this is ~8x faster than pre-pr code on Filter AD, but it does (currently) make more queries, so we'll see how it goes on production.

* I added `.includes` on the results of Filter AD - hopefully that will speed things up as well

* One note on the use of the new-ish `test_shape && geography` pattern: Russia and U.S.A. both have land close to either side of the anti-meridian, so their bounding boxes (for practical purposes) both essentially cover the entire northern hemisphere, so in those two cases at least adding the `&&` bounding box test makes things slower instead of faster (because the spatial index doesn't help much there).

### DIFF
--- a/app/models/geographic_item.rb
+++ b/app/models/geographic_item.rb
@@ -640,7 +640,8 @@ class GeographicItem < ApplicationRecord
     # !! CURRENTLY ASSUMES no geographic_item_id crosses the anti-meridian. This
     # is true for GA and AD shapes, but not necessarily for GEOREF shapes.
     def covered_by_geographic_items_sql(geographic_item_scope)
-      f = geographic_item_scope.select(
+      #f = geographic_item_scope.select(
+      f = ::GeographicItem.where(id: geographic_item_scope).select(
         '(ST_DUMP(geography::geometry)).geom AS the_geom'
       )
 
--- a/app/models/geographic_item.rb
+++ b/app/models/geographic_item.rb
@@ -639,9 +639,8 @@ class GeographicItem < ApplicationRecord
 
     # !! CURRENTLY ASSUMES no geographic_item_id crosses the anti-meridian. This
     # is true for GA and AD shapes, but not necessarily for GEOREF shapes.
-    def covered_by_geographic_items_sql(geographic_item_scope)
-      #f = geographic_item_scope.select(
-      f = ::GeographicItem.where(id: geographic_item_scope).select(
+    def covered_by_geographic_items_sql(geographic_item_id)
+      f = ::GeographicItem.where(id: geographic_item_id).select(
         '(ST_DUMP(geography::geometry)).geom AS the_geom'
       )
 
@@ -656,7 +655,17 @@ class GeographicItem < ApplicationRecord
         ') AS f'
       )
 
-      "geography && (#{geom}) AND " + subset_of_sql(geom).to_sql
+      precheck = "geography && (#{geom}) AND "
+      # Russia and U.S.A. both have land just either side of the anti-meridian,
+      # so doing a pre-bounding box check actually checks against basically
+      # every shape in the northern hemisphere (i.e. the spatial index helps
+      # very little), which turns out to be quite slow even though it's just
+      # rectangle intersection.
+      if geographic_item_id.include?(192) || geographic_item_id.include?(30449)
+        precheck = ''
+      end
+
+      precheck + subset_of_sql(geom).to_sql
     end
 
     #
--- a/app/models/geographic_item.rb
+++ b/app/models/geographic_item.rb
@@ -639,16 +639,7 @@ class GeographicItem < ApplicationRecord
 
     # !! CURRENTLY ASSUMES no geographic_item_id crosses the anti-meridian. This
     # is true for GA and AD shapes, but not necessarily for GEOREF shapes.
-    def covered_by_geographic_items_sql(geographic_item_scope, one_shape)
-      if one_shape
-        bounding_box_match =
-          "geography && (#{geographic_item_scope.select(:geography).to_sql})"
-
-        return bounding_box_match + ' AND ' +
-          # TODO change back to .arel
-          subset_of_sql(Arel.sql(geographic_item_scope.select(:geography).to_sql)).to_sql
-      end
-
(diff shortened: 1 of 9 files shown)


app/models/geographic_item.rb
```diff
@@ -655,14 +657,18 @@ def covered_by_geographic_items_sql(*geographic_item_id)
...
       )
 
-      precheck = "geography && (#{geom}) AND "
       # Russia and U.S.A. both have land just either side of the anti-meridian,
       # so doing a pre-bounding box check actually checks against basically
       # every shape in the northern hemisphere (i.e. the spatial index helps
       # very little), which turns out to be quite slow even though it's just
       # rectangle intersection.
-      if geographic_item_id.include?(192) || geographic_item_id.include?(30449)
+      skip_precheck = [192, 30449, 132534].any? do |x|
```
mjy: @kleintom This should be constantized at minimum.  Referencing IDs in code needs to be a (very) rare exception, as we want data to stand alone from the application ultimately.  I worry this just bandages complex items coming from Gaz too?

mjy: This feels like a tradeof situation where somethings are just going to be slower. We don't want to get hit in a algorithim whack-a-mole here.

lib/queries/asserted_distribution/filter.rb
```diff
@@ -151,42 +151,59 @@ def wkt_facet
...
+        k = ::GeographicArea.descendants_of(j).pluck(:id) # Add children that might not be caught because they don't have shapes
 
-        ::AssertedDistribution.from('(' + s + ') as asserted_distributions')
+        geographic_area_ids = (j + k).uniq.join(',')
+        if geographic_area_ids.empty?
+          return ::AssertedDistribution.none
+        end
+
+        ::AssertedDistribution
+          .where("asserted_distributions.asserted_distribution_shape_id IN (#{geographic_area_ids}) AND asserted_distributions.asserted_distribution_shape_type = 'GeographicArea'")
```
mjy: Prefer "native": `.where(asserted_distributions: {asserted_distribution_shape_id: geographic_area_ids, asserted_distribution_shape_type: 'GeographicArea'})`

lib/queries/asserted_distribution/filter.rb
```diff
@@ -151,42 +151,59 @@ def wkt_facet
...
+          .joins(:geographic_items).where(geographic_items: i).pluck(:id)
 
-        s = 'WITH query_wkt_ad AS (' + l.all.to_sql + ') ' +
-          ::AssertedDistribution
-          .joins('JOIN query_wkt_ad as query_wkt_ad1 on query_wkt_ad1.id = asserted_distributions.asserted_distribution_shape_id')
-          .where(asserted_distribution_shape_type: 'GeographicArea')
-          .to_sql
+        # TODO: There are only 58 GAs with shape that have a descendant without
+        # shape: could this be a lookup table instead?
+        k = ::GeographicArea.descendants_of(j).pluck(:id) # Add children that might not be caught because they don't have shapes
```
mjy, line 182: @kleintom does `descendants_of` add an `.order()` clause?  If so major speedup by doing `.unscope(:order)`

lib/queries/asserted_distribution/filter.rb
```diff
@@ -151,42 +151,59 @@ def wkt_facet
...
       end
 
-      def from_wkt_gazetteer(wkt_shape)
-        i = ::GeographicItem.joins(:gazetteers).where(::GeographicItem.covered_by_wkt_sql(wkt_shape))
+      def self.gazetteers_for_geographic_items(geographic_items_sql)
+        i = ::GeographicItem.joins(:gazetteers).where(geographic_items_sql)
 
-        j = ::Gazetteer.joins(:geographic_item).where(geographic_item: i)
+        j = ::Gazetteer
+          .joins(:geographic_item).where(geographic_item: i).pluck(:id)
```
mjy: Is there any difference to using a join here to the `i` query exactly?  Or maybe that's what you were saying you changed from

lib/queries/asserted_distribution/filter.rb
```diff
@@ -151,42 +151,59 @@ def wkt_facet
...
-          .where(asserted_distribution_shape_type: 'Gazetteer')
-          .to_sql
+        gazetteer_ids = j.join(',')
+        if gazetteer_ids.empty?
+          return ::AssertedDistribution.none
+        end
 
-        ::AssertedDistribution.from('(' + s + ') as asserted_distributions')
+        ::AssertedDistribution
+          .where("asserted_distributions.asserted_distribution_shape_id IN (#{gazetteer_ids}) AND asserted_distributions.asserted_distribution_shape_type = 'Gazetteer'")
```
mjy: See 'prefer native' above.

lib/queries/otu/filter.rb
```diff
@@ -427,22 +446,16 @@ def otu_geo_facet
...
+        if geo_mode != true # exact or descendants
+          return referenced_klass_union([a,b,c])
         end
 
-        referenced_klass_union([a,b,c])
+        # Spatial.
+        i = ::Queries.union(::GeographicItem, [a,c])
+
+        from_geographic_items(
+          ::GeographicItem.covered_by_geographic_items_sql(i.pluck(:id))
```
mjy: Feels strange to pass IDs, not joins, but this is gut feeling.

lib/queries/asserted_distribution/filter.rb
kleintom: It doesn't, but good reminder to keep that in mind.

lib/queries/asserted_distribution/filter.rb
kleintom: I changed it to `::Gazetteer.joins(:geographic_item).merge(i)` (though I'm a little surprised that works since the columns in geographic_items_id are unqualified).
Planning to merge on green but please let me know if you had something else in mind, and thanks for the great feedback.

PR #4482, opened 2025-07-15
Update a couple of DwC import help strings

kleintom:
@LocoDelAssembly could you take a look?

### DIFF
--- a/app/javascript/vue/tasks/dwca_import/lang/help/en.js
+++ b/app/javascript/vue/tasks/dwca_import/lang/help/en.js
@@ -1,4 +1,3 @@
-//  New taxon name
 //  language: en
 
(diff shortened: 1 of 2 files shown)


LocoDelAssembly:
I suppose the referencing `ImportDataset::DarwinCore::{Checklist, Occurrences}::MINIMUM_FIELD_SET` would be too cumbersome or impossible to do? @jlpereira

mjy:
We should make a JSON serving endpoint exposing that constant if these data are thereI think.  That would keep us dynamic.  Here we are essentially creating a constant in views.

kleintom:
Fair enough, I'll look into it.

kleintom:
To be completed elsewhere.

Issue #4073, opened 2024-10-03
[Bug]: Searching by spatial does not include collecting events that only have a Geographic Area but no georeference


tmcelrath:
### Steps to reproduce the bug

```bash
1. When I assign a CO to a CE that has only a geographic area (e.g. it's a county-level record so a georeference is meaningless)
2. Then I go to Filter COs
3. And I select "Geographic Area" to Filter
4. And I select "Spatial"
5. Then that CO/CE does not get returned until I add a georeference, even if the GeoArea assigned has a polygon/shapefile. 

Request blocker as this is preventing us from properly downloading records for an upcoming paper.
```

### Expected behavior

Should return all COs/CEs with a GeoArea that spatially matches AND those that have a georeference that matches. 

### Environment

Production

### Version

0.44.2

### Browser Used

Firefox

mjy:
@tmcelrath Need a filter -> JSON example please?

tmcelrath:
Not showing up:
https://sfg.taxonworks.org/tasks/collection_objects/filter?per=500&otu_id%5B%5D=164018&geographic_area_id%5B%5D=33446&geographic_area_mode=true&extend%5B%5D=dwc_occurrence&extend%5B%5D=repository&extend%5B%5D=current_repository&extend%5B%5D=collecting_event&extend%5B%5D=taxon_determinations&extend%5B%5D=identifiers&exclude%5B%5D=object_labels&page=1

tmcelrath:
Should be showing up under first result but isn't: https://sfg.taxonworks.org/tasks/accessions/comprehensive?collection_object_id=788501

mjy:
Notes for resolution, none are 100% statisfactory

* Add `expand_buffer=<meters> option to let users clip in other near-similar shapes; only present when target has shape
* Add `match_name=true` option to let users do lateral/down expansions of the result as well

Evolution beyond this will be piecemeal until  is to move to a single unified dataset (geoboundaries?) with optional user Gazeteer.

https://github.com/SpeciesFileGroup/taxonworks/issues/1954
https://github.com/SpeciesFileGroup/taxonworks/issues/3627

mjy:
More notes
* [ ] Refactor to make modes generalized and addative
* [ ] We are missing extensions in `spatial` the include non-spatial descendants

lvhart2:
[...]

PR #4502, opened 2025-07-28
Support spatially filtering on CE GAs #4073

kleintom:
image
image

With the checkbox checked, spatial queries return results matching against both georeferences and CE geographic area when there are no georefs. (You can't match against only CE GA.) I included a help bubble on the checkbox but I still feel like the text in the screenshot coule be improved?

This should apply to all of the filters that reference CEs: BA, CE, CO, FO, OTU.

@mjy I think the code should be 'fine', but can you take a look to see if this is the way we want to do this, i.e. with a separate checkbox? Spatial queries basically take twice as long with the checkbox checked (and I don't see how to combine the queries; we're joining CE to GI in two different ways and then unioning the result).

### DIFF
--- a/lib/queries/biological_association/filter.rb
+++ b/lib/queries/biological_association/filter.rb
@@ -20,6 +20,7 @@ module Queries
         :geo_mode,
         :geo_shape_id,
         :geo_shape_type,
+        :geo_ce_geographic_area,
         :object_biological_property_id,
         :object_object_global_id,
         :object_taxon_name_id,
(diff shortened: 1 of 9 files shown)


lib/queries/biological_association/filter.rb
```diff
@@ -152,6 +153,7 @@ class Filter < Query::Filter
       attr_accessor :geo_mode
       attr_accessor :geo_shape_id
       attr_accessor :geo_shape_type
+      attr_accessor :geo_ce_geographic_area
```
mjy: @kleintom Spell out `ce`, I know, long, but no abbreviations.

Issue #4423, opened 2025-06-09
OTU citations


mjy:
_Draft_

Aggreating the OTU citation summary related issues here.

The general need is for a "complete" list of Citations on data related to this OTU.  We want need this to:
1) Display on Browse OTU
2) Display on TaxonPages via a new(?) inventory endpoint
3) Aggregate citations for monograph rendering

## Implementation

There are generic methods in lib/catalog that are likely of use, i.e. we should probably use this concept/basis to aggregate citation data.

* Create a new OTU catalog entry 
* Enumerate and iterate to add related data to the entry
  * A object meta-map of resources should be queried is likely needed.  Probably a candidate for ApplicationEnumeration.  For example `@otu.citable_relations(:has_many)`.  This would return some has_many/has_one/belongs to summary, or maybe all.  We'd iterate through each, using `.joins(:ciations)`.
* Must consider coordinate OTUs

* Add an `inventory` endpoint for OTUs

# Related issues
* https://github.com/SpeciesFileGroup/taxonpages/issues/255
* https://github.com/SpeciesFileGroup/taxonpages/issues/256
* https://github.com/SpeciesFileGroup/taxonworks/issues/4064

kleintom:
@jlpereira Let me know if you need anything more from my end on this, but I think it's ready to be José-ified.

jlpereira:
It's done

PR #4503, opened 2025-07-30
Add Otu inventory endpoint and catalog for enumerating related citati…

kleintom:
…ons #4423

This is (currently, mostly) complementary to timeline: it enumerates all citations on objects *associated with* otus that are coordinate with a given otu.

Expecting this to need more work:
* doesn't yet follow associations on the taxon names (only otus) of the otu and its coordinate otus.
* I'm not really clear yet on if/how it will be combined with `timeline`, or if it has all of the information we'll want.
* (It currently loads its data, but doesn't display it, in Browse Otu, for testing.)

### DIFF
--- a/lib/catalog/otu/inventory_entry.rb
+++ b/lib/catalog/otu/inventory_entry.rb
@@ -1,0 +1,48 @@
+# A Catalog::Entry that contains the biological history of an otu via associated
+# data of that otu (images, asserted distributions, observations, etc.).
+class Catalog::Otu::InventoryEntry < ::Catalog::Entry
+
+  def initialize(otu)
+    super(otu)
+    true
+  end
+
+  def build
+    from_self
+    true
+  end
+
+  def to_html_method
+    :otu_catalog_entry_item_to_html
+  end
+
+  def from_self
+    a = ::Otu.coordinate_otus(object.id).load
+    b = ApplicationEnumeration.citable_relations(Otu).values.flatten(1)
+
+    a.each do |o|
+      b.each do |r|
+        association = o.send(r)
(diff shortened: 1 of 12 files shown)


lib/catalog/otu/inventory_entry.rb
```diff
@@ -0,0 +1,223 @@
...
+          end
+        end
+      end
+    end
+  end
+
+  # Filter out STI subclass relations that are redundant with their parent class
+  # (e.g., `protonym` belongs_to duplicates `taxon_name` belongs_to)
+  # @return [Hash] { relation_name => reflection }
+  def filter_sti_relations(relation_names)
```
mjy: Let's move to ApplicationEnumeration, taking class as a way to config.

lib/catalog/otu/inventory_entry.rb
```diff
@@ -0,0 +1,204 @@
...
+    # #process_has_relations is complex due to:
+    # - STI tables (e.g., protonym → taxon_names table)
+    # - Polymorphic sources
+    # - Different join patterns for each through type
+    # So here we just let rails handle everything.
+    includes_hash = through_relations.each_with_object({}) do |rel, hash|
+      hash[rel] = { citations: [:source, :notes, :topics] } # for views
+    end
+
+    coordinate_otus = ::Otu.where(id: coordinate_otu_ids).includes(includes_hash).to_a
```
mjy, line 153: Is this a candidate for memoization in this module?

kleintom: This one's only called once; `from_self` computes the same but with different includes, and that's only created once as well. I think these are okay as is.

PR #4511, opened 2025-08-08
Reset some of the conditions under which cached maps are created

kleintom:
@mjy I think we've discussed all of these, but could you sanity check me here - we're *NOT* going to produce cached maps for:
* otus without a taxon name
* otus whose taxon name is a Combination
* otus whose Protonym taxon name is species group
* Asserted Distributions with `is_absent == true`

### DIFF
--- a/app/helpers/otus_helper.rb
+++ b/app/helpers/otus_helper.rb
@@ -223,7 +223,7 @@ module OtusHelper
       }
     }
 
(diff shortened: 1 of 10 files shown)


mjy:
I thnk `otus without a taxon name` is technically OK, but we can leave off until we hit a problem. I think your assumptions are correct for the rest.

PR #4516, opened 2025-08-11
Cached map updates

kleintom:
Three different issues here:
* Remove CachedMapItem::translate_by_alternate_shape - see commit message for reasoning
* When translating a shape, always include any ne_states shapes contained in the shape. On 3i this takes cached map construction time from 2h10m to 2h23m, 1.076x longer. I tend to think this is worth it, but welcome other thoughts. (Reminder to self: the main issue solved here is that the buffering we use when finding translates can shrink small ne_states to nothing, making them not match, even when they're subsets of the shape being matched against.)
* Fix an occasional 500 we've seen in production


mjy:
>  1.076x longer

That's fine IMO.

app/models/georeference.rb
```diff
@@ -93,6 +91,8 @@ class Georeference < ApplicationRecord
   has_many :collection_object_otus, -> { unscope(:order) }, through: :collection_objects, source: 'otu'
   has_many :field_occurrence_otus, -> { unscope(:order) }, through: :field_occurrences, source: 'otu'
 
+  # TODO: this should probably go away, it's misleading now with multiple
+  # throughs possible.
   has_many :otus, through: :collection_objects, source: 'otus'
```
mjy, line 97: Try go elminate or rename for a more specific source of the otu, perhaps add a parallel method for the other sources of otus.

kleintom: I'll do a followup on this one (`.otus` might not be an easy association to adjust).

PR #4517, opened 2025-08-12
4514 Don't allow delete on last AD citation (or last FO TD)

kleintom:
It's mentioned in the commit message(s), but I was unable to get AD to raise by any means when update! `_destroy` on its origin_citation was rejected (it simply didn't destroy the origin_citation, with no error). LLM suggests this may be the combination of has_one and _destroy (there is no such issue with the `citations` has_many).

The solution was to not allow `_destroy` on origin_citation_attributes (which wasn't being done anyway).

The FO/TD situation is entirely anlagous to the AD/Citations situation, so FO was updated as well.


kleintom:
I wound up in the weeds on this one wading through rails validation/association/has_one vs has_many/reset gotchas, I think I need some more time to sort that out and decide if maybe we want to do this differently. On the plus side, hooray for random specs catching technical rails details I wasn't thinking about.

kleintom:
FWIW, some notes on some of the struggles getting this one right (partly notes for myself):
Pretty much all of the issues were related to the interconnectedness of some of our associations: e.g. on AssertedDistribution we have `source`, `origin_citation`, and `citations`:
* `source` causes creation of an `origin_citation`, `origin_citation` causes the creation of a member of `citations`.
* none of the reverse implications are true (`origin_citation` does *not* create `source`, even though source is through origin_citation)

The creations don't happen until `after_save`, or somewhere in there. Before then - in validations for example - `origin_citation` will be empty, even if `source` is set, etc.
* That complicates accessing those associations, since once an association has been accessed on an ActiveRecord, its value won't be changed again until the association is reset. E.g. if `source` is set and you read `origin_citation` in a validation, then `origin_citation` will *not* be set after save on that active record until you reload the record or reset the association, even though you might expect that saving the `source` is when rails might notice that `origin_citation` should get a new value.

It also complicates checking for whether or not a citation is going to exist after save from a validation, since `marked_for_destruction?` can appear on `origin_citation`, e.g., but not on the corresponding member of `citations`, *even when both already existed on the ad*. [Where it appears seems to match the attributes you set the nested `_destroy` on.] So if you want to know if, e.g., the `source` set on an ad will survive after save has completed, you need to check:
* source
* origin_citation (because if origin_citation goes away, source goes away)
* citations (because if there's a citations member matching origin_citation and that member goes away then origin_citation goes away and source goes away)

The other main issue I was never able to figure out completely was how to pass an error from `before_delete` on an association to its parent so that it respected ! vs not-! (e.g. no raise on `update`, raise on `update!`). I eventually just worked around this, at chatgpt's urging, by moving the processing that handled nested _destroy to the parent, and having the child only concern itself with 'direct' destroys on the child. (The main issue with the original setup was the case where you were doing a _destroy on a has_one like origin_citation, where sometimes you got expected raise/non-raise, sometimes not (it was working on AD, not on FO).)

PR #4535, opened 2025-08-26
4140 observation matrix row annotation cleanup (impacts unify, omr destroy)

mjy:


kleintom:
This fix applies to the case where you want to unify two otus, and the otus are separate rows in a matrix.

This mostly applies to the basic case where you have just two otu rows that need to be merged; if the rows have more than one reference (from other dynamic row items e.g.) this will likely still fail. Feel free to post issues for those cases (though a more 'intrusive' solution will likely be necessary then).

PR #4552, opened 2025-09-17
4527 Interactive key updates to support filtering a dichotomous key

kleintom:


kleintom:
The initial symptoms here were:
* Create lead items for a key from a matrix using 'start an OTUs list':
 
image

* Send the key otus to an interactive key, use a descriptor to narrow down the list, send back to the key. Udpate the key with the descriptor texts.

image


* Create a new couplet, now try to do the same again *with a smaller number of otus than the interactive key covers*: send the right side otus to an interactive key. Note when the key loads that the number remaining is the same as the otu list from your key (or possibly smaller). Now select a state in an unused descriptor as before: note the # of remaining otus jumps as though the key is no longer filtered by the `otu_filter` query param (because it isn't).

Fix that with the first patch. Now note that when you select a descriptor the matrix is reloaded, but without any awareness that the descriptor was selected (you can see descriptor selection data is never sent on the matrix load in devtools network panel). The second patch fixes that.

kleintom:
@jlpereira Could you take a look at this when you get a chance? I'm not very familiar with the matrix code.

kleintom:
BTW I now see the first patch made the second patch necessary because the first patch now calls `setParamsFilter` with the value of `GetFilter`, which adds `selected_descriptors` to `state.filters` where it didn't exist in the initial `state` definition.

Previously we weren't actually setting state.filters when we were doing `matrixFilter.otu_filter = otuIds` (see the patch 1 note), so state.filters didn't have a `selected_descriptors` value at all at the time of descriptor selection, and so the computed `selected_descriptors` value in `getFilter` wasn't being overwritten by a `selected_descriptors === undefined` value like it is with patch 1 but not patch 2. Whew.

PR #4560, opened 2025-09-25
4550 fix sqed issue with pattern None

kleintom:


mjy:
LGTM.  @LocoDelAssembly can you double check the docker changes make sense to you?

PR #4568, opened 2025-10-03
Only display primary types as Type Specimens in distribution maps

kleintom:
Did I get the primary types right?

### DIFF
--- a/app/helpers/type_materials_helper.rb
+++ b/app/helpers/type_materials_helper.rb
@@ -55,9 +55,9 @@ full_original_taxon_name_tag(type_material.protonym)
   # codes !!
   def type_material_is_primary_type(type_material)
     (type_material.protonym.nomenclatural_code == :iczn &&
-      TypeMaterial::ICZN_PRIMARY_TYPES.include?(type_material.type_type)) ||
+      ::ICZN_PRIMARY_TYPES.include?(type_material.type_type)) ||
     (type_material.protonym.nomenclatural_code == :icn &&
-      TypeMaterial::ICN_PRIMARY_TYPES.include?(type_material.type_type)) ||
+      ::ICN_PRIMARY_TYPES.include?(type_material.type_type)) ||
     ![:icn, :iczn].include?(type_material.protonym.nomenclatural_code)
   end
 
--- a/app/models/type_material.rb
+++ b/app/models/type_material.rb
@@ -47,10 +47,6 @@ class TypeMaterial < ApplicationRecord
     'paralectotypes' => Lot
   }.freeze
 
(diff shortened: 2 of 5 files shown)


app/helpers/type_materials_helper.rb
```diff
@@ -51,4 +51,14 @@ def options_for_type_type_select
     options_for_select((TypeMaterial::ICZN_TYPES.keys + TypeMaterial::ICN_TYPES.keys).uniq.sort, selected: 'holotype')
   end
 
+  # !! Only (currently) meaningful for ICN and ICZN - returns true for all other
+  # codes !!
+  def type_material_is_primary_type(type_material)
+    (type_material.protonym.nomenclatural_code == :iczn &&
```
mjy, line 57: @kleintom I think we can maybe be faster with the object loading and targetting a specific value only?

Maybe a select() on the preceeding  o.type_materials.includes(:protonym?

Not sure the peformance gain, coulud also check the n+1 queries.

app/models/type_material.rb
```diff
@@ -47,6 +47,10 @@ class TypeMaterial < ApplicationRecord
     'paralectotypes' => Lot
   }.freeze
 
+  ICZN_PRIMARY_TYPES = [
```
mjy: *Might* go in config/initializers/constants.

app/models/type_material.rb
```diff
@@ -61,6 +65,8 @@ class TypeMaterial < ApplicationRecord
       'isosyntypes' => Lot
   }.freeze
 
+  ICN_PRIMARY_TYPES = ['holotype', 'neotype', 'lectotype', 'syntype', 'epitype'].freeze
```
mjy: See constants comment above, maybe.

app/helpers/type_materials_helper.rb
kleintom: It seems the options for selecting specific columns off of a `.includes` included table on a relation are:
* define an association where the columns you want are baked into the association
* Use (new to me)
```ruby
ActiveRecord::Associations::Preloader.new.preload(
  type_materials, :protonym, Protonym.select(:id, :nomenclatural_code)
)
```
That would work but has the disadvantage that the preload is done immediately, not lazily, so you wind up doing two queries everytime and are perhaps worse off in the small-#-of-associateds case.

So in this case for now at least I've left it as-is; I expect to post a separate pr addressing n+1 in `OtusHelper#otu_distribution` which looks to have multiple levels of n+1 and should be much much bigger gains.

app/models/type_material.rb
kleintom: I moved to initializer, see what you think.

PR #4582, opened 2025-10-14
Accelerated geojson distribution responses

LocoDelAssembly:
```
With this PR  Completed 200 OK in  6347ms (Views:  5634.9ms | ActiveRecord:  651.4ms ( 7804 queries, 1 cached) | GC:  754.3ms)
Without       Completed 200 OK in 16387ms (Views: 13095.3ms | ActiveRecord: 3235.4ms (38953 queries, 1 cached) | GC: 1465.8ms)
```

Tested with "GeoVin (all)" sandbox project using _Triatoma infestans_ OTU (7788 "collection objects" this OTU alone).

Unfortunately there are still massive amounts of queries like below which make response time still high:
```
  Identifier Pluck (0.1ms)  SELECT "identifiers"."cached" FROM "identifiers" WHERE "identifiers"."identifier_object_id" = $1 AND "identifiers"."identifier_object_type" = $2 AND (identifiers.project_id = $3 OR identifiers.type ILIKE 'Identifier::Global%')  [["identifier_object_id", 1429214], ["identifier_object_type", "CollectionObject"], [nil, 445]]
```

### DIFF
--- a/spec/models/collecting_event/geo_spec.rb
+++ b/spec/models/collecting_event/geo_spec.rb
@@ -32,7 +32,7 @@ describe CollectingEvent, type: :model, group: [:geo, :shared_geo, :collecting_e
     collecting_event.save!
     a = Georeference::Wkt.create!(wkt: 'POINT (10 10)', collecting_event: collecting_event)
     b = FactoryBot.create(:valid_georeference_verbatim_data, collecting_event: collecting_event)
-    expect(collecting_event.preferred_georeference.reload).to eq(b)
+    expect(collecting_event.reload.preferred_georeference).to eq(b)
   end
 
   context 'with political areas and collecting events generated' do


app/models/collecting_event/georeference.rb
```diff
@@ -16,7 +16,7 @@ module CollectingEvent::Georeference
     accepts_nested_attributes_for :gpx_georeferences
 
     def preferred_georeference
-      georeferences.sort_by(&:position).first
+      georeferences.sort_by { |g| [g.position.nil? ? 1 : 0, g.position.to_i] }.first
```
mjy, line 19: Should we just use raw sql to start and pick the first?! `georeferences.order(:position).first`

LocoDelAssembly: No, that would cause an extra query for every CO. Given that it is very unlikely that a collecting event has lots and lots of georeferences it is best to eager load them and sort in memory.

kleintom:
@LocoDelAssembly if you see anything in #4584 that you think is worth grabbing, please feel free (it's more chatgpt's code than it is mine), just ignoring it is also completely fine, no need to justify. I have a few things to get through before I'd get back to thinking about it again.

LocoDelAssembly:
@kleintom It is probably best we continue on your branch and enable builds on it. Maybe moving the branch on this repo also to facilitate contribution.

LocoDelAssembly:
Superseded by #4585

PR #4584, opened 2025-10-14
Reduce n+1 for OtusHelper#otu_distribution

kleintom:
Code mainly by chatgpt.

With 1 FO, 1 Type specimen, 6 COs, and 6 ADs we went from 92 queries to 34 queries, according to this chatgpt script run in console: via 'load /tmp/devPerf.rb'; DevPerf.measure("otu_distribution") { ApplicationController.helpers.otu_distribution(Otu.find(<n>)) }

Perhaps we can find a cleaner way to include the 'if association already exists' checks added here?

Note the `conn.disabled_query_cache!`
```ruby
module DevPerf
  def self.measure(label)
    conn = ActiveRecord::Base.connection
    conn.disable_query_cache!
    total = 0
    sub = ActiveSupport::Notifications.subscribe('sql.active_record') do |*args|
      e = ActiveSupport::Notifications::Event.new(*args)
      next if e.payload[:name] == 'SCHEMA' || e.payload[:cached]
      total += 1
    end
    t0 = Process.clock_gettime(Process::CLOCK_MONOTONIC)
    yield
    ms = ((Process.clock_gettime(Process::CLOCK_MONOTONIC) - t0) * 1000).round(1)
    ActiveSupport::Notifications.unsubscribe(sub)
    puts "[#{label}] queries=#{total}, elapsed=#{ms}ms"
  ensure
    conn.enable_query_cache!
  end
end
```

There's still a GI load we're missing:
```
DevPerf.measure("otu_distribution") { ApplicationController.helpers.otu_distribution(otu) }
  Otu Load (0.4ms)  SELECT "otus".* FROM "otus" INNER JOIN "taxon_names" ON "otus"."taxon_name_id" = "taxon_names"."id" INNER JOIN "taxon_name_hierarchies" ON "taxon_names"."cached_valid_taxon_name_id" = "taxon_name_hierarchies"."descendant_id" WHERE ("taxon_name_hierarchies"."ancestor_id" IN (1159408))
  TaxonName Load (0.4ms)  SELECT "taxon_names".* FROM "taxon_names" WHERE "taxon_names"."id" IN ($1, $2, $3, $4)  [["id", 1159417], ["id", 1159415], ["id", 1159408], ["id", 1159416]]
[...]

PR #4585, opened 2025-10-15
Reduce query count/increase speed in OtusHelper#otu_distribution

kleintom:
See #4582 #4584 for history.


LocoDelAssembly:
TODO
===

Verify if we need `georeferences.min_by { |g| [g.position.nil? ? 1 : 0, g.position.to_i] }` or not. Specs are all working without it but that is likely due to georeferences not being preloaded. `position` is nullable.

app/helpers/identifiers_helper.rb
```diff
@@ -92,9 +92,21 @@ def simple_identifier_list_tag(object)
   # @return [String, nil]
   #   a list of identifiers *without* HTML
   def identifier_list_labels(object)
-    ids = visible_identifiers(object).pluck(:cached)
-    return nil unless ids.any?
-    ids.join(', ')
+    if object.association(:identifiers).loaded?
+      return nil unless object&.has_identifiers?
+
+      project_id = defined?(controller) && controller ? sessions_current_project_id : nil
```
mjy: Feels like we want a scope on identifiers like `visible_to_project(project_id)`?

mjy:
@kleintom This looks OK to me, some new paradigms to grok (loaded?), but understood.  Feel free to merge if still relevant.

kleintom:
Merging as is, but @LocoDelAssembly let's continue to persue your CACHE and other optimizations here.

PR #4588, opened 2025-10-20
4407 anatomical part

kleintom:
Remaining TODOs:
- [ ] DwC occurrence rows for anatomical parts
- [x] `is_material` cleanup: shouldn't be able to edit it in descendants; validate CO can only be is_material, etc
- [x] Add `is_material` validation for containerized APs
- [x] Add soft validation for uri

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ This project <em>does not yet</em> adheres to [Semantic Versioning](https://semv
 - Added soft validation for unjustified emendation
 - Support repeated attributes with different values (to be ORed) in Collecting Event filter's Collecting Event attribute facet [#4622]
 - Batch create Labels from Collecting Event filter [#3814]
+- New status/control area on Complete DwCA Settings page
 
 ### Changed
 
@@ -43,6 +44,7 @@ This project <em>does not yet</em> adheres to [Semantic Versioning](https://semv
 - Filtering on Data Attributes where the same attribute is repeated with a different value (matches to be ORed)
 - Filtering on Data Attributes in Filter Asserted Distributions
 - Prioritize abbreviated genus and species name in TaxonName autocomplete
+- Make sure file is provided when doing bibtex batch loads
 
 [#3814]: https://github.com/SpeciesFileGroup/taxonworks/issues/3814
 [#4556]: https://github.com/SpeciesFileGroup/taxonworks/issues/4556
--- a/app/javascript/vue/components/Filter/Table/TableResults.vue
+++ b/app/javascript/vue/components/Filter/Table/TableResults.vue
@@ -109,10 +109,12 @@
                 @click.stop="
                   () =>
                     copyColumnToClipboard(
-                      list
-                        .filter(rowHasCurrentValues)
-                        .map((item) => item[attr])
-                        .join('\n')
+                      escapeHtml(
+                        list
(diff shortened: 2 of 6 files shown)


kleintom:
- [ ] Publish Hookaido and change gemfile to use published gem.

mjy:
@kleintom Some perhaps final bits:

* [ ] Let's include OTU object tag in the AP object tag, after I think , like `[part name]: [otu_object_tag]`.  This will be necessary in the observation matrix for example.
* [ ] Style/hyperlink the URI in `show` in a format that indicates clicking will take you out of the app.  @jlpereira Do we have a general link style for this? This `http://purl.obolibrary.org/obo/UBERON_6004475 `

app/controllers/anatomical_parts_controller.rb
```diff
@@ -0,0 +1,154 @@
...
+    @anatomical_parts = AnatomicalPart.select_optimized(sessions_current_user_id, sessions_current_project_id, params[:target])
+  end
+
+  def graph
+    @nodes, @origin_relationships =
+      AnatomicalPart.graph(params.require(:id), params.require(:type))
+  end
+
+  def ontologies
+    render json: ::Hookkaido.ontologies
```
mjy, line 120: We may ultimately create a constant of known *NOT* antomical part ontologies and subtract those from this list.

mjy: Or maybe Hookkaido can provide a function like `::Hookkaido.ontologies_in('UBERON')`

app/controllers/anatomical_parts_controller.rb
```diff
@@ -0,0 +1,154 @@
...
+  private
+
+  def set_anatomical_part
+    @anatomical_part = AnatomicalPart.find(params[:id])
+  end
+
+  # Only allow a list of trusted parameters through.
+  def anatomical_part_params
+    params.require(:anatomical_part).permit(:name, :uri, :uri_label,
+      :is_material, :preparation_type_id, :preparation_type,
```
mjy, line 151: Note to self- why do we have `:preparation_type` here?

app/models/anatomical_part.rb
```diff
@@ -0,0 +1,423 @@
...
+  end
+
+  # @return [Scope]
+  #    the max 10 most recently used anatomical_parts
+  def self.used_recently(user_id, project_id, used_on = '')
+    old_types = valid_old_object_classes
+    new_types = valid_new_object_classes
+    return [] if !old_types.include?(used_on) && !new_types.include?(used_on)
+
+    t = used_on.constantize.arel_table
```
mjy, line 143: @kleintom confirming you see this flagged security warning.

mjy:
@kleintom This is an amazing amount of work.  I've gone through the majority now looking just at the architecture/patterns, I see no red flags. If we merge now we can layer the News branch @jlpereira is working on on top of it.

I.e. please merge at your convenience.

app/models/anatomical_part.rb
kleintom: Yah I went through all of the warnings, we seemed go on all of them to me - here we check that `used_on` is in either `valid_old_object_classes` or `valid_new_object_classes` prior to constantizing it, that seems fine to me (?).

kleintom:
Thanks @mjy! Merging.

PR #4593, opened 2025-10-22
4581 Persistent API endpoint udpates

kleintom:


kleintom:
Still need to add an api endpoint that returns a png for an image id (related to supporting TIFF handling in other systems)

kleintom:
Added new `/images/:1/as_png` and `/api/v1/images/:1/as_png` routes which return the original image as a png.

@mjy, do you remember why we return a jpg for `original_png`? Is it okay to switch to png (as I do in this pr for the new `as_png` routes)?

https://github.com/SpeciesFileGroup/taxonworks/blob/45aff08fb6807cf49312efb4d416c7b07141ef7d/app/views/images/api/v1/_attributes.json.jbuilder#L6

https://github.com/SpeciesFileGroup/taxonworks/blob/45aff08fb6807cf49312efb4d416c7b07141ef7d/app/helpers/images_helper.rb#L22-L30

https://github.com/SpeciesFileGroup/taxonworks/blob/45aff08fb6807cf49312efb4d416c7b07141ef7d/app/controllers/images_controller.rb#L194-L197

https://github.com/SpeciesFileGroup/taxonworks/blob/45aff08fb6807cf49312efb4d416c7b07141ef7d/app/models/image.rb#L451-L459

PR #4600, opened 2025-10-25
Permit subclass SV meta declaration to override base class SV declara…

kleintom:
Written with chatgpt.

Permit subclass SV meta declaration to override base class SV declaration
    
Fixes `fix:` for icn/icnp `:sv_not_specific_relationship` SV - the previous version attempted to add a different `fix:` in the subclass than in the base class, which couldn't be done.
    
Now it can, by overwriting the base class sv with a new subclass declaration (previously declaring in both places would have resulted in two different SVs).
    
STR (e.g.):
  * create two ICNP (or ICN) genera in New Taxon Name
  * synonymize them using 'Junior synonym of' taxon_name_relationship
    
Get a 500. Get a 500 when the page reloads. See no TNR SVs.
    
Exception report:
```
    An ActionView::Template::Error occurred in browse#index:
    
      undefined method 'fix=' for an instance of SoftValidation::SoftValidation
      lib/soft_validation/soft_validation.rb:53:in 'block in SoftValidation::SoftValidation#initialize'
    
    Github link:
    
      https://github.com/SpeciesFileGroup/taxonworks/blob/ca0670348/lib/soft_validation/soft_validation.rb#L53
    
    Request:
    
      * URL        : https://sandbox.taxonworks.org/tasks/nomenclature/browse?taxon_name_id=2402420
```

### DIFF
--- a/app/models/taxon_name_relationship/icn/unaccepting/synonym.rb
+++ b/app/models/taxon_name_relationship/icn/unaccepting/synonym.rb
@@ -2,6 +2,15 @@ class TaxonNameRelationship::Icn::Unaccepting::Synonym < TaxonNameRelationship::
 
   NOMEN_URI='http://purl.obolibrary.org/obo/NOMEN_0000372'.freeze
 
+  # Override the base-class metadata for this SV
+  soft_validate(
+    :sv_not_specific_relationship,
+    set: :not_specific_relationship,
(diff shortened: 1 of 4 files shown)


kleintom:
@proceps When you get a chance could you take a look?

Issue #4603, opened 2025-10-28
Model: CitationsDocumentation


mjy:
When we have things like BHLNames index, and people start giving us URLs of BHL pages then we want to mantain page-level metadata.

In discussion this probably belongs in a new model linking Citations to precise locations in Documents.

We can expand documentation to include the page_metadata object discussed elsewhere.

We can think of things like character_offset for OCR level precision of the OCR of Documents.

`page_metadata` can model or persist the page triple of BHL - `page_num`, `page_id`, `item_id`.

mjy:
If we dont' have a PDF, then we may need a Document::Virtual::Bhl model to create on demand so that we can crossreference Citation to Documentation that references that model.

PR #4605, opened 2025-11-01
4603 BHL url to CitationDocumentation

kleintom:
# Functionality to add (please correct/amend/etc.)
@mjy has a much fuller/clearer picture of the goals here, these are my scattered rememberings of what we discussed.

* User is on a TaxonName
  * They may even already have a source and maybe citation on the TN
    * They find that source in BHL
    * They want to record the BHL reference information with their TN

BHL reference should be more than just a url, it should include page mapping (from BHL page numbers to article page numbers) and perhaps ocr mapping as well
 * with eventually maybe a way to send this validated-by-human mapping back to GN/BHL

Relation to #4603: as discussed there, likely we'll want to create a 'virtual document' to represent the BHL document, and a (new) CitationDocumentation object to associate the (TN, TN source, BHL copy of TN source) data, which is where the page mapping etc data will live.


kleintom:
This is just a small first start with many TODOs.

The main blocker in sight at this point is that `Document` has multiple required `file_` attributes, which will make it challenging to subclass to a `Document::Virtual` class not having any local file associations: https://github.com/SpeciesFileGroup/taxonworks/blob/47b28d34f716f60ce9656c2b7dd999eaac7f1f11/db/schema.rb#L744-L748

app/models/source.rb
```diff
@@ -395,6 +395,35 @@ def clone
...
+        item_id: h[:item_id], page_num: h[:page_num] || "1"
+      )
+    end
+
+    return [nil, []] unless bhl_source
+
+    [bhl_source, bhl_source.matching_sources]
+  end
+
+  def matching_sources
```
mjy, line 416: There were some inklings of this in the IsData concerns, similar, identical rules.  Didn't get too far.

PR #4606, opened 2025-11-03
4592 Fix async batch_update

kleintom:
Fixes two issues:
1) we didn't have `project_id` when running batch updates async
* Explicitly add `project_id` and `user_id` when running any query_batch_update

2) existing batch async specs were giving false positive because in specs, background jobs run in the same thread as the spec and were therefore inheriting Current from the values auto-set in the (model) specs.
* Remove any existing `Current` values when running background jobs in specs
  * add specs for that

* Add specs for *all* query_batch_update uses, sync and async

### DIFF
--- a/app/models/concerns/shared/batch_by_filter_scope.rb
+++ b/app/models/concerns/shared/batch_by_filter_scope.rb
@@ -38,6 +38,11 @@ module Shared::BatchByFilterScope
       r.total_attempted = c
       r.async = async
 
+      # Fail loudly now, otherwise we'll fail silently in the background.
+      if async && (user_id.nil? || project_id.nil?)
+        raise TaxonWorks::Error, "user_id (#{user_id}) or project_id (#{project_id}) not set in batch_by_filter_scope! '#{filter_query}, #{mode}, #{params}"
+      end
+
       r = self.process_batch_by_filter_scope(
         batch_response: r,
         query: q,
--- a/app/models/concerns/shared/query_batch_update.rb
+++ b/app/models/concerns/shared/query_batch_update.rb
@@ -41,6 +41,8 @@ module Shared::QueryBatchUpdate
       a = request.filter
 
       if request.run_async?
+        # Fail loudly now, otherwise we'll fail silently in the background.
+        raise TaxonWorks::Error, "user_id or project_id not set in query_batch_update! '#{request}'" if request.user_id.nil? || request.project_id.nil?
         a.all.find_each do |o|
           o
             .delay(run_at: Proc.new { 1.second.from_now }, queue: :query_batch_update)
--- a/spec/models/protocol_relationship_spec.rb
+++ b/spec/models/protocol_relationship_spec.rb
@@ -21,7 +21,9 @@ RSpec.describe ProtocolRelationship, type: :model, group: :protocol do
       params: {
         protocol_id: c1.id,
         replace_protocol_id: protocol_relationship.id
-      }
+      },
+      user_id: User.first.id,
+      project_id: Project.first.id
     )
 
     perform_enqueued_jobs
@@ -42,7 +44,9 @@ RSpec.describe ProtocolRelationship, type: :model, group: :protocol do
       params: {
         protocol_id: c1.id,
         replace_protocol_id: protocol_relationship.id
-      }
+      },
+      user_id: Current.user_id,
+      project_id: Current.project_id
     )
 
     perform_enqueued_jobs
--- a/spec/models/biological_association_spec.rb
+++ b/spec/models/biological_association_spec.rb
@@ -158,6 +158,44 @@ describe BiologicalAssociation, type: :model do
     expect(b.reload.biological_relationship).to eq(r)
   end
 
+  specify '.batch_update() (async) raises error when user_id is missing' do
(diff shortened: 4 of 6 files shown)


app/jobs/batch_by_filter_scope_job.rb
```diff
@@ -15,6 +15,10 @@ def perform(
     project_id: nil, user_id: nil
   )
     begin
+      # Set a 'session' context in this background job:
+      Current.user_id = user_id
```
mjy, line 19: My gut feeling is that we need to send user_id and project_id in as params here, and pass them through, not set Current at this point, but still looking.

app/models/concerns/shared/batch_by_filter_scope.rb
```diff
@@ -38,6 +38,11 @@ def batch_by_filter_scope(
       r.total_attempted = c
       r.async = async
 
+      # Fail loudly now, otherwise we'll fail silently in the background.
+      if async && (user_id.nil? || project_id.nil?)
```
mjy, line 42: Would be best to handle this in specs, but OK for now. I.e. ensure we never need to Raise.

app/models/concerns/shared/query_batch_update.rb
```diff
@@ -45,6 +41,8 @@ def query_batch_update(request)
       a = request.filter
 
       if request.run_async?
+        # Fail loudly now, otherwise we'll fail silently in the background.
+        raise TaxonWorks::Error, "user_id or project_id not set in query_batch_update! '#{request}'" if request.user_id.nil? || request.project_id.nil?
```
mjy, line 45: See prior comment about specs > raise.  Again, OK here.

lib/queries/query/filter.rb
```diff
@@ -834,6 +837,12 @@ def only_project?
       a
     end
 
+    # See also only_project?
```
mjy, line 840: I see, you didn't change `only_project?`, but rather extended an alternative- good.

spec/models/otu_spec.rb
```diff
@@ -322,10 +322,13 @@
       q = ::Queries::Otu::Filter.new({otu_id: [o0.id, o1.id]})
 
       params = {
+        async_cutoff: 1,
         otu: { taxon_name_id: t.id },
+        user_id: Current.user_id,
```
mjy, line 327: Can we be more explicit somehow and reference a param? This feels leaky to reference Current directly.

spec/models/protocol_relationship_spec.rb
```diff
@@ -42,7 +44,9 @@
       params: {
         protocol_id: c1.id,
         replace_protocol_id: protocol_relationship.id
-      }
+      },
+      user_id: Current.user_id,
```
mjy, line 48: See comments on "leaky"

spec/support/job_current_isolation.rb
```diff
@@ -0,0 +1,58 @@
+# Make inline job execution in tests simulate a clean worker thread/process.
+# This ensures jobs don't inherit Current.user_id, Current.project_id, etc.
+# from the caller's thread, matching production worker behavior.
+#
+if Rails.env.test?
+  module JobIsolationHelpers
+    def with_job_isolation
+      # Save all contexts
+      previous_current = Current.attributes
```
mjy, line 9: Do we really want *all* Current? For now its safe because we only use proj/user, so I think this is OK.  We just must be vigilant long term about *not* extending Current use.

app/jobs/batch_by_filter_scope_job.rb
kleintom: The tricky part to just sending them as params is that we're often doing nested creates, in which case you would have to dig through the params to look for nesteds and then add user/project to each as well. This seems better?

mjy: I think it's OK to set a context at the beginning of each job, same paradigm as each requests.  As long as we keep it in one place.

app/models/concerns/shared/batch_by_filter_scope.rb
kleintom: Here the problem is that it's the callers of the `batch_by_filter_scope` model concern that need to provide user/project; those come from controllers, so everytime somebody adds a new use of `batch_by_filter_scope` or `query_batch_update` whose call chain starts in a controller, it's *at the controller* that user/project needs to be inserted. We can't control/spec that at this point. 
So yes, existing users should be speced, but if there's a spec writing/execution issue, I'm thinking better to raise than fail silently.

spec/models/otu_spec.rb
kleintom: Yah this felt awkward, and it's more awkward in specs than in actual use (where the values would have passed from a previous caller). I think we can switch to using the user/project of the data we're sending.

spec/support/job_current_isolation.rb
kleintom: We want whatever Current is set to in background jobs run from the UI - I was assuming that would always be nothing, but right, that could change in the future.

mjy: Well- it won't change hopefully, this mess is all a reflection of Current in general.  *Very* glad we have conciously kept it minimally used.

PR #4615, opened 2025-11-14
Prevent autcomplete exceptions on empty strings

kleintom:
Multiple of our controller do `params.require(:term)` even though we weren't preventing empty term from the client side.

```
An ActionController::ParameterMissing occurred in sources#autocomplete:

      param is missing or the value is empty: term
      app/controllers/sources_controller.rb:155:in 'SourcesController#autocomplete'

    Github link:

      https://github.com/SpeciesFileGroup/taxonworks/blob/dccc6df08/app/controllers/sources_controller.rb#L155

    Request:

      * URL        : https://sandbox.taxonworks.org/sources/autocomplete?escape=false&method=id&tag_id=source_id_for_quick_search_form&term=%20
      * HTTP Method: GET
      * Parameters : {"escape" => "false", "method" => "id", "tag_id" => "source_id_for_quick_search_form", "term" => " ", "controller" => "sources", "action" => "autocomplete"}
      * Timestamp  : 2025-11-14 16:40:49 UTC
```

### DIFF
--- a/app/javascript/vue/components/ui/Autocomplete.vue
+++ b/app/javascript/vue/components/ui/Autocomplete.vue
@@ -410,7 +410,9 @@ export default {
     },
 
     update() {
-      if (this.type.length < Number(this.min)) return
+      if ((this.type.length < Number(this.min)) || (this.type.trim() === '')) {
+        return
+      }
 
       this.clearResults()
 


kleintom:
STR: type a space in one of the rails model search boxes (not vue).

Apparently 17 controllers still have `params.require(:term)` for their autocomplete action - this patch prevents those from exploding on white-string searches from the rails search boxes.

It does *not* prevent searches like ' time' or 'any '.

@jlpereira do you feel good about this? (This may still be an issue with the vue autocomplete sending empty searches if you're interested, I'm not sure.)

jlpereira:
@kleintom I think we have this bug in vue autocomplete too. If you want to fix that too, add it to this same PR.

PR #4631, opened 2025-12-08
Convert Collection Object to Field Occurrence

kleintom:
* Copies overlapping attributes and `belongs_to`
  * skips dwc_occurrence_id (FO will generate its own)
* moves overlapping `has_one` and `has_many`
  * skips dwc_occurrence UUID (FO will generate its own)
  * preserves value of a Global::UUID (could argue to *not* do this as well...)

It's a red button with confirmation modal (since it destroys the CO on success) on CSD:
image

@mjy could you take a look at the "transmute" code that moves associations?


app/models/field_occurrence.rb
```diff
@@ -80,17 +80,99 @@ def biological_association_indices
...
+  # @return [Integer, String] returns the new FieldOccurrence ID on success, or an error message on failure
+  def self.transmute_collection_object(collection_object_id)
+    co = CollectionObject.find_by(id: collection_object_id)
+
+    return 'Collection object not found' if co.nil?
+    return 'Collection object must have a collecting event' if co.collecting_event_id.nil?
+    return 'Collection object must have at least one taxon determination' if co.taxon_determinations.empty?
+
+    if co.total.nil? && co.ranged_lot_category_id.nil?
+      return 'Collection object must have either total or ranged lot category'
```
mjy: This should be impossibe to not have because of prior validation.

app/models/field_occurrence.rb
```diff
@@ -80,17 +80,99 @@ def biological_association_indices
...
+    end
+
+    if co.total.present? && co.total <= 0
+      return 'Collection object total must be positive'
```
mjy: Redundant with CO validation.

app/models/field_occurrence.rb
```diff
@@ -80,17 +80,99 @@ def biological_association_indices
...
+    end
+
+    incompatible_identifiers = co.identifiers.select do |identifier|
+      identifier.is_a?(Identifier::Local::CatalogNumber) ||
+      identifier.is_a?(Identifier::Local::FieldNumber) ||
+      identifier.is_a?(Identifier::Local::RecordNumber) ||
+      identifier.is_a?(Identifier::Local::TripCode)
```
mjy: I think better to add validation restrictions on Identifier classes if we want.  Then when we move them over _and_ they are invalid we can destroy them or fail.  I.e. try to move them all, if there are some that are invalid, abort the process.

app/models/field_occurrence.rb
```diff
@@ -80,17 +80,99 @@ def biological_association_indices
...
+    return 'Collection object has a repository assignment. Please remove repository before converting.' if co.repository_id.present?
+    return 'Collection object has a current repository assignment. Please remove current repository before converting.' if co.current_repository_id.present?
+    return 'Collection object has a preparation type. Please remove preparation type before converting.' if co.preparation_type_id.present?
+    return 'Collection object has type materials. Please remove type materials before converting.' if co.type_materials.any?
+    return 'Collection object has sqed depictions. Please remove sqed depictions before converting.' if co.sqed_depictions.any?
+    return 'Collection object has extracts. Please remove extracts before converting.' if co.extracts.any?
+    return 'Collection object has sequences. Please remove sequences before converting.' if co.sequences.any?
+
+    fo = FieldOccurrence.new(
+      total: co.total || 0,  # DB requires NOT NULL, use 0 for ranged lots
```
mjy, line 102: Can we us `co.attributes` directly to get all the attributes without worrying about mapping?

Hmm- do we need to allow ranged_lots for FOs too, probably.  Even though ranged_lots are not yet used widely.

app/models/field_occurrence.rb
```diff
@@ -80,17 +80,99 @@ def biological_association_indices
...
+      FieldOccurrence.transaction do
+        # Move first taxon determination to satisfy FieldOccurrence validation
+        first_td = co.taxon_determinations.first
+        fo.taxon_determinations << first_td
+
+        fo.save!
+
+        # Move remaining taxon determinations using update_column to avoid acts_as_list conflicts
+        co.taxon_determinations.reload.each_with_index do |td, index|
+          td.update_column(:taxon_determination_object_id, fo.id)
```
mjy: think we can pass by object-  `td.update(taxon_determination_object: fo)`

app/models/field_occurrence.rb
```diff
@@ -80,17 +80,99 @@ def biological_association_indices
...
+          td.update_column(:taxon_determination_object_type, 'FieldOccurrence')
+          td.update_column(:position, index + 2)  # Start at position 2 (first determination is at 1)
```
mjy: Should have to never touch position, it's handled automagically.

lib/utilities/rails/transmute.rb
```diff
@@ -0,0 +1,83 @@
...
+  #   * through – Skipped as these are convenience relations built on top of
+  #               other associations.
+  #
+  # @example
+  #   co = CollectionObject.find(123)
+  #   fo = FieldOccurrence.new(collecting_event: co.collecting_event)
+  #   fo.save!
+  #   Utilities::Transmute.move_associations(co, fo)
+  #
+  def self.move_associations(source, target)
```
mjy, line 31: @kleintom There are perhaps subtle relationships to unify code- might check that out to see if there are edge cases to consider.  Maybe not, because we're not merging records, but something feels similar.

lib/utilities/rails/transmute.rb
```diff
@@ -0,0 +1,83 @@
...
+    source.class.reflections.each_value do |reflection|
+      # Skip belongs_to - those should already be set on target
+      next if reflection.macro == :belongs_to
+
+      # Skip through associations - they're convenience relations
+      next if reflection.options[:through]
+
+      # Only process if target class has the same association
+      next unless target.class.reflections.key?(reflection.name.to_s)
```
mjy: Nice.

lib/utilities/rails/transmute.rb
mjy: Wonder if we should do a project id check here, just for redundancy, ensure source/target are in the same world.

mjy: Should this be private, in the sense that we want to restrict its use to be inside a transaction?

spec/lib/utilities/rails/transmute_spec.rb
```diff
@@ -0,0 +1,219 @@
+require 'rails_helper'
+
+describe Utilities::Rails::Transmute, type: :model do
+  describe '.move_associations' do
+    context 'has_one associations' do
+      specify 'moves origin_citation' do
+        collection_object = FactoryBot.create(:valid_collection_object,
+          origin_citation_attributes: { source: FactoryBot.create(:valid_source) })
```
mjy: I think you can directly use `source:` and it will back propagate the citation?

app/models/field_occurrence.rb
kleintom: I seem to get errors on any column unknown to `fo` (preparation_type_id, repo_id, etc), so sticking with opt-in instead.

app/models/field_occurrence.rb
kleintom: Claude got stuck on this somehow and I 'helped' it workaround what I too thought should have worked and now does again... Thanks for calling it out.

PR #4642, opened 2025-12-15
Upgraded rails gem to 8.1

LocoDelAssembly:
TODO
===
VERY CAREFULLY go through `bin/rails app:update`

DO NOT MERGE AT THIS POINT


kleintom:
When running `rake parallel:spec[6]` locally:

`
Calling `distinct` with `update_all` will raise an error in Rails 8.2.
 (called from block in Identifier::Local::CatalogNumber#update_dwc_occurrence at /home/klein/sandbox/taxonworks/app/models/concerns/shared/dwc_occurrence_hooks.rb:40)
`

`
DEPRECATION WARNING: `config.active_support.to_time_preserves_timezone` is deprecated and will be removed in Rails 8.2 (called from <top (required)> at /home/klein/sandbox/taxonworks/config/environment.rb:9)
`

PR #4668, opened 2026-01-06
4663 Support venn in background jobs and subqueries

kleintom:
@mjy It feels like this came out clean, what do you think? 

The issue was that venn params were not being permitted in the params hash used to reconstitute param-ified queries - should now work for venn in background jobs (where queries are passed by their params) and venn in subqueries (where subqueries are passed by params).


mjy:
Thanks @kleintom, looks clean indeed.

Issue #4659, opened 2025-12-17
[Idea]: translate filter facets with complex relations to a textual description, updated as the user selects facet options


kleintom:
### Feature or enhancement

Brian Fisher mentioned that he's seen this sort of thing in other apps: as you change options it displays your combination of currently selected options via a textual description at the top of the page.

Matt commented that our behavior for *between*-facet results will likely always be AND, but this idea might still be useful for cases like that being discussed in #4634 (where up until that issue, it wasn't clear if results were being 'and'ed or 'or'ed together).

### Location

Filters

Issue #4634, opened 2025-12-11
Task - CE Filter: find one field attribute within another field attribute


AntWeb-org:
### Feature or enhancement

The new functionality based on #4324  and #4622  works really well to build filter results.

Is there a way to find one field attribute within another field attribute? Maybe add a "within" option?

For example, I want to find all CEs from the Southwestern Research Station in Arizona collected by B.L. Fisher. Result should be 5 CEs.
I would filter all CEs of the field **verbatim_collector**, value “fisher” and the field **verbatim_locality**, value “southwestern”.

Right now, I am getting _all_ CEs including the **verbatim_collector** “fisher” and _all_ CEs including the **verbatim_locality** “southwestern”.

If I use the Venn panel, I still have to perform 2 separate filters in 2 separate windows.

I have been testing this feature but perhaps I'm missing something.

### Location

Task - CE Filter

### Screenshot, napkin sketch of interface, or conceptual description

**CE filter field parameters and results including verbatim_collector, value "fisher" from Madagascar:**

<img width="1486" height="827" alt="Image" src="https://github.com/user-attachments/assets/51ebac22-6c95-4aa3-9660-1e245ab6bf4d" />

bpescador:
When using the Collection Event (CE) filter with multiple text fields populated, the search logic appears to combine fields using OR rather than AND. This produces inflated and misleading result sets when users are trying to narrow a query.

This behavior makes the filter effective for batch editing / adding data, but unreliable for scoping or validating subsets of CEs.

**Steps to Reproduce**

Go to Collection Events → Filter

Enter:

verbatim_locality: southwestern

verbatim_collector: fisher

Run the filter

**Expected Behavior**

The filter should return only Collection Events that contain both:

"southwestern" in verbatim_locality, and

"fisher" in verbatim_collector

In this case, the expected result set is ~5 CEs.

**Actual Behavior**

The filter returns ~43,076 CEs, which appears to include:

all CEs containing "fisher" OR

all CEs containing "southwestern"

This includes many unrelated CEs (e.g., from Madagascar) that match only one of the fields.

**Why This Matters**

Users reasonably assume that filling multiple filter fields will narrow results (AND logic).

The current behavior is counter-intuitive and misleading, especially for data auditing and validation workflows.

It is difficult for users to tell whether they are working on a correct subset of CEs.

Even a UI hint or toggle would help avoid incorrect assumptions.

**Suggested Solutions (any of the following)**

[...]

PR #4686, opened 2026-01-09
4634 And/Or for attribute facets

kleintom:
@mjy, @jlpereira This is a proposal to see what you both think before I finish things up (not looking for a review, it's unfinished code).

Reminder of where this came from (#4634): the Filter CE Attributes facet currently ORs results for different attributes where AND was desired/expected.

Constraint: *server-side* facets **always** AND their results.

Given that constraint, we want:
Proposal 1) to allow the user to choose to AND/OR results within a given *server-side* facet (since we can't control AND/OR outside of those bounds). [This is slightly different than what I proposed in 4634.]

That requires that users have some concept of what the server-side facets are, *in the UI*, which at the moment is not at all clear. The proposal is to:
Proposal 2) separate the user attribute selections into separate tables labelled with the server-side facet(s) corresponding to those tables, see screenshots. (Another option here would be to go the #4659 route) 

I (personally at least) already refer to the UI filter panels as facets, so I called these server-side facets "groups" instead. The affected panels would be `Data Attributes`, `Import Attributes`, and `By Attributes (for a model)` (José have I missed any?) Here's what it currently looks like.

image

image

The one UI By Attributes panel actually splits into 4 panels server-side, and the Data Attributes panel splits in 6(!).

Proposal 3? You can see it would also be pretty trivial at this point to offer the and/or option for each group if we want to.

### DIFF
--- a/app/javascript/vue/components/Filter/Facets/CollectingEvent/FacetCollectingEvent/ByAttribute.vue
+++ b/app/javascript/vue/components/Filter/Facets/CollectingEvent/FacetCollectingEvent/ByAttribute.vue
@@ -41,7 +41,7 @@
       />
       Or
     </label>
-    <span class="small-text margin-small-left">results from different rows</span>
+    <span class="small-text margin-small-left">results from different groups</span>
   </div>
   <table
     v-if="selectedFields.length"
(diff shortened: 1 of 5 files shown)


kleintom:
TODOs for Tom's memory:
* [ ] any predicate setting doesn't survive reload
* [ ] I think there's one of the combinations, with no value maybe? that we aren't actually filtering on yet
* [ ] For the By Attribute facet, allow reuse of predicates with different values

Issue #1451, opened 2020-04-23
Model - TaxonName - Auto destroy OTU !is_in_use?


mjy:
When I destroy a TaxonName, and it has a "blank" OTU, then destroy that OTU.

* [ ] Ignore matrix row settings

PR #4728, opened 2026-01-30
Delete OTU with TN when OTU is "empty" #1451

kleintom:


app/controllers/taxon_names_controller.rb
```diff
@@ -73,6 +73,28 @@ def update
   def destroy
     parent_id = @taxon_name.parent_id
     @taxon_name.destroy
+
+    if !@taxon_name.destroyed? &&
```
mjy: Logic needs to be in the model, not here.  All of destroy should be wrapped in a transaction, the easiest way to do this is to put it in the callback chain.

spec/controllers/taxon_names_controller_spec.rb
```diff
@@ -163,4 +163,92 @@
 
     include_examples 'DELETE #destroy', TaxonName
   end
+
```
mjy: We almost never write controller specs, this needs to be model.

PR #4754, opened 2026-02-13
Cached map processing updates for memory usage and speed

kleintom:
The main changes:
* As @LocoDelAssembly guessed, `Parallel` instantiates its entire input (even when it's `find_each`) prior to forking, so we now just pass arrays of ids rather than full ARs
* We pass a `context` object (loaded off a bulk query) to CachedMapItem#stubs when in batch mode to avoid extra per-item queries
* We add registration items to a list in bulk mode and then insert them all at once instead of per-item

Run on 3i, ~1.5hrs
image

### DIFF
--- a/lib/tasks/maintenance/cached/maps.rake
+++ b/lib/tasks/maintenance/cached/maps.rake
@@ -190,8 +190,17 @@ namespace :tw do
             level1_geographic_name: nil,
             level2_geographic_name: nil
           ).distinct
+          untranslated_items = CachedMapItem.where(
+            level0_geographic_name: nil,
+            level1_geographic_name: nil,
+            level2_geographic_name: nil,
+            untranslated: true
+          )
 
           puts "Labelling #{items.count} CachedMapItems."
+          if untranslated_items.exists?
+            puts "Skipping #{untranslated_items.count} untranslated CachedMapItem rows (#{untranslated_items.distinct.count(:geographic_item_id)} geographic_item_ids) because untranslated rows do not receive geographic name labels."
+          end
 
           Parallel.each(items.each, progress: 'labelling_geographic_items', in_processes: cached_rebuild_processes ) do |o|
 


lib/tasks/maintenance/cached/maps.rake
```diff
@@ -176,19 +172,65 @@ namespace :tw do
           # TODO: this doesn't currently account for FOs
           q = Georeference.joins(:otus).where.missing(:cached_map_register).distinct
```
mjy: We don't want all, just the preferred: `Georeference.where(position: 1)....`

lib/tasks/maintenance/cached/maps.rake
```diff
@@ -176,19 +172,65 @@ namespace :tw do
...
-                reconnected ||= Georeference.connection.reconnect! || true # https://github.com/grosser/parallel
-                g.send(:create_cached_map_items, true)
+              reconnected ||= Georeference.connection.reconnect! || true # https://github.com/grosser/parallel
+
+              # Pre-compute OTU context for the slice in 1 query instead of
+              # 2 queries per georeference (N+1 elimination).
+              # Chain: georeference -> collecting_event -> collection_objects
+              #        -> taxon_determinations (position 1) -> otu
+              otu_rows = Otu
+                .joins(collection_objects: [:collecting_event])
```
mjy: with FO should be `joins(:collecting_events)` I think.

lib/tasks/maintenance/cached/maps.rake
```diff
@@ -176,19 +172,65 @@ namespace :tw do
...
+                  'otus.taxon_name_id'
+                )
+
+              # Group by collecting_event_id since georeference belongs_to
+              # collecting_event - hash lookup replaces per-record queries.
+              # Only include OTUs that have a taxon_name_id — OTUs without
+              # taxon names have no hierarchy and don't contribute to cached maps.
+              ce_otu_lookup = {}
+              otu_rows.each do |ce_id, otu_id, taxon_name_id|
+                next unless taxon_name_id
```
mjy: Just exclude earlier in sql no need to reference taxon_name_id, `Otu.with_taxon_name_id`

lib/tasks/maintenance/cached/maps.rake
kleintom: Done, though - minor issue - I think there's an open issue (or maybe just a code comment?) for being able to create a multi-polygon georef. So like a group of islands wouldn't show up in cached maps, just whichever island was added first.

kleintom:
Latest changes since the last sandpaper run that went over a day:

### Translations
* switched translation computation back to 2 queries as a defensive move to keep the planner from building in the wrong order and regressing performance (it's been challenging to keep that from happening, though that (probably) wasn't the issue with the long translations in the previous run)
  * locally the only translation query over 10 minutes is for Greenland - *not* related to anti-meridian issues, but maybe to high-latitude and large-ish vertex-count issues, even though it only bounding-box-intersects 6 shapes :/ We'll want to figure this out at some point. We now log these long translations so we should know exactly what to focus on soon.
* Discovered that production is on an older PostGIS/GEOS version than I am locally, which (apparently) explains why ADs in my local DB were running without any intersection errors but had intersection errors in production.
  * Workaround here is to cut out the erroring part of the translation query that causes those GEOS errors, when we catch such an error during translation (i.e. we run twice, with different queries, for those shapes like Russia's GI 30449
```
  - Ring 68: vertices at lon = -179.9999999999999 (just barely west of the antimeridian), latitudes 65.07 to 68.98                                                                                                                                                                                                                                                                                                                          
  - Ring 160: vertices at lon = 180.0 (exactly on the antimeridian), latitudes 65.07 to 68.98
  - Ring 92/94: same pattern near latitudes 70.99 to 71.54  
```
The part of the query removed was added last year to avoid small holes in maps caused by large buffers during the translation query reducing small states to nothing - if that happens here we'll live with it until production GEOS version has caught up to one that doesn't have this issue.


### AD Caching
* Avoid re-calling `q_ga` per-slice in CMI batch creation
  * Best guess as to what caused the massive slowdown in this query, though from my local runs it only explains like 4 hours longer, not 20 [it does reveal a bad planner flip when slice count passes 4k or so ... maybe it was worse on production on a different PG version?? Don't know]
* Switched AD caching to a bulk PG upsert rather than 1-by-1 ruby processing. (This follows what's becoming a very familiar pattern: avoid AR and ruby and move to SQL if you want performance. In this case the logic is so simple that I (personally) don't feel bad that we now have a ruby path for single AD processing and a bulk SQL path for rake processing.) This took local AD processing from 1 hour to 30 seconds.
  * The upsert should make us concurrency safe (between Parallel forks)
  * The new path relies on a join to existing translations - we no longer do any translation creation during bulk caching tasks (if it failed during the translation task it's just going to fail again during caching).

### Georef Caching
* Georef caching now singles out the "simple"/fast case where the georef is a point (thanks to @mjy for the idea) and only does the slower path when not a point.
* We now build maps for all georefs through FOs (as long as they're related to an OTU of course) in addition to COs [still need to remove the old Georef otus association that only goes through COs... it bit me here]
* We now only process position=1 georefs (per Matt)

### Migrations
* **There are two new uniqueness indices** on CMIs and CMRs for upserts.

kleintom:
I'll make some more changes here for tomorrow (just as an FYI).

kleintom:
Fixed a couple issues related to idempotency (re-running tasks that have already been run).

*Now* it's ready!

lib/tasks/maintenance/cached/maps.rake
```diff
@@ -47,6 +47,100 @@ namespace :tw do
       #
       namespace :maps do
 
+        def format_elapsed(start_time)
```
mjy, line 50: Could be a lib/utilities/time(?) method

lib/tasks/maintenance/cached/maps.rake
```diff
@@ -173,181 +273,408 @@ namespace :tw do
...
+                ON default_gagi.geographic_area_id = ga.id
+              WHERE
+                ad.asserted_distribution_shape_type = 'GeographicArea'
+                AND COALESCE(ad.is_absent, FALSE) = FALSE
+                AND cmr.id IS NULL
+                AND otus.taxon_name_id IS NOT NULL
+            SQL
+
+            gz_base_sql = <<~SQL.squish
+              SELECT
```
mjy, line 481: @kleintom This is fine for now, but don't forget you can use native AR composition then `.to_sql` to get raw strings as well.  Most of these seem like they are mostly composable in that way.

lib/tasks/maintenance/cached/maps.rake
```diff
@@ -173,181 +273,408 @@ namespace :tw do
...
+            translation_elapsed_ms = ((Time.current - translation_start) * 1000).round(1)
+            if translation_elapsed_ms >= slow_translation_ms
+              context = translation_failure_context(geographic_item_id, geographic_area_based)
+              puts " SLOW translation geographic_item_id:#{geographic_item_id} geographic_area_based:#{geographic_area_based} elapsed_ms:#{translation_elapsed_ms} threshold_ms:#{slow_translation_ms} total_matching_ads:#{context[:total_matching_ads]} sample_asserted_distribution_ids:#{context[:sample_ad_ids].join(',')} sample_project_ids:#{context[:sample_project_ids].join(',')}"
+            end
+          rescue ActiveRecord::StatementInvalid, ActiveRecord::RecordNotFound => e
+            context = translation_failure_context(geographic_item_id, geographic_area_based)
+            puts " FAILED translation geographic_item_id:#{geographic_item_id} geographic_area_based:#{geographic_area_based} elapsed_ms:#{translation_elapsed_ms} total_matching_ads:#{context[:total_matching_ads]} sample_asserted_distribution_ids:#{context[:sample_ad_ids].join(',')} sample_project_ids:#{context[:sample_project_ids].join(',')} #{e.to_s.gsub(/\n/, '')}"
```
mjy, line 674: If we are screen logging and we want to get fancy we can also do Rainbow('string').blue.bold stuff.

PR #4766, opened 2026-02-20
4716 BiologicalAssociation quick form AnatomicalParts

kleintom:
* The BA slice operates exactly as it does today when the AnatomicalPart box is not checked:
image

When the (saved-on-session) AP box is checked, you get two more AP checkboxes, one for subject of the BA, one for object of the BA. When you check either one of those, you have the option to create an AP on the source, or an AP on a selected OTU/CO/FO as the related:
image

image

The AP names you're given to select from are all those that currently exist in the project - the standard blue color comes from APs with a name, the other color is APs with a URI/URI_label (the URI is shown as a tooltip). [Note there can be duplicate AP names of either type, but we don't track ids here, only the name is being used.]

* If the source or object is a CO and you say you want to put an AP on it, you'll be required to select a TaxonDetermination for the CO
* if the source or object is an FO, you'll be allowed to specify if the AP on the FO `is_material`.

The object radial count (the pie count) only counts BAs *on the source of the radial*, not ones you create from an AP on that source. The table that previously existed in the slice is still there, and shows the same BAs as it did before: those with that source as the subject of the BA.

To see BAs you've created that have an AP on that source as the subject of the BA, there's now a button to click to display a larger modal table:
image
image

The original slice matched the information you were entering for your BA with existing BAs on the source of the slice - the new slice does the same, expanded to BAs with subject an AP on the source of the slice. So creating a "new" BA with source an AP, that already exists in the system, but with a new citation, will not create a new BA, it will just add the Citation to the existing BA (as expected, following old non-AP behavior).

image


app/controllers/biological_associations_controller.rb
```diff
@@ -45,14 +45,23 @@ def list
   # POST /biological_associations
   # POST /biological_associations.json
   def create
-    @biological_association = BiologicalAssociation.new(biological_association_params)
+    if create_with_anatomical_parts?
+      service = ::BiologicalAssociations::CreateWithAnatomicalParts.new(biological_association_params)
+      success = service.call
```
mjy: Feels like a new paradigm being introduced here.  If we the service always executes then we could chain it to new, and not need .call.  not sure.

app/controllers/biological_associations_controller.rb
```diff
@@ -45,14 +45,23 @@ def list
...
+      @biological_association = service.biological_association
+    else
+      @biological_association = BiologicalAssociation.new(biological_association_params)
+      success = @biological_association.save
+    end
+
     respond_to do |format|
-      if @biological_association.save
+      if success
```
mjy: More native is something like this?

```
@biological_asssocation.save
if biological_association.valid? 
...
end

File: app/models/anatomical_part.rb [hunk 3]
```diff
@@ -163,6 +163,31 @@ def self.used_recently(user_id, project_id, used_on = '')
     end
   end
 
+  # @return [Boolean] default is_material value for a new AnatomicalPart given its origin
+  def self.default_is_material_for(origin)
+    origin.class.base_class.name == 'CollectionObject'
+  end
+
+  # @return [Array] unique name and URI templates from this project, sorted by label
+  def self.templates(project_id)
```
mjy, line 173: Might be a helper method.

claude/log/4716_ba_ap_selection_2026-02-19_review.md
```diff
@@ -0,0 +1,84 @@
+# Branch Review: 4716_ba_ap_selection
```
mjy: We intending to commit this log?  Probably not?

mjy:
Just the log commit is blocking this I think @kleintom.

kleintom:
TODO followups: merge this work with the New Biological Assocations II task.
* this branch should probably switch to BAII's AP selector 
  * probably includes switching to BAII's endpoint for fetching project AP names
* BAII should probably switch to this branch's call to create AP and BA at the same time in a single transaction on the server

Probably also:
Add pinning for subject and/or object APs.

PR #4798, opened 2026-03-10
De-dup geometry values in OTU geojson helper returns

kleintom:
Note we need to still return all OTUs, we just don't need to return geometries more than once.

Taxonpages and Browse OTU were already de-duping client side and visually/functionally appear unaffected by this change.

AI code and summary:

  Benchmark across 5 OSF OTUs (before → after):

  | OTU    | Features | Nulls | UniqueGeom | Before bytes | After bytes | Reduction |
  |--------|----------|-------|------------|--------------|-------------|-----------|
  | 812683 |      501 |   214 |        263 |   13,432,967 |   6,101,702 |      -55% |
  | 837005 |      135 |    80 |         43 |    2,955,659 |     631,174 |      -79% |
  | 849801 |      224 |    59 |        133 |    3,889,286 |   1,773,223 |      -54% |
  | 849851 |      155 |    45 |         78 |    1,735,797 |     518,953 |      -70% |
  | 926534 |      190 |    61 |         99 |    5,853,542 |   3,273,142 |      -44% |
  | **Total** | **1205** | **459** | **616** | **27,867,251** | **12,298,194** | **-56%** |

  UniqueGeom is the count of distinct geometry strings among non-null features.
  Where UniqueGeom < (Features - Nulls), remaining duplicate geometries are
  cross-category (e.g. the same geographic area cited as both an asserted
  distribution and via a collection object's collecting event). These are not
  deduped as different categories render in different colors.

  Note: visually overlapping shapes can still be legitimate. For example,
  Austria appears as both a TDWG Level 3 and TDWG Level 4 geographic area;
  two records citing different gazetteer levels produce two distinct shape keys
  and both geometries are correctly retained.

### DIFF
--- a/app/helpers/asserted_distributions_helper.rb
+++ b/app/helpers/asserted_distributions_helper.rb
@@ -84,13 +84,13 @@ module AssertedDistributionsHelper
   end
 
   # @return [Hash] GeoJSON feature
-  def asserted_distribution_to_geo_json_feature(asserted_distribution)
+  def asserted_distribution_to_geo_json_feature(asserted_distribution, skip_geometry: false)
     return nil if asserted_distribution.nil?
     return nil unless asserted_distribution.has_shape?
 
     return {
       'type' => 'Feature',
-      'geometry' => RGeo::GeoJSON.encode(asserted_distribution.geo_object), # TODO: optimize
+      'geometry' => skip_geometry ? nil : RGeo::GeoJSON.encode(asserted_distribution.geo_object), # TODO: optimize
       'properties' => {
         'base' => {
           'type' => 'AssertedDistribution',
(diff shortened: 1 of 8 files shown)


kleintom:
@jlpereira could you please give this a try with Browse OTU and TaxonPages in case there's some functionality I might have broken that I don't know about?

The only change is that more (duplicated) `geometry` keys in geojson hashes have null values now.  The client-side de-duping you've been doing *shouldn't* be necessary anymore, but maybe we keep it for now? Whatever you think.

kleintom:
As the commit message notes, we're not de-duping shapes across categories: e.g. the same shape can show up as an AD shape (orange) and a type shape (blue). It would be possible to combine those but it's not clear it's worth it, at least for now.

PR #4815, opened 2026-03-13
4764 Update backend code to add OTU data through Asserted Distributions that aren't based directly on OTUs

kleintom:
Specifically we now associate OTUs via Asserted Distributions on
* Biological Associations with OTU subject and/or object
* Biological Association Graphs with vertex BA having an OTU subject and/or object
* depictions/conveyances/observations that have an OTU as object

We also create cached maps via those ADs for such OTUs now and process them in rake indexing.

### DIFF
--- a/spec/helpers/otus_helper_spec.rb
+++ b/spec/helpers/otus_helper_spec.rb
@@ -318,63 +318,6 @@ describe OtusHelper, type: :helper do
           expect(features.count { |f| f['geometry'].nil? }).to eq(1)
         end
 
-        specify 'BiologicalAssociationsGraph AD on same shape is deduplicated' do
-          br = FactoryBot.create(:valid_biological_relationship)
-          ba = BiologicalAssociation.create!(
-            biological_association_subject: otu3,
-            biological_association_object:  otu3,
-            biological_relationship: br,
-            project: otu3.project
-          )
-          bag = FactoryBot.create(:valid_biological_associations_graph)
-          FactoryBot.create(:valid_biological_associations_biological_associations_graph,
-            biological_associations_graph: bag,
-            biological_association: ba)
-          FactoryBot.create(:valid_biological_associations_graph_asserted_distribution,
-            asserted_distribution_object: bag,
-            asserted_distribution_shape: shared_ga2)
-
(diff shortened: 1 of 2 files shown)


app/models/asserted_distribution.rb
```diff
@@ -186,6 +186,41 @@ def otu
...
+      ids = []
+      ids << ba.biological_association_subject_id if ba.biological_association_subject_type == 'Otu'
+      ids << ba.biological_association_object_id if ba.biological_association_object_type == 'Otu'
+      ids.uniq
+    when 'BiologicalAssociationsGraph'
+      asserted_distribution_object.biological_associations.flat_map do |ba|
+        ids = []
+        ids << ba.biological_association_subject_id if ba.biological_association_subject_type == 'Otu'
+        ids << ba.biological_association_object_id if ba.biological_association_object_type == 'Otu'
+        ids.uniq
```
mjy, line 208: This doesn't look correct, why 2 .uniq.  This is not returning the ids, but the uniq ids of the loop.

lib/tasks/maintenance/cached/maps.rake
```diff
@@ -652,6 +649,90 @@ namespace :tw do
...
+              AND ad.asserted_distribution_object_id = bag.id
+            JOIN biological_associations_biological_associations_graphs babag
+              ON babag.biological_associations_graph_id = bag.id
+            JOIN biological_associations ba
+              ON ba.id = babag.biological_association_id
+              AND ba.biological_association_subject_type = 'Otu'
+
+            UNION
+
+            SELECT ad.id AS ad_id, ba.biological_association_object_id AS otu_id
```
mjy, line 696: These feel like compasable as ActiveRecord.  Too much raw SQL here.

app/models/asserted_distribution.rb
kleintom: `ids.uniq` is to cover the case where the subject and object otu id are the same, but yah since the outer loop has `end.uniq` we can dispense with `ids.uniq`, I'll fix.

mjy: Why are the `ids = []` and logic in that loop at all then?

Issue #4823, opened 2026-03-17
[Bug]: Filter OTU on Gazetteer union of Geographic Areas can return fewer (sometimes 0) results than the GAs


kleintom:
### Steps to reproduce the bug

```bash
1. In the Afrotropical Bees project, create a union Gazetteer from Namibia and South Africa
(from the shapes with most uses).
--> Note the gazetteer does get created, even though it returns a toast error 'Linear rings must have 0 or >= 4 points'. The saved shape looks basically fine aside from some slivers between the borders, which is not unexpected (but maybe the cause of the error).
2. Filter OTU on the South Africa GA --> 1311 results
Filter OTU on the Namibia GA --> 374 results
Filter OTU on the South Africa + Namibia Gazetter --> 0 results
```

### Screenshot

<img width="512" height="300" alt="Image" src="https://github.com/user-attachments/assets/a098960d-eff8-4a1b-b6be-314189d7534f" />

### Environment

Production

proceps:
Not sure how it works, but if in the OTU filter, I put both a gazetteer entry and both of the countries (S Africa and Namibia), I can filter 1400 recoreds, presumably species from both S Africa and Namibia. But if I filter without gazetteer entry, the result is about 300, the species which are reported in both countries (AND filterring). 
The OTU filter itself (as well as CO filter and AD filter), should be updated to allow both AND and OR searches.

kleintom:
This will be fixed in the next release: all *new* unions going forward should be correct, some already created unions may still be affected.

PR #4840, opened 2026-03-20
4823 gaz union imprecision

kleintom:
A lot of words for not much code.

See #4823 for a graphic of what caused the issue (in short, floating point rounding issues between a shared boundary of two shapes where the shared vertices should be exactly the same but are barely not: that leading to border slivers in the union that cause that union to not necessarily contain its constituent shape(s)).

The fix here is to buffer the union by `1e-7 degrees` (in geometry (not geography) space), which is about 11mm at the equator. [Another fix considered was this "trick" of buffer by +epsilon and then buffer by -epsilon, which essentially removes slivers of width less than epsilon, and it *does* remove the slivers we want removed, but it also comes with its own rounding errors that mess up containment on the outer boundary of the union instead.]
See the long commit message in the first commit for the (ai-generated) test and results that suggested this would be a good value. Here's the summary:
```
========================================================================
Summary across all test cases
========================================================================
Total inputs tested:                 37
Inputs failing without buffer:       19
Largest vertex error observed:       8.517e-05 degrees [ignore this, it's the result of a not tight enough query]

Per-buffer-value results (all test cases combined):
  Buffer (°)    Inputs passing / total
  none           18/37
  1e-12          31/37
  1e-11          ✓ all pass
  1e-10          ✓ all pass
  1e-09          ✓ all pass
  1e-08          ✓ all pass
  1e-07          ✓ all pass
  1e-06          ✓ all pass
  1e-05          ✓ all pass

Minimum buffer that fixes ALL test cases: 1.0e-11 degrees
Production value (1.0e-07) is 4 order(s) of magnitude above minimum
```

Since existing, already created gazetteer unions could also have this issue, **this pr includes a migration to apply the 1e-7 buffer to *all* gazetteers already in the system.**

The other issue considered here was the 'linear rings must have >0 and <=4 vertices' error when the SA+Namibia union was saved. That appears to have been from the old rgeo union operation (we now use postgis for the union); best guess is that it was coming from preview - but not save - going through a geojson parse (or something) that caused those slivers to again barely change coordinates and lead to invalid rings. (Preview was being called immediately after save.) In any case the preview call on save has been removed, we're now buffering those unions to make them more valid, and we've added an `ST_MakeValid` on the result of the buffer call for good measure (maybe not needed? docs don't state that ST_Union output is guaranteed valid).

For similar reasons we now *shrink* intersections by 1e-7, though this case is far less frequent and useful for spatial queries.

Some notes:
* **`ST_Buffer` drops z-values, so we can't support z-values on unions with this new code.** (There are many other places where we use ST_Buffer, but this might be the only place where we save the result(?).) It's "fixed" here by running `ST_Force3d` on the result, which returns z=0 to all coordinates before saving.
* buffering does increase the number of vertices of the unioned shapes (but the `&&` pre-filter operator we do in spatial queries seems to make that not matter much, at least on the original SA+Namibia union)
* this fix seems to work on the original case and on around a dozen other adjacent shapes the Filter OTU results were tested on (see the long commit message of the last commit here), *but*:
  * note that filtering on the union can *correctly* return fewer results than the sum of the result counts on its parts, since the same OTU can have an AD on each member of the union, e.g. [It's probably not so easy to verify as a user that the results are correct, for this reason.] All one can say for sure is that the union count should be at least as large as each result count of the components of the union.
  * as we've been discovering in researching geoboundaries, our states are not always contained in their countries, our counties are not always contained in their states. This is another source of counts mismatches: when you go to the union, now the union of two states may contain a county that previously was not contained by either component of the union separately, and so now you're seeing larger counts than expected for the union. That's the way it is for now; we're discussing ways to improve this in the move to geoboundaries.


mjy:
> ST_Buffer drops z-values, so we can't support z-values on unions with this new code. (There are many other places where we use ST_Buffer, but this might be the only place where we save the result(?).) It's "fixed" here by running ST_Force3d on the result, which returns z=0 to all coordinates before saving.

Being aware of this will be important.  Please add a comment on GeographicItem reflecting this.  We don't support Z, it was always "if needed", so for now it's fine.

Otherwise - I agree with "merge without migration".   Probably need a community discussion earlier than later on their being OK with migration.

PR #4869, opened 2026-04-06
4383 Georef from Gazetteer shape

kleintom:
* Adds the ability to select a Gazetteer whose shape should be used to create a Georeference:
image

* It can be much more common for a georef to have many (many) vertices for its shape, so we now truncate all UI references to DWC's `footprintWKT` value, while still returning the full value for API calls and dwc exports (is that what we want?).

* Since Gaz shapes can be (generally are) much larger than current typical Georeference shapes, we now destroy Georef shapes (including error shapes, not that those really need it) once they're no longer being used.
  * 'being used' is determined by ApplicationEnumeration, the same as OTU-being-used is computed in the Taxon Name case.
  * Tightened up the existing Gazetteer version of this
    * No longer allow Gaz to be deleted when referenced by an AD

* Fixed up GeoJSON serialization for display in the Georef panel (see screenshot `Coordinates` column) to support all shape types, including GeometryCollection
  * Added a js spec file for this, even though it won't run in CI - cc @jlpereira, happy to do this some other way
    * Use the new serialization code in Gaz display (formerly had its own code)

Ready for review.


jlpereira:
Add a spec file sounds really good to me, we started an implementation for this some ages ago but we close it 

https://github.com/SpeciesFileGroup/taxonworks/pull/3037

mjy:
This looks great.  Merge away.

PR #4884, opened 2026-04-20
4881 cross-project contamination followup

kleintom:
This adds two new safeguards to help prevent cross-assignment of project ids:
* on `find_or_create_by` and similar calls on *models* (not on relations), we now inject `project_id` where relevant and not already present. This prevents find errors like `Otu.find_or_create_by(name: 'other_project_name')`.
  * this really protects the `find` part of `find_or_create_by` - in the `create` case we already insert `Current.project_id` if project_id isn't already present.
  * if we want to make `project_id` required we could (run-time) throw instead, but in practice we're often doing something like `Otu.find_or_create_by(taxon_name_id:)` where we already know the taxon_name_id is in the correct project (this is considered safe) - plus this is consistent with auto-insertion of `Current.project_id` on create.
  * This *DOES NOT* guarantee `project_id` is set on all relations, only on *models* - relations would be much trickier. Coders *must still practice manual review of all queries* as they relate to this issue; personally I would still advocate for `project_id` as a required param in edge cases, rather than relying on this new guard (as, e.g. in `Tag::batch_create` where I want to feel certain). It's too easy to not use it in the model case where it maybe technically isn't needed and then not use it in the relation case where it may be absolutely necessary (for the `find` part at least).
    * Using `project_id` in the maybe-not-necessary cases also protects against bugs where data we thought was safe turns out to not be so because of a bug (like last week's case)
* On *every* create and update on project-associated models, we now check that any `belongs_to` data belongs to the same project as the parent. This prevents errors like the one we saw last week: `Otu.create!(taxon_name_id: id_from_a_different_project)`.
  * the price is the `project_id` lookup for each `belongs_to`, on every save and update. (We could probably check `will_save_change_to_attribute?` on updates to save queries - and we can if you want - I'm just not currently feeling trusting about that bit, for no reason I can remember.)


mjy:
Thanks @kleintom, let's merge.

Issue #4445, opened 2025-06-18
Unify projects (backend)


mjy:
As a user I want to unify two projects.

PR #4886, opened 2026-04-21
4445 unify projects

mjy:
#4445 From a early Claude prompt, beware.  Do not merge.


app/models/taxon_determination.rb
```diff
@@ -151,6 +151,44 @@ def dwc_occurrences
...
+  #
+  # We bypass acts_as_list entirely via update_columns and restore the record's
+  # original position. If that position is already occupied (by the earlier-ID
+  # sibling that was reset to 1 by acts_as_list), we bump it to the end to make
+  # room — preserving the semantically important position-1 accepted
+  # determination.
+  #
+  # @param target_project_id [Integer]
+  # @return [true] signals the migrator that the record is already persisted
+  def handle_unify_conflict(target_project_id)
```
mjy: How can we have conflict?  All merged data have their own OTUs, and therefor TDs are isolated to those OTUs.

app/models/taxon_name.rb
```diff
@@ -270,8 +270,8 @@ def reload(*)
 
   # Rails 7 experiments have after_commit creating a whack-a-mole situation
   # (though leave after_commit on TaxonNameRelationship)
-  after_commit :set_cached, unless: Proc.new {|n| n.no_cached || errors.any? }
-  after_commit :set_cached_warnings, if: Proc.new {|n| n.no_cached } # Should definitely be after commit
+  after_commit :set_cached, unless: Proc.new {|n| n.no_cached || n.errors.any? || Thread.current[:tw_taxon_name_no_cached] }
```
mjy: @kleintom Can you isolate Thread.current to it's own method, so we can change internals consistently if needed?

db/migrate/20260512104247_delete_duplicate_roles.rb
```diff
@@ -0,0 +1,33 @@
+class DeleteDuplicateRoles < ActiveRecord::Migration[8.1]
```
mjy: This is a bi-product of discovery, not a function of unify, correct?

db/migrate/20260512104248_delete_duplicate_taxon_name_relationships.rb
```diff
@@ -0,0 +1,184 @@
...
+  #                   PARTITION BY type, subject_taxon_name_id, object_taxon_name_id, project_id
+  #
+  # We keep the lowest-id record in each group and delete the rest.
+  #
+  # Safety check: raises if any higher-id duplicate has notes (not auto-moveable).
+  # Citations on higher-id duplicates are re-pointed to the surviving lower-id record
+  # (skipping any source already cited there) before the duplicate is deleted.
+  def up
+    transaction do
+      duplicates_with_notes = execute(<<~SQL).to_a
```
mjy: Would prefer to see this, and the other method, be isolated into a maintainence library, and called from here.

lib/housekeeping/projects.rb
```diff
@@ -97,6 +97,7 @@ def is_community?
   private
 
   def project_foreign_keys_are_in_same_project
+    return if Thread.current[:tw_project_unification]
```
mjy: See comment on (very) base method call above.

app/models/taxon_determination.rb
kleintom: The general project_id change doesn't preserve `position`: when you move the first TD (which may not be position 1) to the new project it automatically gets position 1, and then later you may try to move the actual position 1 and get a conflict.

lib/project_unification/cached_rebuilder.rb
```diff
@@ -0,0 +1,108 @@
+# Rebuilds cached fields after project unification
+#
+# Many models cache computed values for performance. After bulk migration,
```
mjy, line 3: Why recalculatated?  No new relationships from the other project are directly related to any object being moved over, I'm not sure how composition would change?

db/migrate/20260512104247_delete_duplicate_roles.rb
kleintom: The two migrations are for issue #4920 cleanup that we need to have cleaned up in order for project unify to be able to run.

lib/project_unification/migrator.rb
```diff
@@ -0,0 +1,246 @@
+# Handles the actual data migration between projects.
+#
+# Processes models in *reverse* MANIFEST order using two strategies:
+#   fast  — bulk SQL UPDATE, no per-record validation
+#   slow  — per-record valid? to detect target-project uniqueness conflicts,
+#            then save!(validate: false) on success
+#
+# !! A note on somewhat implicit assumptions as we move objects one at a time
+# from source to target project:
+# * we run in reverse manifest order: that means that when we validate-save in
```
mjy, line 10: In its composition I have not thought hard about reverse manifest order and what that means.  Just FYI.

lib/project_unification/model_classifier.rb
```diff
@@ -0,0 +1,150 @@
...
+# Slow track: per-record valid? check to detect uniqueness conflicts against the
+#   target project, then save!(validate: false) on success.
+#
+module ProjectUnification
+  class ModelClassifier
+    # Bulk SQL UPDATE — no per-record validation. Safe only when the model has
+    # no project-scoped uniqueness validators, OR when every such validator's
+    # scope includes a FK to a project-specific model (making cross-project
+    # collisions impossible). See model_classifier_spec.rb for the tripwire.
+    FAST_TRACK = %w[
```
mjy, line 13: We should have a spec, if we don't, that MANIFEST is completely covered by sum(TRACKS)

lib/project_unification/cached_rebuilder.rb
kleintom: That was my conclusion as well, but I didn't feel confident enough in the TN case, where we do change the parent of the top names (see my TODO below). If you think that's not a concern I'm happy to remove this code.

lib/project_unification/migrator.rb
kleintom: Thanks, yah, I think in theory it could matter if there are validations of a particular type (my "are there any such??" comment here), but in practice I don't think it does. It's not hurting anything though, so I left it.

lib/project_unification/model_classifier.rb
kleintom: Yup, it's there.

db/migrate/20260512104248_delete_duplicate_taxon_name_relationships.rb
kleintom: Is it a concern that we'd be calling a lib function that could change later (if the models change) from a migration where the data being migrated would stay forever on today's model defns?

mjy: More the idea that we'd screw up and need to run this again?

lib/project_unification/cached_rebuilder.rb
mjy: OK.  Let's just comment to this effect and leave it in I think?

Issue #4888, opened 2026-04-21
Remove TW user ids from api output


kleintom:
From our gitter conversation my understanding of where we left off was:
* José: The filters are exposed in the API and could filter through housekeeping
* Matt: We should not provide `user_id` in the `api/v1`.  Created/updated at are useful, we could generlize those.
 Good point about the housekeeping exposure

So the question is if/how to include housekeeping like `user_id`s in api output, presuming some people find it useful for filtering.

I'll post a preliminary PR in a minute to scope this out a little more.

PR #4889, opened 2026-04-21
Remove name ids and names from api output

kleintom:
@jlpereira @mjy Kind of throwing this out here (untested) so we can all see the scope of the issue. 

If we just take this pr as it currently stands, we'll be removing *all* user ids and names from public api output, except for the following dwc fields (which go to GBIF if/when a project goes to GBIF (I think)):
* georeferencedBy, recordedBy, identifiedBy

IIRC there are a bunch of api endpoints that can do filtering.


kleintom:
Issue #4888

mjy:
I think my vote is to go for it.  @jlpereira it might break some things?  

The one place we might want to consider keeping name is on News API?  So we can credit the person?  Ulitmately  we could use Roles there for that though- maybe we should waight for that.

jlpereira:
I think it's fine, at least we're not using user_ids in TP. For news endpoint we already provide creator and updater names.

mjy:
Let's merge on green @kleintom

PR #4890, opened 2026-04-22
Attempt to fix up some of our intermittent spec failures

kleintom:
Essentially all the digging was done by claude/codex ai (with some back and forth commit-wise where I balked at too-broad changes); clear causes and fixes in some cases, solid mitigation in others.

Adds a script at `bin/random_spec_failures` that checks gh ci logs over the past n days to generate an html report of recent "random" failures.

Planning on running in ci a few times and then landing if all still looks good.

### DIFF
--- a/spec/features/tasks/sources/verbatim_author_year_source_spec.rb
+++ b/spec/features/tasks/sources/verbatim_author_year_source_spec.rb
@@ -8,11 +8,18 @@ describe 'Verbatim author year to source', type: :feature, group: :sources do
     context 'when I visit the task page' do
       before { visit verbatim_author_year_source_task_path }
 
-      xspecify 'page loads without error' do
-        expect(page).to have_text('Verbatim author/year to Source')
+      specify 'page loads without error' do
+        expect(page).to have_text('TaxonName verbatim author/year without citations')
       end
 
-      context 'with taxon names having verbatim author and year' do
+      context 'with taxon names having verbatim author and year', js: true do
+        # Current must be set in the test process so that the parent_is_root factory
+        # trait can look up / create the root TaxonName via housekeeping callbacks.
+        before do
+          Current.user_id    = @user.id
+          Current.project_id = @project.id
+        end
+
         let!(:taxon_name1) {
           FactoryBot.create(:valid_protonym,
             verbatim_author: 'Smith',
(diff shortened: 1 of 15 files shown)


kleintom:
Some of the issues uncovered locally were likely because I either covered the running browser specs or switched to a different desktop, causing browser behavior to change enough to make things fail. I'm never going to just sit and watch local browser specs run, so I'm in favor of fixing those where we can, but content editor specs, e.g. - source of much flakiness - is now commented as not-expected-to-pass under those "occluded" conditions. Maybe we should have an easy headless option for running locally for this reason, and only switch to visible browser for debug?

kleintom:
The first run on this pr claiming to fix random failures had a random failure xD. The fail message was "woah there, login first", which is very familiar from failed spec runs - in this case it was on `otus_spec.rb`, but we have a record from `people_spec.rb` as well, both rack_test (non-js) feature specs.

Codex's theory, and one that sounds right to me, is that rack_test specs *can (and sometimes do) use multiple db connections during a single spec run* (even though rack_test is single-threaded, the requests rack_test makes to rails can cause rails to choose more than one connection from its pool of connections over the course of a spec. Experimenting with a pool of 1 connection (default/current is 5), which would probably also fix this issue, produced failures in real-browser/js specs.). That's clearly an issue with using the db cleaner's transaction strategy for those specs (as we were) because the spec can create the transaction and login under one connection, but then later in the same spec switch to a different connection with no knowledge of the transaction data, and boom: you're not logged in.

So we've now switched to truncate strategy for db cleanup between rack_test specs - we were already doing truncate on browser/js feature specs (presumably for the same reason).

Here's a timing comparison with the truncation change vs a spec run from yesterday:
```
Latest successful branch build 24786781071:

  - job 0: 14m 20s
  - job 1: 13m 55s
  - job 2: 14m 59s
  - job 3: 14m 24s
  - average: about 14m 25s

  Yesterday’s successful development build 24741823005:

  - job 0: 10m 10s
  - job 1: 12m 24s
  - job 2: 14m 42s
  - job 3: 14m 00s
  - average: about 12m 49s

  Comparison:

  - average test-job time is up by about 1m 36s
  - slowest shard went from 14m 42s to 14m 59s, so only about 17s slower at the top end
  - most of the average increase comes from the faster shards getting slower, especially job 0
```

Merging since testing seems stable (10/10 runs green), but of course welcome any other solutions (or theories) on this issue.

spec/features/tasks/sources/verbatim_author_year_source_spec.rb
```diff
@@ -8,11 +8,20 @@
...
+      specify 'page loads without error' do
+        expect(page).to have_text('TaxonName verbatim author/year without citations')
       end
 
-      context 'with taxon names having verbatim author and year' do
+      context 'with taxon names having verbatim author and year', js: true do
+        # Current must be set in the test process so that the parent_is_root factory
+        # trait can look up / create the root TaxonName via housekeeping callbacks.
+        before do
+          Current.user_id    = @user.id
```
mjy, line 19: This makes me nervous in feature specs, we want to ensure Current is set only via the login process here.  I'm not sure there is a precident for touching it elsewhere?

kleintom: I'll see if I can fix some other way or else re-xspecify. (I knew you were going to get me on this one :P :))

PR #4898, opened 2026-04-24
4828(2) batch update synonym type

kleintom:
This is a straight copy of https://github.com/SpeciesFileGroup/taxonworks/pull/4838 that I forgot to resubmit (sorry Geoff!) after I submitted the original without review on an api change. See that pr for notes.


app/javascript/vue/components/radials/taxon_name_relationship/components/SynonymTypeSlice.vue
```diff
@@ -0,0 +1,112 @@
...
+  </div>
+</template>
+
+<script setup>
+import PreviewBatch from '@/components/radials/shared/PreviewBatch.vue'
+import UpdateBatch from '@/components/radials/shared/UpdateBatch.vue'
+import { TaxonNameRelationship } from '@/routes/endpoints'
+import { ref, computed, onMounted } from 'vue'
+
+const SYNONYM_BASES = [
```
mjy, line 58: I think this should be discovered from an API call.

mjy:
@kleintom @jlpereira See my request to move the constant out of Vue.js to the backend, if we can get that tweaked this is good to merge IMO.

PR #4900, opened 2026-04-27
2141 Find all BAs for Browse OTU task, not just those that have the OTU as direct subject/object

kleintom:
* Now computes BAs for an OTU by finding:
  * otu as subject/object of a BA
  * otu as preferred (first) TD of a CO or FO that's a subject or object of a BA
  * otu as 'origin otu' of an Anatomical Part
* Cached maps now accounts for ADs on all such BAs for an otu
* Fixed n+1 issues for BA display in Browse OTU


mjy:
@kleintom Took a 3 minute pass through- LGTM given we have good spec coverage.  Like seeing slightly less giant raw SQL.

kleintom:
Thanks @mjy

PR #4901, opened 2026-04-27
4847 Move author name matching in New Source to Ruby, add more matching

kleintom:
This now works as follows, with 2 queries:
Query 1:
* today's (limited) matching --> very fast

Query 2: slower, but more extensive matching; only adds to the list of possible matches, never moves existing ones from the first query. All of the following should now be matched (IIRC...), along with alternate spellings:
* John Smith
* J Smith, J. Smith
* John-James Smith, John James Smith, Johnjames Smith
* J.J. Smith, J. J. Smith, J J Smith

Maiden names:
* Smith Jones, John
* Smith-Jones, John
* Smith, John
* Jones, John

The most difficult match claude could create locally took 0.4 seconds (matching against ~220k people) - currently *both* queries run on an author before the next author's queries are sent; locally that seems perfectly fine, but if it's too slow on production we can switch to run all of the fast queries first and then all of the slower queries.


kleintom:
@proceps Could you try this out to see if it fits your needs? Branch `4847_name_match_ruby`.

kleintom:
Oops, I included the ISBN fix on this branch by mistake, I'll remove before commit and resubmit that one as a separate PR (it doesn't impact this branch at all).

app/javascript/vue/tasks/sources/new_source/components/bibtex/AuthorMatchModal.vue
```diff
@@ -240,6 +261,48 @@ function normalizeName(name) {
   return (name || '').toLowerCase().trim().replace(/\s+/g, ' ')
 }
 
+// Score how closely a person's first name matches the parsed first name.
+// Parts are compared positionally: 2 points for exact word match, 1 for matching initial.
+function scoreFirstNameMatch(person, parsedFirstName) {
```
mjy: Not fully sure here- but this feels like more logic than we want on the client side?  How can we test this in isolation is maybe what I'm getting at.

kleintom: Yah, agreed, thanks - moved that functionality server-side and added specs.

kleintom:
The new commits
* move name ranking functionality server side, with specs
* match author strings to author roles by name (the old way [note this was never always possible, e.g. which of 'J Smith' and 'J L Smith' should match with 'John Smith' and 'Jack Smith'?]) and now also - once the modal has been re-opened and possible person matches have been reloaded on a row - by person id.
  * the new case is for when the person matched to an author string has an alternative spelling, e.g. which can't (easily) be matched to the original string.
* TODO: there's still a minor issue when the user opens the matcher, matches one name, closes, reopens and matches another name, etc: names appear in New Source in the order they were selected, not in the order they appear in the matcher.

mjy:
@jlpereira Can you look at 

"TODO: there's still a minor issue when the user opens the matcher, matches one name, closes, reopens and matches another name, etc: names appear in New Source in the order they were selected, not in the order they appear in the matcher."

We need to keep this order aligned I think or people will lose authors/order going back and forth from verbatim.

Issue #4701, opened 2026-01-19
Batch source upload vs in-project


kleintom:
* Given a tsv of source results obtained from some non-TW provider
  * I'll write a script that uses `api/v1/bibliography` to find out if each source is a) not in TW, b) in TW but not in my project, c) in TW and in my project
    * The script then converts my original source list into a bibtex list that can be batch imported by TW

The breakdown is that if a source is a) or c) then everything is great. In case b) where I know the source is in TW but not in my project, there's no automation path, I'll have to add each source to my project one at a time (or upload them all and then unify one at a time) (?).

The request is to be able to batch add existing sources to a project.

Best perhaps would be to allow (some kind(s) of) source matching on import, with the option to add existing not-in-project to project when there's a match. (In this case nobody would need to write a script, TW would do it for them...)

PR #4904, opened 2026-04-28
When creating a bibtex source, if the bibtex serial issn matches a db serial, use it #4701

kleintom:
Do so even if the serial title doesn't match - issn is a GUID, and there are many subtle title differences in TW serial titles entered by humans.

The scenario that used to cause mysterious batch import source errors was:
```
 1. Incoming BibTeX has journal title "X" and ISSN "123"
 2. DB has a Serial named "X" (name match, `s`) — but with no ISSN, or a different ISSN
 3. DB has a different Serial with ISSN "123" (identifier `i`) — but a different name

The old code required both conditions — s exists AND s == i.identifier_object — to set serial_id on the new source. With those two different serials in the DB, that equality check fails, serial_id stays blank, and the fallback tries to build a new Serial "X" with ISSN "123"... which already belongs to the other serial and so raises an error.
```

Now we assign the serial from 3 because its UUID issn matches the incoming serial's UUID issn (even though it doesn't match on serial title).

### DIFF
--- a/app/models/source/bibtex.rb
+++ b/app/models/source/bibtex.rb
@@ -689,11 +689,10 @@ class Source::Bibtex < Source
       s = Serial.where(name: journal).first
       i = Identifier::Global::Issn.where(identifier: value, identifier_object_type: 'Serial').first
 
-      # Found an Existing Serial identically named with an assigned Identical ISSN
-      if !s.nil? && (s == i&.identifier_object)
-        self[:serial_id] = s.id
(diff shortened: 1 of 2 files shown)


proceps:
I do not think it is good to create new serials using this way.
I would say, if serial does not match, just add verbatim journal, and live serial_id blank.

kleintom:
The ISSN of the serials match, it's just their titles that don't (which is frequently the case).

kleintom:
We're not creating a new serial here, we're using the existing one from the db that has the same issn as the incoming source that's being created.

PR #4905, opened 2026-04-30
Dropzone paste files with ctrl+v

kleintom:
Ai code.

Motivation: pasting image into depiction annotator slice or New Image, pasting pdf into documentation annotator slice or New Source.

The paste listener is on `document` (since we don't really have a way to focus our dropzone input without triggering the file dialog (?)) - as long as no "text" input is active on paste, Dropzone will check to see if the paste is a file and add it if so.

!! That means every dropzone on a page catches a paste !!, so we turned off dropzone-paste on the image matrix observations builder (the only case we know of with multiple dropzones on the same page).

Seems to work well on Chrome and for single-file pastes on Firefox (but firefox has a 13 year old bug for multiple pastes :( https://bugzilla.mozilla.org/show_bug.cgi?id=864052).

Dropzone seems to intentionally disable this functionality: 
https://github.com/dropzone/dropzone/blob/f50d1828ab5df79a76be00d1306cc320e39a27f4/src/dropzone.js#L372

(because of the firefox bug??)

@jlpereira (@mjy) any thoughts on this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This project <em>does not yet</em> adhere to [Semantic Versioning](https://semve
 - `type_taxon_name_relationship` support to the extend[] parameter in `inventory/summary.json`
 - Quick Forms: Pagination to Biological associations slice
 - CSD: Added a button to load the associated CO when the identifier is already in use by another CO
+- Ctrl+v to paste images and documents in "Drop [image/doc] here"
 
 ### Changed
 
--- a/app/javascript/vue/components/dropzone.vue
+++ b/app/javascript/vue/components/dropzone.vue
@@ -16,8 +16,8 @@ import {
   useTemplateRef,
   watch
 } from 'vue'
-
 import { isJSON } from '@/helpers/objects'
+import { useDropzonePasteManager } from '@/composables/useDropzonePasteManager'
 
 const props = defineProps({
   url: {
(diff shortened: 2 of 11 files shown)


jlpereira:
Tom, I like the idea but I have some doubts. 

Binding this to document could trigger all dropzones actives. Keep in mind that there are other tasks like comprehensive that has 2 dropzones(CO, CE) active all the time. And if this is going to work on a Radial annotator, depending the context of where is opened could trigger others dropzones. Perhaps make the prop false by default instead true.
Keep in mind that we use dropzone in rails views too

kleintom:
Thanks for the good notes José - the new version uses a composable to track the number of dropzones on a page, and you can mark particular dropzones with a prop to "prioritize" them. If there's only one dropzone on a page or there's only one prioritized dropzone, then that one gets the paste, otherwise paste is disabled.

Currently the only "prioritized" dropzones are the ones for the New tab of the documentation, depiction, and conveyance slices in radials (are there others?).

The only non-vue dropzone I'm seeing is for images/new, where the user is already pushed toward the New Image task anyway, so I didn't add paste for non-vue dropzones.

Reminder that you can copy/paste multiple files on Chrome but not Firefox (see above for the longstanding firefox bug on this issue).

jlpereira:
Sounds good to me, the other one that needs to prioritize is in CSD, we have two actives for CO and CE. I would prioritize CO, since I believe it's the most commonly used there. Another option is to display a modal window if there are two or more active dropzones, allowing the user to choose which one to add the images to. However, since this is the only possible scenario right now, I wouldn't implement it unless users specifically mentioned it.

kleintom:
Thanks José!

PR #4906, opened 2026-05-04
2723 iNaturalist import

kleintom:
Enter ids by number or url:
image

After submit you get a table of the ones you submitted, with a refresh button so you can watch progress - here with the first 3 created so far:
image

Currently limited to 50 at a time since the controller does a synchronous api lookup to get initial data.

One awkward part of our import: we assign the inat uuid for an observation as an identifier on the imported FO, but the uuid is only visible from the api on inat - it's not the same as the observation id in the standard 'show' url of an inat observation (though you can replace that oid with the observation's uuid and the link will still work).

I added an inat link on the inat identifier on the browse FO page:
image

And so that users can find out if an observation is already in TW without going through the inat api to find its UUID, there's a `Find` tab on the inat importer task (limit 200, which is inat's api limit for a single request):
image


kleintom:
Other notes:
* We assign uuids where appropriate/available on:
  * FO
  * taxon determination (only when from the uploader - the community id doesn't get a UUID)
  * images
  * sounds
* I added mp4 as a sound type to TW since we already had m4a, which apparently is Apple's extension for mp4 (ish)
* *Once you've imported an inat observation once, you'll be blocked from importing it again by the uuid identifier on the already imported FO*
  * we do *not* handle updating of an existing inat fo (which could see basically every piece of data except the uuid change)
  * even if you just didn't import images initially, and now you want to, you'll need to delete the existing FO (or at least its inat identifier) and reimport (we *do* find and reuse images and sounds leftover by previous imports in this scenario)

app/javascript/vue/tasks/field_occurrences/inaturalist_import/components/SummaryTable.vue
```diff
@@ -0,0 +1,159 @@
...
+          <td>{{ row.image_count ?? 'n/a' }}</td>
+          <td>{{ row.sound_count ?? 'n/a' }}</td>
+          <td>
+            <a
+              v-if="row.status === 'found'"
+              :href="row.browse_url"
+              target="_blank"
+            >Found</a>
+            <a
+              v-else-if="row.status === 'created'"
```
jlpereira: Perhaps it would be better to define the status types as constants in a file, same comment for all places where it is like that

app/javascript/vue/tasks/field_occurrences/browse/components/Panel/PanelIdentifier.vue
```diff
@@ -15,9 +15,18 @@ import useIdentifierStore from '../../store/identifiers.js'
 
 const store = useIdentifierStore()
 
+const INAT_OBSERVATION_TYPE = 'Identifier::Global::Uuid::InaturalistObservation'
```
jlpereira: It could be moved to /constants/identifierTypes.js

IDENTIFIER_GLOBAL_INATURALIST_OBSERVATION

app/javascript/vue/tasks/field_occurrences/inaturalist_import/components/ImportForm.vue
```diff
@@ -0,0 +1,167 @@
...
+    } else {
+      const queued = body.summary.filter(r => r.status === 'queued').length
+      const existing = body.summary.filter(r => r.status === 'already_imported').length
+      const notFound = body.summary.filter(r => r.status === 'not_found').length
+      const noTaxon = body.summary.filter(r => r.status === 'no_taxon').length
+      if (queued) parts.push(`${queued} queued`)
+      if (existing) parts.push(`${existing} already imported`)
+      if (notFound) parts.push(`${notFound} not found on iNaturalist`)
+      if (noTaxon) parts.push(`${noTaxon} skipped (no taxon)`)
+    }
```
jlpereira: This could be simplified by creating constants for status types,
then you could create another constant that use those contants as key and a function that compose labels as value. And then a function to summarize the status

PR #4915, opened 2026-05-07
Try to make unintended creates in Name Matcher less likely

kleintom:
Theories:
* people radio-select a name but nothing happens immediately, so they click Create on the same name from the third column
* people see `Select(ed)` in the third column title and think 'this is how I select a name'

Not sure either of these are really the case...

@jlpereira what do you think? This does mean (as written) that once you radio select in a row, you can't change your mind and Create instead (you'd have to quit the modal and reopen).


jlpereira:
Sounds good to me, maybe we could add a *`none`* option to the matched list when there are matched people, so the user could unselect it and create a new one without need to close the modal.

Issue #4944, opened 2026-05-26
[Bug]: Georeferences on CE2 are dropped by CE1.unify(CE2) when they're outside CE1's GeographicArea


kleintom:
### Steps to reproduce the bug

You can't actually reproduce right now because of #4937, but a month ago you could have, and you'd be able to after that gets fixed.
1. Create CE1 with a geographic area set.
2. Create CE2 with a georef outside CE1's geographic area.
3. Unify CE2 from CE1 (CE1.unify(CE2)).

See the expected failure report 'Geographic item for georeference is not contained in the geographic area bound to the collecting event; Collecting event is assigned a geographic area which does not contain the supplied georeference/geographic item', yet CE2 and its georef are deleted; the georef is not on CE1 and the unify ran even with the error.
...

### Environment

Production

mjy:
This is intended perhaps.  User is deciding which to keep and making a conscious decision.  Suppose they could error but inclined to leave, maybe.  Batch mode is a different beast though.

Issue #4937, opened 2026-05-22
[Bug]: CollectingEvents with georeferences can't be destroyed


kleintom:
### Steps to reproduce the bug

```bash
1. Create a CollectingEvent with a Georeference.
2. Try to destroy the CollectingEvent.
Get an error (500). Happens with CE#unify as well (since that includes a CE destroy).
...
```

### Environment

Production

PR #4947, opened 2026-05-29
CE with georefs destroy fix; unify fixes

kleintom:
Combining issues #4937 and #4944 here since 4937 is required for 4944 and 4944 fixes ones of the symptoms of 4937 (error on CE unify when georefs present [though 4944 fixes a pre-4937 error there).

Fixes:
1. Destroying a CE with a georef raises an error.
2. CE1.unify(CE2) where CE2 has georefs drops the georefs.
3. Fix some issues with the same root issue as 2^ causing:
  * observation_matrix_column_items to be dropped on descriptor unifies
  * observation_matrix_column_items to be dropped on Keyword (CVT) unifies
  * observation_matrix_row_items to be dropped on Keyword (CVT) unifies
  * observation_matrix_row_items to be dropped on TaxonName unifies
4. CE1.unify(CE2) where CE1 has a GA set and CE2 has a georef outside the GA raises an error, but unify ignores the error and deletes CE2 and its georef, instead of the error blocking the unify.
5. \#unify wasn't acts_as_list-aware, so when unifying two CEs with identifiers (e.g.) where new identifiers are added to the top of the list, unifying would both:
  * make the original CE's identifiers appear last
  * put the destroyed CE's identifiers first, in a different order than they were
[TODO? not so cool that the fix for this sticks out in the unify code where everything else is very model-agnostic; is there a more clever way to make it cleaner?]
6. Move the list of unifiable models to Ruby so that we can run specs over the class list.

Brief comments on the bugs themselves:
1) we destroy the GI for the georef from a callback, but then run another callback after that (CE geo heirarchy names) that uses the same AR - *without the fact that the georef's GI has been destroyed* - to run the CE's georef GIs.
2) The \#unify code does *not* process relations that use `class_name:`:
https://github.com/SpeciesFileGroup/taxonworks/blob/332f953551cb12307a8970092d18e4c7e94f09c6/app/models/concerns/shared/unify.rb#L70-L76
but some legit relations that should be processed do used `class_name`, such as georefs on CEs, where the `class_name: ::Georeference` (top-level Georef model) is required so that Ruby doesn't find collecting_event/georeference.rb (georef concern) instead.
3) Similar issues to 2)
4) in \#unify, `log_unify_result` can record a failure in its result hash, but failure in that hash wasn't being checked.
https://github.com/SpeciesFileGroup/taxonworks/blob/332f953551cb12307a8970092d18e4c7e94f09c6/app/models/concerns/shared/unify.rb#L164-L214
6) Two new linting specs are how we now avoid new errors of type 2) and avoid adding new relations on unifiable models without inverse_of
  * there's an existing rake task for inverse_of, but no guarantee that it would run on each new relation added
  * there was one missing inverse_of, on `has_many :observation_matrix_column_items` in Descriptor


mjy:
> observation_matrix_column_items to be dropped on descriptor unifies
> observation_matrix_column_items to be dropped on Keyword (CVT) unifies
> observation_matrix_row_items to be dropped on Keyword (CVT) unifies
> observation_matrix_row_items to be dropped on TaxonName unifies

Does this mean that if one OTU is in a matrix and one is not, and they get unified, that the process results in neither OTU being in the matrix?

kleintom:
> > observation_matrix_column_items to be dropped on descriptor unifies
> > observation_matrix_column_items to be dropped on Keyword (CVT) unifies
> > observation_matrix_row_items to be dropped on Keyword (CVT) unifies
> > observation_matrix_row_items to be dropped on TaxonName unifies
> 
> Does this mean that if one OTU is in a matrix and one is not, and they get unified, that the process results in neither OTU being in the matrix?

The first one (for descriptors) means: if you unify two descriptors, any matrices that had columns on the "destroy" descriptor no longer have a column for that descriptor [in practice the unify may get blocked if there are observations on any of those columns, I'm not sure (the specs recreate the simplest case with no observations)].

The last 3 all deal with dynamic rows and columns: 
* if you have a dynamic row on a TaxonName (so the actual row_items are OTUs on a given taxon name) and you unify the TaxonName with the dynamic row as the "destroy" taxon_name in a unify, then any matrix that had those dynamic rows and otu row items now has neither (they're all destroyed with the taxon name)
* similarly for dynamic rows/columns created by a tag: if the tag is destroyed as part of a unify, the CVT relation to the matrix is dependent: destroy, so those rows/columns/row_items/column_items are silently destroyed.

Because unify doesn't recognize these associations as ones that need to be unified, they never appear in preview or in the unify result that the user sees, they're just silently destroyed in the background.

kleintom:
Merging to make unify changes for #4971 as well.

PR #4951, opened 2026-06-03
Set TN#cached_misspelling correctly after destruction of TNR::misspelling

kleintom:
You never see the bug from New Taxon Name because it always resaves the TN after a TNR delete.

@proceps / @mjy is this okay?

### DIFF
--- a/app/models/taxon_name_relationship.rb
+++ b/app/models/taxon_name_relationship.rb
@@ -480,7 +480,7 @@ class TaxonNameRelationship < ApplicationRecord
 
         if TAXON_NAME_RELATIONSHIP_NAMES_MISSPELLING_ONLY.include?(type_name)
           t.update_columns(
-            cached_misspelling: true,
+            cached_misspelling: destroyed? ? nil : t.get_cached_misspelling,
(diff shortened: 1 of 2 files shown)


mjy:
> it always resaves the TN

Why? This seems to be the real bug.

kleintom:
> > it always resaves the TN
> 
> Why? This seems to be the real bug.

I'm not sure, do you remember @jlpereira?
https://github.com/SpeciesFileGroup/taxonworks/blob/897f3856f949e75f8d786ee55069a998fe38d440/app/javascript/vue/tasks/nomenclature/new_taxon_name/components/relationshipPicker.vue#L273-L279

jlpereira:
I don't remember why we did this. Perhaps it's because we need to retrieve the object again due to some TNR-related change, and we'll probably use update because if we simply reload the taxon, we'd lose data that the user filled in on the form but didn't save.

proceps:
I think the goal is to refresh the page with new cached values and new SV messages.

kleintom:
Merging, we can do a followup on the load vs. save issue if we want.

Issue #4948, opened 2026-06-02
[Bug]: It's possible to create duplicate tags on an object, all with position 1


kleintom:
From Josh, https://github.com/SpeciesFileGroup/taxonworks/issues/4934#issuecomment-4510818798:
```
This was the source I encountered this issue with: Briscoe, M.C. (1947) Insect reconnaissance in Liberia, West Africa. Psyche (Cambridge), 54, 246–255.

I tried clicking save and got the error "Identical keyword attempt"
```

From the db:
```
Tag.where(tag_object_id: 214035)
  Tag Load (0.5ms)  SELECT "tags".* FROM "tags" WHERE "tags"."tag_object_id" = $1 /* loading for pp */ LIMIT $2  [["tag_object_id", 214035], ["LIMIT", 11]]
=> 
[#<Tag:0x00007f01466eb060 id: 695675, keyword_id: 5466, tag_object_id: 214035, tag_object_type: "Source", tag_object_attribute: nil, created_at: "2024-06-27 11:54:40.132461000 +0000", updated_at: "2024-06-27 11:54:40.132461000 +0000", created_by_id: 2592, updated_by_id: 2592, project_id: 80, position: 1>,
 #<Tag:0x00007f01466eaf20 id: 695909, keyword_id: 5466, tag_object_id: 214035, tag_object_type: "Source", tag_object_attribute: nil, created_at: "2024-06-27 11:54:42.189457000 +0000", updated_at: "2024-06-27 11:54:42.189457000 +0000", created_by_id: 2592, updated_by_id: 2592, project_id: 80, position: 1>,
 #<Tag:0x00007f01466eade0 id: 695963, keyword_id: 5466, tag_object_id: 214035, tag_object_type: "Source", tag_object_attribute: nil, created_at: "2024-06-27 11:54:42.554914000 +0000", updated_at: "2024-06-27 11:54:42.554914000 +0000", created_by_id: 2592, updated_by_id: 2592, project_id: 80, position: 1>,
 #<Tag:0x00007f01466eaca0 id: 698353, keyword_id: 5466, tag_object_id: 214035, tag_object_type: "Source", tag_object_attribute: nil, created_at: "2024-06-27 11:54:57.424471000 +0000", updated_at: "2024-06-27 11:54:57.424471000 +0000", created_by_id: 2592, updated_by_id: 2592, project_id: 80, position: 1>,
[...]

PR #4952, opened 2026-06-04
Move tag uniqueness check from concern to model #4948

kleintom:


kleintom:
Well covered by specs I think, merging.

mjy:
Nice normalizations, thanks.

PR #4960, opened 2026-06-10
4920 Rake task to cleanup duplicate Roles and TaxonNameRelationships

kleintom:
The duplicate data that will be cleaned up by this script are currently blocking Project Unifies that we'd like to do (note this is a rake-ified version of the script from internal curation issue 70, see that for more backstory).

The first three commits are no-code findings from AI for the author roles duplicates, the conclusion being that it seems like only Taxon Name Author roles are still producing duplicates. Those should be fixed by José's 068c4eaa09e71fd05ddcf8019fdff00707b3892e.

So it's possible we'll never need this rake task again after this cleanup, but we'll commit it anyway (it's no more work).

Local run output, it takes under a minute:
```
[klein:~/sandbox/taxonworks2] <3.4.7> 4920_dup_roles_and_tnrs+ 18s 1 ± rake tw:maintenance:duplicates:cleanup_roles_and_taxon_name_relationships user_id=2698 PREVIEW=false
********************************************************************************
!! RUNNING FOR REAL !!
********************************************************************************

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
Keeping 107 dup roles, destroying 110 dup roles...
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
..............................................................................................................

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
Keeping 293 dup taxon name relationships, destroying 300 dup taxon name relationships...
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
............................................................................................................................................................................................................................................................................................................

Role records written to /tmp/role_dup_cleanup-20260610_180452
TaxonNameRelationship records written to /tmp/tnr_dup_cleanup-20260610_180510
```

### DIFF
--- a/lib/tasks/maintenance/duplicates.rake
+++ b/lib/tasks/maintenance/duplicates.rake
@@ -1,0 +1,319 @@
+namespace :tw do
+  namespace :maintenance do
+    namespace :duplicates do
+
+      desc 'Destroy duplicate Roles and TaxonNameRelationships. Dry run by default; use PREVIEW=false to commit. Requires user_id=<id>.'
+      task cleanup_roles_and_taxon_name_relationships: [:environment, :user_id] do
+        $stdout.sync = true
+
+        preview = ActiveModel::Type::Boolean.new.cast(ENV.fetch('PREVIEW', true))
+
+        puts '*' * 80
+        puts preview ? 'Dry run, transaction rolled back at the end' : '!! RUNNING FOR REAL !!'
+        puts '*' * 80
+        puts
+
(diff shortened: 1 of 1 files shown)


mjy:
LGTM @kleintom

PR #4962, opened 2026-06-11
Add extends=descendants to the TN monograph api endpoint

kleintom:
The count guard in the descendants case is intended to protect the db from somebody requesting all that info for the entire taxon tree, e.g., all pre-loaded.

We could maybe switch to paged results if that's needed.


mjy:
> We could maybe switch to paged results if that's needed.

Probably a safer more generic approach that specific guards?

PR #4973, opened 2026-06-23
Updates to the Cite Verbatim Taxon Name Authors task

kleintom:
These were discussed at last week's Community Consult.

The main changes:
* add a Source selector to the modal currently showing the 10 most recent Sources when you click on the Cite button (so that you can choose a source that was entered long ago but never processed until now)
* once a row's names have been cited, the 'Filter Names' link and 'TaxonNames' button no longer work (because the names no longer have verbatim); now those go away and the 'TaxonNames' button turns into a 'Page numbers' button that you can click to add the page to each Citation that was created by the Cite button click:
image

### DIFF
--- a/app/javascript/vue/routes/endpoints/SoftValidation.js
+++ b/app/javascript/vue/routes/endpoints/SoftValidation.js
@@ -16,11 +16,15 @@ export const SoftValidation = {
       filterParams(params, permitParams)
     ),
 
-  find: (globalId, config) =>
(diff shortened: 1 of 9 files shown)


kleintom:
@jlpereira note the app/javascript/vue/routes/endpoints/SoftValidation.js change - it looks like Claude found an issue where we're sending a 3rd param on an axios.get, but axios.get only supports two params.

PR #4981, opened 2026-06-30
Match OTUs task minor updates

kleintom:
Some small adjustments on the Match OTUs task:
* raised the names limit to 3k from 1k - we discussed raising it to 10k in chat, but at 5k (with the current code) Chrome warned that the browser was unresponsive when building the table.
  * At 3k, building the table takes around 10s, doing the matching takes around 10s
  * Maybe we'll want to switch to a virtual scroll if we really want to support higher limits (@jlpereira)

* Added a 'copy column' button to the columns it makes sense for - in practice I just wanted the `otu_id` column since I already had the scientific name column.
image

* Made the regexes apply when the user clicks the 'Match' button
* Added help bubbles to columns/options I didn't really understand and to explain what clicking an individual row does
* Added support for ctrl+v csv pasting in addition to drag-and-drop

### DIFF
--- a/app/javascript/vue/tasks/otus/match_by_taxon_name/app.vue
+++ b/app/javascript/vue/tasks/otus/match_by_taxon_name/app.vue
@@ -160,15 +160,15 @@ function applyModifiers(name) {
   return result.trim()
 }
 
-function computeMatchStrings() {
+function computeMatchStrings(matchAll = false) {
   const hasActiveModifier = modifiers.value.some((m) => m.active && m.pattern)
 
   rows.value.forEach((row) => {
     if (row.isEmpty) return
 
-    if (hasActiveModifier && row.selected) {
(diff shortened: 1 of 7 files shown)


mjy:
> Made the regexes apply when the user clicks the 'Match' button

All the facet changes are supposed to trigger an update automatically IIRC.I honestly don't remember why the "Match" button is there.

kleintom:
> All the facet changes are supposed to trigger an update automatically IIRC.I honestly don't remember why the "Match" button is there.

Currently facet changes only trigger match updates on rows that are checked.

If every facet change applies to every row then the "Match" button could go away, and maybe the checkboxes go away as well??

mjy:
> only trigger match updates on rows that are checked.

Yes, this is definitely wanted.  I meant to imply that scoping for "every row", sorry for confusion.

PR #5014, opened 2026-07-21
Combination touchups

kleintom:
@mjy/@proceps I think the first commmit is a straightforward fix, but please take a look at the second commit:

While investigating issues related to the SV fix I got onto this code that's come up in CCs several times that auto-adds a verbatim name to Combinations whenever a gender on the combination's genus is added and the Combination doesn't already have a verbatim name. 

The change here is that we now only add the verbatim name when the gender change actually changes an ending - the thinking being that
* we don't want to be using verbatim names more than needed, and
* it's not generally clear to users why this verbatim name shows up in the first place, so let's limit when we add it.

Am I missing anything?

### DIFF
--- a/app/models/combination.rb
+++ b/app/models/combination.rb
@@ -722,9 +722,8 @@ class Combination < TaxonName
       begin
         TaxonName.transaction do
           # update_column(:parent_id, check.parent_id) ## do not use this, it breaks the taxon_name_hierarchies
-          parent_id = check.parent_id
-          self.save
-          return true
+          self.parent_id = check.parent_id
+          return true if self.save
         end
       rescue
       end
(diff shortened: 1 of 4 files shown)


kleintom:
@proceps, note the failing sv has been covering up many invalid Combinations in your and other projects. I haven't yet been able to figure out how they're being created.

mjy:
My quick thinking.

@proceps added the verbatim check for import/checking after the fact IIRC.  We do indeed need to work to remove this entanglement at a universal level IMO.

For batch/import options we should explicitly set the verbatim name at creation, IF needed (optional).  I understand that @proceps wants this as a worflow/check system, but workflow/check getting to integrated in data makes a mess, so we need to keep thinking of how to isolate this. 

I personally would never write verbatim unless user is batch loading (script/tool), and it has been opted-in to, it should never be automatic behind the scenes IMO, even in the gener change/not.

kleintom:
@proceps and I discussed some, he looked over the changes and was okay with both. I think my framing would be that this is a step forward by removing unnecessary verbatim names in the gender-change case that everybody already agrees on :)

PR #5028, opened 2026-07-26
5027 project organization

mjy:
@kleintom @jlpereira This is ready for code review I think.  Our main goal here is to allow for TW intances to gather data about its users in an opt-in way.  Lets you configure projects to 1 or more org, lets Orgs have a geographic ID (to get country), and lets them be depicted (to get logos of orgs).  The logo bit will need some refinement, but that will come in "Render" model updates.


app/helpers/workbench/navigation_helper.rb
```diff
@@ -5,7 +5,7 @@ module Workbench::NavigationHelper
 
   NO_NEW_FORMS = %w{Confidence Attribution ObservationMatrixRowItem ObservationMatrixColumnItem ObservationMatrixRow ObservationMatrixColumn Note Tag
   Citation Identifier DataAttribute AlternateValue TaxonNameClassification
-  GeographicArea ContainerItem ProtocolRelationship Download}.freeze
+  GeographicArea ContainerItem ProtocolRelationship Download ProjectOrganization}.freeze
```
kleintom, line 8: There's no New form because of this, but a non-admin user can still destroy a Project Org from its nav radial or its show page.

app/controllers/project_organizations_controller.rb
```diff
@@ -0,0 +1,94 @@
+class ProjectOrganizationsController < ApplicationController
+  include DataControllerConfiguration::ProjectDataControllerConfiguration
+
+  before_action :set_project_organization, only: %i[show destroy]
+  after_action -> { set_pagination_headers(:project_organizations) }, only: [:index], if: :json_request?
```
kleintom, line 5: Do we want `:require_superuser_sign_in` for create and destroy here?

app/javascript/vue/tasks/projects/organizations/app.vue
```diff
@@ -0,0 +1,82 @@
...
+    })
+    .finally(() => {
+      isLoading.value = false
+    })
+}
+
+function addOrganization(organization) {
+  ProjectOrganization.create({
+    project_organization: { organization_id: organization.id }
+  }).then(({ body }) => {
```
kleintom: Might want to put `.catch(() => {})` on these since they could in theory fail, and then we'd have an uncaught exception.

app/views/api/v1/base/index.json.jbuilder
```diff
@@ -1,9 +1,29 @@
...
+  json.array! open_api_projects.includes(project_organizations: [:organization]) do |project|
+    json.name project.name
+    json.project_token project.api_access_token
+    json.data_curation_issue_tracker_url project.data_curation_issue_tracker_url
+
+    json.organizations project.project_organizations do |project_organization|
+      organization = project_organization.organization
+
+      json.extract! organization, :id, :name, :alternate_name, :legal_name,
+        :address, :email, :telephone, :geographic_area_id
```
kleintom: Maybe show the GA name instead of just the id for the api?

app/views/project_organizations/autocomplete.json.jbuilder
```diff
@@ -0,0 +1,12 @@
+json.array! @project_organizations do |project_organization|
```
kleintom: Search boxes aren't activated in show or list yet (not sure if they're intended to be).

kleintom (commented):
Looks good to me.

kleintom:
I almost forgot one other thing: on the Data tab, Project Organizations shows up in the Core (top) section, which isn't what I would have guessed.

app/controllers/project_organizations_controller.rb
mjy: Done.

app/javascript/vue/tasks/projects/organizations/app.vue
mjy: @jlpereira please?

kleintom: I can do these quick

app/views/api/v1/base/index.json.jbuilder
mjy: Done.

app/javascript/vue/tasks/projects/organizations/app.vue
kleintom: done

app/views/project_organizations/autocomplete.json.jbuilder
mjy: Don't think we need them, it's just a couple.

PR #5030, opened 2026-07-28
4971 unify updates

kleintom:
@mjy, could you take a look please? If looking at individual commits, keep in mind that I changed up some of the fixes more than once. Also: the two big chunks of changes in unify.rb and tnr.rb are just indentation changes.

Fixes for the following unify issues:
**A)** 
`n` --`r1`--> `k`[eep]
`n` --`r2`--> `d`[estroy]
where `r1` and `r2` are the same type of taxon_name_relationship. (Same issue in the other direction.)

`k.unify(d)` would fail because the uniqueness validation on TNR is a straight rails validation, and rails puts its validations on the column name (ending in "_id" for a FK) and not the association name (not ending in "_id"), which is what `unify` checks.

The eventual fix here was to do away with trying to compare the error column with the column we reassigned: continuing to hand-change the error columns one by one is (probably) doable but we catch them late (and it doesn't feel right), trying to automate it quickly turned into wackamole, and I'm not seeing any issue with just always checking to see if there's a duplicate and unifying the duplicate if there is one. If there *is* a single duplicate then we can try to merge, and if the unify merge succeeds, then what's the problem (honest question :)? The result is what we wanted.

**B)**
TNs `k`(eep) and `d`(estroy), `s` a synonym of `d` (or any TNR related to `d` I think).

`k.unify(d)`

During the unify we move the synonym from `d` to `k`, reload the synonym (we just do), and then unify commits its transaction, which is when `after_commit` on the synonym actually runs.
But when we reload the TNR its "dirty" bit is reset, and `after_commit` only fires when an object has been changed, so the `after_commit` on TNR that udpates its subject/object never runs, so the synonym's `cached_valid_id` stays `d.id` instead of getting updated.

Now everytime we try to access `s` it has a subject or object that points to a TN that no longer exists, boom. [This was Dima and Donald's issue.]

I tried out a couple fixes here, but eventually I decided that it was actually TNR that was using a dangerous pattern here: relying in `after_commit` on a bit that gets cleared on every save/update/reload (where in the "unify" case the `after_commit` may run "much" later than an intermediate save/reload/update). The fix was to add an ivar to TNR (`@taxon_name_id_reassigned`) that would track whether the subject/object_taxon_name_id of a TNR AR had ever been changed, so that the correct thing would happen in `after_commit`.

**C)** An issue specific to ADs, because they're required to always have at least one citation:

`ad_keep`, `ad_destroy`, "same" citation on both.

`ad_keep.unify(ad_destroy)`

Prior to a) the inner Citation merge used to fail, and we used to reload after a failed inner unify, and if the reload was valid we called the merge successful - that was causing us to quietly leave `ad_destroy`'s citation on itself, followed by destroying `ad_destroy`, its citation, and any annotations the citation had which should rather have been moved to `ad_keep`'s identical citation.

With a) we successfully saw the two citations as "identical" when moved to `ad_keep`, so we tried to merge-destroy one, which went boom because the citation said "I have to stick around for my AD!" even though that AD was about to be destroyed.

The fix here was to use Rails' builtin thread-local global store `ActiveSupport::CurrentAttributes` (used once before in a similar situation) to track the "destroy" objects of unify as it goes through its inner/outer unify/merge calls - then when Citation looks to see if it's still needed for its object (an AD), it can look on the store and say "oh, my AD is about to destroyed, it's okay to let myself be destroyed". [In the end I considered this global cleaner and more comprehensive than threading the outer destroy object into the inner merge code and then adding a callback for Citation to read it.]

(It turns out this is a non-issue for FO's required TD, since TDs come with a `position` value that causes us to take a different path, but that's a bit accidental, so TD sets and checks the global as well for future-proofing.)

Minor fixes:
d) After b) completes, the `keep` AR used to have an out-of-date subject or object taxon_name value on its AR which could explode; now we update the AR.

e) Fixed a typo in unify checking for a cross-project unify.


kleintom:
FYI I'll have some followup commits that rely on the unify fix(es) here to clean up some AP/OR/unify issues that are probably less urgent, but planning on waiting for this to clear for now before posting.

mjy:
@kleintom Thanks for the great explanation. This is all fine by me. Some thoughts, not requiring changes:

> But when we reload the TNR its "dirty" bit is reset

In the context of unify() can we "simply" turn it back on? 

> UnifyDestroyContext

I like this idea of managing more(?) context here. Does @taxon_name_id_reassigned only matter in unify context?  Could we assert that for now it does, and think about a generic `@reassigned` state being moved to this class.  Haven't thought it through fully- maybe not, thoughts?

kleintom:
Thanks @mjy.
> > But when we reload the TNR its "dirty" bit is reset
> 
> In the context of unify() can we "simply" turn it back on?

In this particular case we probably could, but in general whether or not that works would depend on whether changing that one attribute knock-on changes others, and whether after_commit also checks for changes in those others. (Probably safe but also probably not verifiably safe.)

kleintom:
**D**) Unifies where the keep and/or destroy object have *multiple* associations coming from the same object. The known cases are:
* keep/destroy = TN, associated object is TNR - the only TNR classes allowed to have both subject/object the same are `OriginalCombination` ones (`OriginalSpecies`, `OriginalGenus`, etc).
* associated object is BA, keep/destroy = OTU, CO, etc.
* keep/destroy = OTU, associated object is OtuRelationship
* [keep/destroy = Serial, associated object is Serial, associations are translations (has_many) and translation_from_serial (belongs_to) - the fact that only one of these is belongs_to means there's no issue]

Consider the TN case:
`k` with a self-referential OriginalSpecies TNR, `d` with its own OriginalSpecies TNR. Each TNR has a subject and an object relation - the key issue is that currently the `unify` code moves those associations *one at a time*. So if we're doing `k.unify(d)`, we see `d`'s OriginalSpecies associations and want to move them both to `k`, starting with whichever one comes first, say `object_taxon_name_id`: we set `tnr_d.object_taxon_name_id = k` but we still have `tnr_d.subject_taxon_name_id = d`, plus `k` still has its own `k --original_species--> k` TNR. Note that the current state of `tnr_d` is not a true statement, before or after the intended unify: it currently says 'the original species of k is d', but `k`'s relation says 'the original species of k is k'. Both can't be true so we have a validation error. It's not a uniqueness error, so the unify fails.

If we had moved *both* sides of `tnr_d` before checking the result we would have changed `tnr_d` to a duplicate of `tnr_k`, which we know how to handle.

That's the solution added here: we now move *all* foreign keys on a given associated object the first time we encounter any foreign key on that object. That reflects the intended final state of the unify and gives the correct result (when nothing else fails).

Other additions:
* tightened down the linting spec that checks that unifiable models have their association inverses set up correctly - if they don't `unify` can either fail, or, worse, succeed with corrupted data.
* many specs to cover the TN/BA cases:
  * `k` has a self-ref
  * `d` has a self-ref
  * both have a self-ref
  * `k` has a `k-->d` ref, `d` has a self-ref
  * etc.

The validation on TNR is stricter than the validation on BA, so fewer of the TN cases are valid to begin with, and a couple of them still fail for the following reason:
* we now process all relations from each object at the same time to reflect the final state of unify viz-a-viz that object, but we don't yet process *all associated objects* to reflect final unify state prior to checking validations. In some of the (rarer) TNR cases that actually matters - that fix will be in a followup.

PR #5031, opened 2026-07-28
5013 autocomplete mark

kleintom:
* switched to a single mark function that:
  * doesn't mark inside html tags (<...>) or the common html entities (`&amp;` e.g.)
  * does *not* mark across html boundaries
* *all* autocompletes now mark their output (though we may decide that's not what we want)
* switched to a new mark color because the previous color didn't work against the gray of OTUs:
image
image
image
image

@jlpereira / @mjy any feedback? I'm not really tied to any of this; @jlpereira feel free to finish this one off if you want to experiment :)

### DIFF
--- a/app/assets/stylesheets/vendor/jquery/autocomplete.scss
+++ b/app/assets/stylesheets/vendor/jquery/autocomplete.scss
@@ -60,11 +60,6 @@
     font-size: 10px;
     font-weight: 300;
   }
-  mark {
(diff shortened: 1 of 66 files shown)


jlpereira:
Looks good to me. It could just be bold text, but it might get lost among the colored tags. I think it's fine.

mjy:
Let's try it.

PR #5034, opened 2026-07-31
4965 virtual leads

gdower:
This branch adds the ability to cite keys in literature without turning them into an interactive key by using the existing leads model but extending it to allow virtual leads. The UI supports a step 1 workflow of searching for a citation (or using the pinboard), checking if existing keys are cited from that citation and showing them to edit instead, scoping by parent Otu and adding children, then bulk adding annotations onto the children in step 2.

@mjy also had the idea of using the radial linker from the filter Otu task do advanced filtering on Otus and then cite a key with the result list, which is supported.

I think the bulk annotations interface could be abstracted out into another task for doing bulk annotations in a spreadsheet-like interface, although being able to embed it into tasks like this seems useful in order to support data entry workflows. The spreadsheets interface detects what annotations were added to the object within a project, and automatically adds those columns. It is also easy to add additional columns. You can expand to full screen to have a data-entry-focused view and keyboard navigation is spreadsheet-like:

Screenshot_20260731_105201

I envision a step 3 in the future that might facilitate turning a literature citation into an interactive key.

Open questions:

1) The task does include a radial navigator for keys but it mostly links to interactive key features and that may not be helpful. Perhaps I should remove it.

2) The radial linker does cap at sending a max of 2000 taxa to the task, but based on querying the prod database most keys have less than 100 taxa (max: 262), so I think capping to 2000 is reasonable? A toast message warns that the result list was capped.

3) I would like to add the ability use biocuration classes as an annotation type because ScaleNet in particular wants to put sex and life stage on the keys and it would be better to use that annotation type than keyword/tags, but I did not do that in this branch.


gdower:
I cut the virtual leads task down to just step 1 of the previous task. The table annotator is re-implemented in a separate branch as a separate task (although I am not opening a PR on that yet): https://github.com/SpeciesFileGroup/taxonworks/tree/table_annotator

app/javascript/vue/tasks/leads/cite_key/app.vue
```diff
@@ -0,0 +1,1247 @@
...
+            />
+            Is publicly accessible?
+          </label>
+        </div>
+
+        <p
+          v-if="rootGlobalId"
+          class="small_type margin-small-top"
+        >
+          Use the radial annotator above to record attributes that apply to the
```
kleintom: Need to add the model annotator (quick forms - not the radial annotator) to be able to add biocurations

app/javascript/vue/tasks/leads/cite_key/app.vue
```diff
@@ -0,0 +1,1247 @@
...
+          class="feedback feedback-info padding-small text-center"
+        >
+          No taxa in this key yet. Add OTUs above, or check
+          <em>Add descendants on save</em> in the metadata block and save again.
+        </div>
+      </template>
+    </BlockLayout>
+</template>
+
+<script setup>
```
kleintom, line 289: Followup nit: add 'alt+s' keyboard shortcut for save (maybe others)

app/javascript/vue/tasks/leads/cite_key/app.vue
```diff
@@ -0,0 +1,1247 @@
...
+      <template #header>
+        <h3>Tags</h3>
+      </template>
+
+      <template #body>
+        <div
+          v-if="!leadKeywords.length && !tags.length"
+          class="small_type padding-xsmall"
+        >
+          No tags have been applied to any keys in this project yet. Add one via
```
kleintom: [followup] Currently you need to reload the page to see new tags added from the radial annotator, but the annotator does have an emit you can listen for to update tags on the page as they're added/removed on the annotator.

app/javascript/vue/routes/routes.js
```diff
@@ -103,7 +103,8 @@ const RouteNames = {
   SourceHub: '/tasks/sources/hub',
   TypeMaterial: '/tasks/type_material/edit_type_material',
   UnifyObjects: '/tasks/unify/objects',
-  UnifyPeople: '/tasks/unify/people'
+  UnifyPeople: '/tasks/unify/people',
+  CiteKey: '/tasks/leads/cite_key'
```
kleintom: Alphabetize please

app/javascript/vue/tasks/leads/cite_key/app.vue
```diff
@@ -0,0 +1,1247 @@
+<template>
```
kleintom, line 1: [Followup] This will be the longest .vue file in the codebase; that's a pretty strong signal that we can/should break out some smaller components from this monolith.

app/javascript/vue/tasks/leads/cite_key/app.vue
```diff
@@ -0,0 +1,1247 @@
...
+          <RadialNavigator
+            :global-id="rootGlobalId"
+            exclude="Edit"
+          />
+        </div>
+      </div>
+    </template>
+
+    <template #body>
+        <div class="field label-above">
```
kleintom: indentation is off here

app/javascript/vue/tasks/leads/cite_key/app.vue
```diff
@@ -0,0 +1,1247 @@
...
+    })
+    .catch(() => {
+      existingKeys.value = []
+    })
+    .finally(() => {
+      existingKeysLoading.value = false
+    })
+}
+
+function addSpecies(otu) {
```
kleintom, line 523: nit: I think we try to use `addToArray` for this

app/javascript/vue/tasks/leads/cite_key/app.vue
```diff
@@ -0,0 +1,1247 @@
...
+      descendants: true
+    }
+  }
+
+  if (descendantsFilter.value === 'valid') {
+    params.taxon_name_query.validity = true
+  }
+
+  return Otu.where(params).then(({ body }) => {
+    body.forEach((otu) => {
```
kleintom, line 582: [followup] Make this a batch call instead of one create per call?

app/views/citations/_attributes.json.jbuilder
```diff
@@ -8,6 +8,19 @@ json.citation_source_body citation_source_body(citation)
 if extend_response_with('citation_object')
   json.citation_object do
     json.partial! '/shared/data/all/metadata', object: citation.citation_object, extensions: false
+    if citation.citation_object.is_a?(::Lead)
```
kleintom: I don't think we want to start this pattern (loading extra data for a *specific* annotated model in the attributes for an annotation model), we probably need to bite the bullet and just do the Lead-data lookup after loading the citation.

kleintom:
I added 3 commits from Claude discovered by Claude's `code-review` task since they all seemed legit to me; and then a followup cleanup commit where I had broken convention in the Keys Hub (with buttons that replicated slices in the existing radial navigator) and another comment cleanup commit where we were using 'interactive' as an adjective in reference to dichotomous keys.

I'll post some review comments shortly as well.

kleintom (commented):
Most of the review comments can wait for followup, but please remove the Lead data from the Citation view, and I think the ones below we'll want to address before landing.

* [x] The delete on OTUs from the Cite a Key task doesn't currently work because the leads controller doesn't permit destroy/delete on non-root leads (so that we don't unintentionally leave a couplet with only one lead); we probably want to add a `destroy_simple_lead` action or similar that only works on virtual leads. [Side note: I think you don't see the error that gets reported because we're not catching that one, and we're reporting success at a different stage, something like that - if you delete, see the success message, but then reload the page, you see the error message]

* [x] Leads navigator radial - we still need to redirect most of those slices to their appropriate task depending on whether the lead is virtual or not:
* leads_controller:
  * edit should check for is_virtual and redirect to the cite task if so
* app/controllers/tasks/leads/print_controller.rb
  * before_action :redirect_if_virtual <-- add this
* similar for other lead navigator radials slices that would otherwise load a virtual lead where a non-virtual was expected

* [x] add `is_virtual` to app/views/leads/_attributes.html.erb and api attributes

* [x] http://localhost:3001/leads/list and http://localhost:3001/leads/ both *do not* currently show simple keys (cf. leads_controller#index->html and #list), not sure if that's what you intended or not.
* [x] neither does /api/v1/leads (the new /api/v1/leads/simple does)

* [ ] add the new task to the leads navigator; in the task controller you'll need to check for is_virtual and redirect to the "normal" edit route if not virtual
  * that's the only way to get to that task from the keys hub

* [ ] In antweb I search on taxon name 'azteca' in Filter OTU, check the 20 results it returns, and send to Cite Key. The Cite Key task says 'Prefilled 20 taxa from Filter OTUs' and then something about inferring the parent Azteca - the parent otu is set as expected, but I don't see any of the 20 OTUs listed as children. Did this maybe stop working after some redesign?

app/views/citations/_attributes.json.jbuilder
mjy: We definitely don't want to start that pattern.

mjy: DEFINITELY not with annotations.

app/controllers/leads_controller.rb
```diff
@@ -198,6 +268,27 @@ def delete_children
     end
   end
 
+  # DELETE /leads/1/destroy_simple_lead.json
+  # Simple (virtual) keys are flat by construction and are managed by the
+  # cite_key task. This action lets that task destroy either a species-child
+  # leaf or the entire simple key from the same endpoint. It refuses anything
```
kleintom, line 274: "or the entire simple key" --> iirc there's no builtin "destroy all descendants of this key when I'm destroyed" callbacks in Lead, so I don't think this would destroy the entire key if called on a simple key root?

AFAICT it's only being called in children right now (which I think is right - let the existing destroy handle the recursive destruction of the entire tree), so maybe just adjust the comment and assert that it's being called on something with no children.

kleintom:
* [ ] When you click the New button, clear any query params added from a Filter OTU push (otherwise a page refresh re-adds whatever was on the filter result)

app/javascript/vue/tasks/leads/cite_key/app.vue
kleintom: Ah, this comment is referring to a lead annotator, I was thinking something that had biocurations instead. Is the idea that they're just using data attributes for life stage, sex, etc. directly on the key?

Issue #130, opened 2017-05-04
Task - resolve OTUs to taxon names


mjy:
```
   when I open the task      
       then I see a paginated list of OTUs that have taxon_name_id # nil   (similar to otus/list)
         and each name is wrapped in its own asynchronous form
           and a fuzzy name match for possible TaxonNames lists possible quick selections
           and a autocomplete is available to find names manually
           when I select a taxon name
                 and I click update
                     and there is success
                         then the row is removed
                              and the recent updates list below the paginated OTUs is updated
       then I see a list of recently updated OTUs at the bottom of the page
```

![screen shot 2017-05-04 at 2 59 19 pm](https://cloud.githubusercontent.com/assets/22600/25722647/7bdc6f44-30da-11e7-947e-0707c532623c.png)

mjy:
Cross reference and potential alternative approach- https://github.com/SpeciesFileGroup/otu_mapping_widget

jrflood:
Overtaken by external events.  @mjy observed an already complete example of this functionality at a conference and our experiment was abandoned.

mjy:
Scaffolding this now.

PR #5036, opened 2026-08-02
Add Assign taxon name to OTU task

mjy:
Scaffolding for #130.  Significant work remains.  Will update here when ready to handoff.


mjy:
Scaffolded now.  Going to step way from this for a bit, happy to have others test/play.

Known issues:
* [x] This is the first time we use Autoselect en-masse.  It's showing a general problem, each instantiation is making a server call- 500 instance, 500 calls.  Here we want the same autoselect for all 500 instance, so 1 call suffices.  Unsure of where to store state, but it's almost certainly a basic init option, perhaps `inherit`esque- so only the first initializes, then the rest inherit.

jlpereira:
You could create a simple store in the AutoSelectField folder with a global ref that maintains the state of the autoselects. If they share the same configuration, they can retrieve it directly from there. This way, they can share the state between them, and only one autoselect needs to make the request to fetch the data. The others can simply read it from the global state.

Another option is to avoid displaying the autoselects directly. Instead, display a button that, when clicked, shows the autoselect and focuses the input

PR #5042, opened 2026-08-10
Improve Prawn's fonts for viewing non-latin characters

kleintom:
@jlpereira Could you give this one a second look when you get a chance? Maybe you know a better way to do this.

STR:
* Filter Source for a source whose bibliographic citation would have diacritics in it, e.g. author Güçlü (in hoppers).
* Select Download formatted from the download select
* download as pdf

Without this fix you get a 500 because prawn's default fonts are wimpy and can't encode diacritical characters etc.

Comparison to the current approach:
* one-off copy of font files from the npm pdfjs-dist package, *not* part of a package that will get updated
  * no legal issues afaict
* fonts are embedded, so pdf will be slightly larger, not a big deal
* the actual appearance of the font seems identical to the existing output, at least on my machine


jlpereira:
It looks good to me. My only doubt is whether it's better to do this in the controller or create a helper for it. I don't know if we generate PDFs in another location where this problem exists

The second thing is that this will probably fix the 500 error, but Liberation Sans has no CJK coverage, so Chinese/Japanese/Korean text won't 500 after this fix, but probably it will render as blank glyphs.

kleintom:
Thanks @jlpereira 

> whether it's better to do this in the controller or create a helper for it

good call, I think we could create a vendor wrapper for prawn to deal with this

> Liberation Sans has no CJK coverage

Ugh. I don't know too much about fonts so this is ai-guided:
prawn does have a `.fallback_font` method for this kind of issue, so we can use that, but:
AFAICT, a CJK font file like SourceHanSans-Regular.ttf is going to be in the 34MB range if we want to support *all* CJK. I'd rather not add that as a repo dependancy, so some possible options:
* we could make the 34MB font file a docker dependancy, so just people running raw dev would lose out or have to install it themselves
* subset only the CJK that our db currently uses, requires about 75Kb (for ~70 sources)
  * that means tofu as soon as somebody adds a new CJK source and downloads a citation for it
    * we could check and raise each time we output tofu so that at least it could get fixed in the next release
* ??

pdfs would only embed font data for the characters they actually use, so size there shouldn't be an issue.

jlpereira:
A dependency sounds good to me. I wouldn't subset to what's in our production DB, there could be institutions running TW outside SFG instances, and a subset derived from our data snapshot wouldn't match theirs.

kleintom:
@LocoDelAssembly could you take a look at the docker part of this (or all of it if you want)? We want a capable font so that we can create pdf bibliographies from Filter Sources (via the Formatted download option). The default font for Prawn can't do diacritics, so I added a small font for that, but then José noted that the small font I added doesn't cover CJK. That requires a larger font - the one here is 34MB - so the thinking is that we don't want to put that in the repo but instead just install it with docker installs instead.

LocoDelAssembly:
Not entirely sure about the situation here, but probably it is an option to install the fonts in Dockerfile.base? (Picking suitable Ubuntu packages to get the fonts needed, like Noto fonts if they're good)

PR #5101, opened 2026-09-20
Task "New taxon name": Open "Add new source" link in a new tab

weevil-see:
This is fixing a nuisance I ran into: After creating a new taxon name, I misclicked on "add new source", the task didn't open in a new tab, and navigating back to the last page (New taxon name), I had to re-select the taxon name that I had entered to continue editing. Hope it works like this, did not compile to check.


proceps:
you can just use browser Back button

weevil-see:
I did use the browser back button, but it did not work as expected:
[Screencast_20260921_081523.webm](https://github.com/user-attachments/assets/5940fa16-585c-4493-993b-f8248e5b9af0)

First click did nothing, second led me back to "New taxon name" with a message that the name was successfully created and I was led to create a new name (would have to re-select the last one to continue working). 

Note how the URL is becoming "https://sandbox.taxonworks.org/tasks/sources/new_source?taxon_name_id=2709893", I guess new source and taxon_name_id don't go well together...

jlpereira:
That's a turbolinks issue. I will fix it.

